Issue #5, opened 2015-05-07
Application Design


tarebyte:
I'd like to get feedback as to where in the process this application should be designed. I'm a fan of at least doing some general template design work done near the beginning.

Also since this is a GitHub sponsored open source project what about using [Primer](https://github.com/primer/primer) as a starting point for styling?

johndbritton:
:+1: on using Primer. Are you talking about visual design? I think you're probably going to want to use Primer as a base and add polish as you go. When we get to the end of the summer we might be able to get some visual design help if we ask nicely.

johndbritton:
I accidentally deleted my previous post, so sorry for the duplicate notification.

Definitely recommend using Primer, some lightweight visual design up front is probably worthwhile followed by spending some more time near the end of the summer polishing things up.

PR #6, opened 2015-05-09
Add GitHub's Primer and Remove asset gems

tarebyte:
From discussion in #5

This PR:
* Uses [bower](http://bower.io/) to manage vendored assets
* Installs [primer](https://github.com/primer/primer) (along with [octicons](https://github.com/github/octicons))
* adds `octicon` and `mega_octicon` view helpers with tests
* remove jquery-rails gem and replace it with jquery and jquery-ujs from the vendor folder
* Changes the application layout to make it more browser friendly

Not sure how I feel about checking the vendored assets into version control, but I really don't like gems that are just for assets (bootstrap-sass, jquery-rails etc).

Opinions?


johndbritton:
> Not sure how I feel about checking the vendored assets into version control, but I really don't like gems that are just for assets (bootstrap-sass, jquery-rails etc).

I think you're fine to vendor stuff that comes from bower. It's not my favorite thing, but we even vendor gems in version control for speedier and more reliable deployments (one less dependency).

tarebyte:
> t's not my favorite thing, but we even vendor gems in version control for speedier and more reliable deployments (one less dependency).

It seems better than trying to have bower install the dependencies on deployment to production for anything like that.

Also just so it's written down, there is currently an issue with octicons with the latest sprockets-rails github/octicons#50

johndbritton:
:shipit:

tarebyte:
Sounds good to me, I'll probably wait until github/octicons#50 gets resolved.

PR #7, opened 2015-05-12
Add GitHub authentication for user

tarebyte:
Getting started on authentication for users. I borrowed the `AuthHash` idea/style from https://github.com/24pullrequests/24pullrequests. I think it's a really good example of separating code that doesn't need to be in the `User` model into a service.


db/migrate/20150511185213_create_users.rb
```diff
@@ -0,0 +1,14 @@
+class CreateUsers < ActiveRecord::Migration
+  def change
+    create_table :users do |t|
+      t.string :provider,    null: false
+      t.string :uid,         null: false
+      t.string :nickname,    null: false
```
johndbritton: Does `nickname` represent the user's GitHub username? If so, I'd recommend using `login` for consistency with the GitHub API: https://developer.github.com/v3/users/#get-a-single-user

johndbritton:
Looks good. There are a couple empty test and CSS files in there. I'm assuming you're going to use those since this is WIP, but if you don't you can just delete the files you don't need.

db/migrate/20150511185213_create_users.rb
tarebyte: `nickname` does represent the GitHub username in the [OAuth hash](https://github.com/intridea/omniauth/wiki/Auth-Hash-Schema#schema-10-and-later), but I agree it does make more sense to have the username as `login`.

tarebyte:
@johndbritton I think this implementation good to go, I would like feedback on the test coverage. I never know whether I add too many tests or not enough. Or if I'm duplicating tests somewhere.

Thanks!

app/models/user.rb
```diff
@@ -0,0 +1,17 @@
+class User < ActiveRecord::Base
+  validates_presence_of   :provider, :uid, :login, :email, :token
```
johndbritton: Stroring `provider` here may be unnecessary, the only auth this app will ever use is GitHub.

app/services/auth_hash.rb
```diff
@@ -0,0 +1,58 @@
...
+      name:     name,
+      token:    token
+    }
+  end
+
+  private
+
+  attr_reader :user_hash
+
+  def provider
```
johndbritton: Same deal here, we're only going to have one provider.

db/migrate/20150511185213_create_users.rb
```diff
@@ -0,0 +1,14 @@
+class CreateUsers < ActiveRecord::Migration
+  def change
+    create_table :users do |t|
+      t.string :provider,    null: false
```
johndbritton: If you take out provider, don't forget to remove it here.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,47 @@
+require 'test_helper'
+
+class SessionsControllerTest < ActionController::TestCase
+  def setup
+    request.env['omniauth.auth'] = OmniAuth.config.mock_auth[:github]
+  end
+
+  test 'new should redirect to /auth/github' do
```
johndbritton: Kind of a personal preference of mine not to use `should` in test case descriptions as it's unnecessary. `#new redirects to /auth/github`

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,13 @@
+require 'test_helper'
+
+class UsersControllerTest < ActionController::TestCase
+  def setup
+    @user = users(:jim)
+  end
+
+  test 'should get show' do
```
johndbritton: Drop `should` here, maybe `test '#show returns success'` or something like that.

test/models/user_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'test_helper'
+
+class UserTest < ActiveSupport::TestCase
+  def setup
+    @github_omniauth_hash = OmniAuth.config.mock_auth[:github]
+    @user                 = users(:jim)
+  end
+
+  context 'validations' do
```
johndbritton: I'm torn about testing validations, we know ActiveRecord is well tested so these should work if they're in your `User` class. This is essentially testing that you called those methods as you expected.

I used to want to test everything like this but I don't usually do it anymore. Up to you if you want to keep these tests or not.

test/models/user_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    context 'uniqueness' do
+      [:uid, :login, :email, :token].each do |column|
+        should validate_uniqueness_of(column)
+      end
+    end
+  end
+
+  test 'create a new valid user from the auth hash' do
```
johndbritton: When I use Rspec I usually have a `describe` block to group all my tests of a particular method. You might consider using a context to do the same here.

```ruby
context '#create_from_auth_hash' do
  test 'creates valid user' do
    assert_difference 'User.count', 1 do
    User.create_from_auth_hash(@github_omniauth_hash)
  end
end

This is the style I'm most used to, again up to you how you want to do it.

**Comment by johndbritton** [comment]
Left a few comments regarding style, this looks good to go.

File: test/models/user_test.rb (hunk 7 above)
Comment (tarebyte) [inline test/models/user_test.rb]: It feels like overkill to me, but I never know. I'll pull these out.

File: test/models/user_test.rb (hunk 8 above)
Comment (tarebyte) [inline test/models/user_test.rb]: If I use `context` then the test looks like this. This is from the [shoulda gem](https://github.com/thoughtbot/shoulda)

```ruby
context '#create_from_auth_hash' do
  should 'create valid user' do
    assert_difference 'User.count', 1 do
      User.create_from_auth_hash(@github_omniauth_hash)
    end
  end
end
```

Now that I look at it again, I agree `should` doesn't  look quite right as part of a test.

This is what I'm going to use for my tests, this reads better to me (I'm not a big fan of nesting).
```ruby
test '#create_from_auth_hash creates a valid user' do
  assert_difference 'User.count', 1 do
    User.create_from_auth_hash(@github_omniauth_hash)
  end
end
```

test/controllers/users_controller_test.rb
tarebyte: :+1:

app/models/user.rb
tarebyte: Sounds good to me!

PR #9, opened 2015-05-13
Fix failure route for Heroku deployment

tarebyte:
I usually test my apps incrementally to make sure everything works on a production environment, this fixes the return of a failure during GitHub authentication.

@johndbritton for the GitHub authentication failures, do I actually need a POST route? Or can I get away with just GET?

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,6 @@ Rails.application.routes.draw do
   get '/login',  to: 'sessions#new',     as: 'login'
   get '/logout', to: 'sessions#destroy', as: 'logout'
 
+  get   '/auth/failure',            to: 'sessions#failure'
   match '/auth/:provider/callback', to: 'sessions#create',  via: [:get, :post]
-  match '/auth/failure',            to: 'sessions#failure', via: [:get, :post]
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,6 @@ Rails.application.routes.draw do
   get '/login',  to: 'sessions#new',     as: 'login'
   get '/logout', to: 'sessions#destroy', as: 'logout'
 
-  match '/auth/:provider/callback', to: 'sessions#create', via: [:get, :post]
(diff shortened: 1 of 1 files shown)


johndbritton:
> do I actually need a POST route? Or can I get away with just GET?

The [relevant code](https://github.com/intridea/omniauth/blob/da86405a781630af172dadbe67796ec88d6c4973/lib/omniauth/failure_endpoint.rb#L28-L32) in Omniauth:

```ruby
def redirect_to_failure
  message_key = env['omniauth.error.type']
  new_path = "#{env['SCRIPT_NAME']}#{OmniAuth.config.path_prefix}/failure?message=#{message_key}#{origin_query_param}#{strategy_name_query_param}"
  Rack::Response.new(['302 Moved'], 302, 'Location' => new_path).finish
end
```

It returns a 302 redirect. According to the Wikipedia [page](http://en.wikipedia.org/wiki/HTTP_302) on 302 responses, it's up to the client to either make the same request type to the new location (POST in our case) or change the request type to GET.

> Many web browsers implemented this code in a manner that violated this standard, changing the request type of the new request to GET, regardless of the type employed in the original request (e.g. POST). For this reason, HTTP/1.1 (RFC 2616) added the new status codes 303 and 307 to disambiguate between the two behaviours, with 303 mandating the change of request type to GET, and 307 preserving the request type as originally sent. Despite the greater clarity provided by this disambiguation, the 302 code is still employed in web frameworks to preserve compatibility with browsers that do not implement the HTTP/1.1 specification.

> As a consequence, the update of RFC 2616 changes the definition to allow user agents to rewrite POST to GET.

Based on all that, you might want to accept both POST and GET, as the behavior differs across various clients.

tarebyte:
:+1: thanks for the insight!

johndbritton:
I find this to be kind of hilarious.

1. Define standard.
2. Ignore standard.
3. Update standard to allow for ambiguity because everyone ignores the standard.

PR #11, opened 2015-05-13
Add classrooms (GitHub organizations)

tarebyte:
This is a WIP, I'll ping when it's ready


app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def show
+  end
+
+  private
+
+  def authorize_user!
+    begin
+      has_user_id = Organization.find(params[:id]).user_ids.include?(current_user.id)
+    rescue ActiveRecord::RecordNotFound
```
tarebyte: Not sure if I need this, I'm pretty sure in a production environment I would get a 404

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      redirect_to dashboard_path
+    else
+      redirect_to back, error: 'Could not delete classroom'
+    end
+  end
+
+  private
+
+  def verify_organization_ownership
+    unless current_user.organization_ids.include?(params[:id].to_i)
```
tarebyte: Not sure this is the way I want to handle this, I also thought about just doing a redirect unauthorized flash thing as well. Thoughts?

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      redirect_to '/404.html'
```
tarebyte: Not sure this is the way I want to handle this, I also thought about just doing a redirect unauthorized flash thing as well. Thoughts?

tarebyte:
This is coming along nicely, I want to add tests the `github_client` with `VCR` and then I think this will be ready for review.

test/controllers/organizations_controller_test.rb
```diff
@@ -0,0 +1,92 @@
+require 'test_helper'
+
+class OrganizationsControllerTest < ActionController::TestCase
+  def setup
+    @user             = users(:tobias)
+    session[:user_id] = @user.id
+  end
+
+  test '#new returns success' do
+    stub_get_json("https://api.github.com/users/#{@user.login}/orgs",
```
tarebyte: This is my first stab at this kind of testing, I definitely want to DRY this up though as there is a lot of repeated code. But overall what do you think?

test/services/github_client_test.rb
```diff
@@ -0,0 +1,38 @@
+require 'test_helper'
+
+class GithubClientTest < ActiveSupport::TestCase
```
tarebyte: I would also like to get feedback on this. VCR is really cool BTW. What do you think about the test coverage / style?

tarebyte:
@johndbritton this is good for a first round of critiques and suggestions. I've also posted below the basic flow of how adding a classroom works for now.

add-a-classroom-2

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    @organization               = Organization.new
+    @users_github_organizations = current_user.github_client.users_organizations.
+                                  collect { |org| [org.login, [org.login, org.id]] }
+  end
+
+  def create
+    org               = JSON.parse(params[:org])
+    login, github_id  = org.first, org.last
+
+    if Organization.where(github_id: github_id).present?
```
johndbritton: There might be an edge case here. What happens if two users have added the same organization as a classroom?

I can think of two ways to handle this:

1. Any user on the owners team can access the classroom using the classroom app
2. Scope classrooms to users

I think the first option makes the most sense.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      redirect_to new_organization_path, alert: 'Classroom has already been added'
+    else
+      if current_user.github_client.is_organization_admin?(login)
```
johndbritton: I recommend using `github_id` instead of login whenever you're doing user actions. That way, there's no chance of an error introduced by a username change.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+        organization  = Organization.new(login: login, github_id: github_id)
```
johndbritton: If you're storing the organization `login` you have to account for updating it whenever there's a change.

app/controllers/organizations_controller.rb
johndbritton: You might consider returning an unauthorized `401` status, redirect to home, and a message in the flash.

app/models/organization.rb
```diff
@@ -0,0 +1,6 @@
+class Organization < ActiveRecord::Base
```
johndbritton: In the interface, you're calling these classrooms. I'd suggest just calling everything classrooms. Organization is a thing on GitHub, `Classroom` is a thing in this app.

app/services/github_client.rb
```diff
@@ -0,0 +1,26 @@
+class GithubClient
+  def initialize(login, token)
+    @login = login
+    @token = token
+  end
+
+  def is_organization_admin?(org)
```
johndbritton: As I commented [above](https://github.com/education/classroom/pull/11/files#r30479861) I'd recommend using the `github_id` here.

app/views/organizations/new.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<%= form_for @organization, url: { action: 'create' } do |f| %>
+  <dl class="form">
+    <dt><%= f.label :organization %></dt>
+
+    <dd class="select-organization">
+      <%= select_tag :org, options_for_select(@users_github_organizations) %>
+    </dd>
+
+    <div class="form-actions left">
+      <%= f.submit 'Add Organization', class: 'btn btn-primary' %>
```
johndbritton: For consistency, use `Classroom` here.

app/views/organizations/show.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1 class="page-header">
+  <%= @organization.login %>
+</h1>
+
+<%= link_to 'Remove Classroom', @organization, class: 'btn btn-sm btn-danger', role: 'button',
```
johndbritton: What does remove classroom do? Will it reset the organization? Anything we need to warn the user about?

test/controllers/organizations_controller_test.rb
johndbritton: I'd create a context for all the test of the same action and group them that way. Once they're grouped, you can reduce duplication in your stubs.

johndbritton: I'm not familiar with Minitest, but I'd imagine you can do something like this:

```ruby
context '#new' do
  stub_get_json("https://api.github.com/users/#{@user.login}/orgs",
    [{ login: 'testorg1', id: 1 },
    { login: 'testorg2', id: 2 }])

  test 'returns success' do
    get :new
    assert_response :success
  end

  test 'has a new organization' do
    get :new
    assert_not_nil assigns(:organization)
  end

  test 'has an array of github organizations' do
    get :new
    assert_equal [['testorg1', ['testorg1', 1]], ['testorg2', ['testorg2', 2]]],
                 assigns(:users_github_organizations)
  end
```

test/services/github_client_test.rb
johndbritton: Looks good to me.

johndbritton:
The GIF looks like you're on the right track, pretty exciting.

app/controllers/organizations_controller.rb
tarebyte: > Any user on the owners team can access the classroom using the classroom app

This makes sense to me as well.

app/controllers/organizations_controller.rb
tarebyte: I'll change this to `title`, and let the user set it. That way I don't have to worry about updating it.

app/models/organization.rb
tarebyte: I can do that, Rails wouldn't let me create a model with the same name as the app. I'll update it to `GithubClassroom`, but I'm open to suggestions for other names.

app/views/organizations/show.html.erb
tarebyte: As or right now all it does is remove the `Classroom` from the DB and remove the relation to the `User`. But that could change in the future. I'm going to move this into the `User` settings page and give some more context.

app/models/organization.rb
johndbritton: I would suggest `Class`, but that's already taken too. It's true that the hardest part of writing software is naming things well.

You could also keep this as organization and use that term in the interface. The Classroom app manages your Organizations.

app/views/organizations/show.html.erb
johndbritton: Another thing to think about, what happens when a user that's associated with a classroom is removed from the owners team? This is another place where you might need to worry about synchronization or getting clever about how you model things.

tarebyte: Would it be possible to add some sort of hook on to the organization to fire an event so that the app could update who has permissions?

tarebyte:
@johndbritton just a heads up I asked about the membership API here: https://github.com/octokit/octokit.rb/issues/590, so that's a good start, if it turns out this is plausible that will be able to cut down my API calls significantly.

As it stands right now using `.com` as the source of truth I make an API call for almost every action on the `OrganizationsController`.  It's not terribly slow, but I notice the speed difference. I'm going to get some metrics to see what the time actually is for each page request.

I'm also going to play around with caching to see how that affects the performance as well. Let me know what you think of the changes.

johndbritton:
> I'm going to get some metrics to see what the time actually is for each page request.

You might want to take a look at https://github.com/peek/peek for displaying request info.

tarebyte:
@johndbritton this is the current state of the app.

add-classroom

I've added issues under the [Classroom Milestone](https://github.com/education/classroom/milestones/Classrooms) just to keep track of what still needs to be done at some point in time.

I would like to start pushing forward on the invitation of users.

app/services/github_client.rb
```diff
@@ -0,0 +1,29 @@
+class GithubClient
+  def initialize(token)
+    @token = token
+  end
+
+  def is_organization_admin?(github_id)
```
johndbritton: I'd recommend renaming to `#organization_admin?`

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+  def set_users_github_organizations
+    @users_github_organizations = current_user.github_client.users_organizations.
+                                  collect { |org| [org.login, org.id] }
+  end
+
+  def update_organization_params
+    params.require(:organization).permit(:title)
+  end
+
+  def verify_is_organization_admin
```
johndbritton: 'is' in a method name usually implies returning a boolean, maybe call this `#ensure_organization_admin`

app/views/organizations/_remove_organization_modal.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      </h3>
+    </div>
+
+    <div class="modal-warning text-center">
+      <p>Please read this before continuing!</p>
+    </div>
+
+    <div class="modal-body">
+      <p>
+        This will not affect your organization on GitHub, however this will remove access for all other users using this org
```
johndbritton: s/affect/effect

johndbritton:
> this is the current state of the app.

Looking good.

> I would like to start pushing forward on the invitation of users.

Onward :fast_forward:

PR #28, opened 2015-06-11
Use Rubocop in Classroom

tarebyte:
I thought it might be nice to have Rubocop setup in Classroom. The end goal is to hook it up to Travis CI.


app/controllers/organizations_controller.rb
```diff
@@ -68,8 +62,9 @@ def set_organization
   end
 
   def set_users_github_organizations
-    @users_github_organizations = current_user.github_client.users_organizations.
-                                  collect { |org| [org.login, org.id] }
+    @users_github_organizations = current_user.github_client.organization_memberships.map do |org|
```
tarebyte: This change allows users to only see the organization they are an admin of at the time of creation. Need to add a test and fix the existing tests for this.

Issue #14, opened 2015-05-20
Caching API Requests


tarebyte:
Currently we aren't caching any of the requests, when I deployed the application to Heroku I didn't notice any performance lags.

However this is something that should be looked into more, as the application grows and we use more end points.

tarebyte:
Done

PR #40, opened 2015-06-20
Permissions and API Caching

tarebyte:
This is an experiment I'm running to see if I can cleanup the way permissions work in Classroom. Also since most of the permissions are API based I thought it might also be a good idea to look at caching.

This uses:
* [Pundit](https://github.com/elabs/pundit) for authorization
* [ActiveSupport::Cache::MemCacheStore](http://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memcachestore) and [Faraday HttpCache](https://github.com/plataformatec/faraday-http-cache) for caching the API requests from Octokit.


tarebyte:
@johndbritton this works pretty well, I'd like to get your opinion on it.

I'm spun up an app with this PR, it's sitting here: https://classroom-development-40.herokuapp.com/

tarebyte:
Closes #14

johndbritton:
Looks like a reasonable approach to me. Keep in mind that there may be some requests you don't want to cache to avoid UX issues. Is there an easy way for you to make those skip the cacheing?

tarebyte:
Just by taking a quick look at the docs no, so I raise an issue just to make sure https://github.com/plataformatec/faraday-http-cache/issues/75.

If not we could just take down the cache to 1 or 2 minutes so that if someone is clicking around the app it's fine but it won't be an issue when you need fresh data.

tarebyte:
@johndbritton I think I've turned off caching for `organization_memberships` going to deploy, and give it a shot

config/environments/production.rb
```diff
@@ -55,7 +55,12 @@
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :mem_cache_store, ENV["MEMCACHEDCLOUD_SERVERS"].split(','), { username: ENV["MEMCACHEDCLOUD_USERNAME"],
+                                                                                     password: ENV["MEMCACHEDCLOUD_PASSWORD"],
+                                                                                     namespace: 'classroom',
+                                                                                     expires_in: 15.minutes,
```
tarebyte: This seems too long, what do you think about 5 minutes? Or even less than that. Maybe 2?

config/initializers/octokit.rb
```diff
@@ -0,0 +1,7 @@
+stack = Faraday::RackBuilder.new do |builder|
+  builder.response :logger
```
tarebyte: I'm going to look at the server logs and see if this is something I want turned on.

config/environments/production.rb
johndbritton: 5 minutes sounds good. I'd set it using something like this:

`expires_in: ENV["REQUEST_CACHE_TIMEOUT"].to_i.minutes || 5.minutes`

tarebyte:
Looking over this again I want to separate this into two PRs.

Issue #51, opened 2015-06-24
Copy invitation doesn't work


johndbritton:
The clipboard link on the invitation link doesn't copy the url to the clipboard.

tarebyte:
I'd hate to add flash to the site, but what do you think about using [zeroclipboard](https://github.com/zeroclipboard/zeroclipboard)?

johndbritton:
Word on the street is that this is possible without flash using `document.execCommand('copy')` in Chrome and maybe [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1012662) too.

I'd go for zeroclipboard and maybe see if they plan to update the lib to avoid flash usage.

tarebyte:
There is a giant thread talking about just that https://github.com/zeroclipboard/zeroclipboard/issues/171

johndbritton:
This comment has the solution from github.com: https://github.com/zeroclipboard/zeroclipboard/issues/171#issuecomment-111130212

I'd say just go ahead with zeroclipboard. We can always upgrade once they make it work with the new hotness.

PR #52, opened 2015-06-24
Add ZeroClipboard

tarebyte:
It works, but I could use some advice on how to not make it jumpy. I tried making a gif but it's too fast for it to show up.

Who would be a good person to ask about that. Someone on the primer team?

/cc @johndbritton


tarebyte:
closes #51

tarebyte:
zerclipboard-primer

PR #56, opened 2015-06-27
Add a GitHub login helper to the user

tarebyte:


spec/models/user_spec.rb
```diff
@@ -25,4 +25,19 @@
...
+
+  describe '#github_client' do
+    it 'sets or creates a new GitHubClient with the users token' do
+      expect(user.github_client.class).to eql(GitHubClient)
+    end
+  end
+
+  describe '#github_login' do
+    it 'gets the users GitHub login' do
+      VCR.use_cassette('user_self') do
```
kdaigle: Generally, I find cassette names to be most valuable when they mimic the class/method being tested. What about something like 'User#github_login`? If you're going to have multiple tests per method, you could add a clarification at the end: `User#github_login:user_not_found`.

tarebyte: That's a great suggestion, I've updated the cassette to reflect this.

PR #57, opened 2015-06-29
Add `test` to the script folder

tarebyte:
To run all of the tests and the linters, you can now run `script/test`. This is also set for the Travis CI configuration in the `.travis.yml` file.

I kept forgetting to run these tests, so I thought it would be nice to have them in one script. And then if I ever update how the tests are run, it will automatically updates on Travis


johndbritton:
http://githubengineering.com/scripts-to-rule-them-all/

PR #65, opened 2015-07-10
Caching Octokit Responses

tarebyte:
From the Octokit README

> Once configured, the middleware will store responses in cache based on ETag fingerprint and serve those back up for future 304 responses for the same resource. See the [project README](https://github.com/plataformatec/faraday-http-cache) for advanced usage.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,8 @@ gem 'autoprefixer-rails'
 
 gem 'coffee-rails', '~> 4.1.0'
 
+gem 'faraday-http-cache'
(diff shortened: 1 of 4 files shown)


tarebyte:
@johndbritton @kdaigle do you have any thoughts on this?

johndbritton:
I don't have any experience with this, but it looks like a pretty straightforward solution that should work.

If you're using ETags I guess there's no case when you wouldn't want to cache a request as in the previous implementation.

PR #66, opened 2015-07-13
Push Starter Code for Assignments and GroupAssignments

tarebyte:
This gives the user the option to push starter code when the user accepts goes to the invitation url.

push_starter_code

/cc @johndbritton


tarebyte:
This still needs tests, but I wanted an initial look over

tarebyte:
Tested it on staging, good to go. @johndbritton if you have any suggestions or concerns let me know! :grin:

johndbritton:
:zap::zap::zap::zap::zap: Nice!

johndbritton:
Looks good. How do you handle a failed push? Does the student have any way to trigger the push again if it doesn't work the first time around?

tarebyte:
As of right now, they can refresh the page and it will trigger the process again.

tarebyte:
There should probably be a better way to do it though.

PR #68, opened 2015-07-20
Classroom User management

tarebyte:
:construction: Not quite ready yet :construction: 

This changes the way the user management happens on Classroom. Currently we check to see if the user is an admin against github.com on almost every request which makes the app :snail: 

All routes that are apart of the Organization are now scoped so that if you are not apart of the `Classroom` and you try and access it you won't be able to view anything.


tarebyte:
@johndbritton this is where it's at right now, if the user has a public GitHub email address it fills in the form and disables it. Otherwise the user has to fill in an email in order to invite the user.

invite_users

PR #73, opened 2015-08-02
User management for classrooms

tarebyte:
Similar to #68, with some modifications.

This PR:
* Allows users to be invited to classrooms via email if they are an org admin
* Introduces route constraints so that only a user who is apart of the organization can see the organization (or a user who is a staff member)
* Adds Staff routing constraints so that only users who are staff can see certain routes
* Adds Sidekiq to monitor background jobs (Staff Route)
* Other various refactors that seemed useful


config/environments/production.rb
```diff
@@ -89,4 +89,17 @@
                               { username: ENV['MEMCACHEDCLOUD_USERNAME'],
                                 password: ENV['MEMCACHEDCLOUD_PASSWORD'] }),
   }
+
+  config.action_mailer.default_url_options = { host: 'classroom-staging.herokuapp.com' }
```
tarebyte: this will change once we know where we are deploying this too :grin:

PR #74, opened 2015-08-04
Create GroupAssignments for pre existing groups

tarebyte:
If a teacher chooses a grouping that already exists, then we should go ahead and create the `GroupAssignmentRepos` for those groups. This is called in a background job.

@johndbritton any suggestions on better testing?


spec/jobs/create_group_assignment_repos_job_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it 'does not create any GroupAssignmentRepos' do
+      assert_performed_with(job: CreateGroupAssignmentReposJob, args: [group_assignment.id]) do
+        CreateGroupAssignmentReposJob.perform_later(group_assignment.id)
+      end
+
+      expect(GroupAssignmentRepo.all.count).to eql(0)
+    end
+  end
+
+  # context 'a GroupAssignment with an existing Grouping with Groups', :vcr do
```
tarebyte: So this tests passes locally if the cassette is not already written, otherwise the test fails.

spec/jobs/create_group_assignment_repos_job_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it 'does not create any GroupAssignmentRepos' do
+      assert_performed_with(job: CreateGroupAssignmentReposJob, args: [group_assignment.id]) do
+        CreateGroupAssignmentReposJob.perform_later(group_assignment.id)
+      end
+
+      expect(GroupAssignmentRepo.all.count).to eql(0)
+    end
+  end
+
+  context 'a GroupAssignment with an existing Grouping with Groups', :vcr do
```
tarebyte: So this tests passes locally if the cassette is not already written, otherwise the test fails.

app/models/concerns/git_hub_repoable.rb
```diff
@@ -42,8 +42,8 @@ def destroy_github_repository
   def push_starter_code
     return true unless starter_code_repo_id
 
-    repository              = GitHubRepository.new(organization.github_client, github_repo_id)
-    starter_code_repository = GitHubRepository.new(organization.github_client, starter_code_repo_id)
+    repository              = GitHubRepository.new(creator.github_client, github_repo_id)
+    starter_code_repository = GitHubRepository.new(creator.github_client, starter_code_repo_id)
```
tarebyte: This is to make sure that the owner of the repository is the token that we are using, just in case they use a personal private repo.

PR #89, opened 2015-08-09
Stop users from removing each other

tarebyte:
For an organization, there are times when removing a user should not be
possible.

* When there is only one active user left
* When a user has created an assignment

Using the before_remove callback we can check for this and stop the
transaction if one of these things happen.

The front end for this will come in another PR.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -9,12 +9,10 @@ class Organization < ActiveRecord::Base
   has_many :group_assignments, dependent: :destroy
   has_many :repo_accesses,     dependent: :destroy
 
-  has_and_belongs_to_many :users
+  has_and_belongs_to_many :users, presence: true, before_remove: [:verify_not_assignment_creator, :verify_not_last_user]
 
   validates :github_id, presence: true, uniqueness: true
   validates :title,     presence: true, uniqueness: { case_sensitive: false }
-
-  after_save :validate_minimum_number_of_users
 
   # Public
   #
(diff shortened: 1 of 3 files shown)


johndbritton:
Is this still relevant?

tarebyte:
@johndbritton yes it is, but the UI is not in place yet.

tarebyte:
I don't know if it is essential to ship it out with the first release though, thoughts?

johndbritton:
What's the use case for deleting users? It doesn't seem necessary to me.

Organization admins are auto added, so if they visit the app again they'll just get access. No idea why you'd want to delete a student, since they only get access to their own (or their group) assignments.

tarebyte:
@johndbritton the use case I see is that if you want to remove the TA's from Classroom after the semester is over but you want to continue using the Organization.

johndbritton:
Wouldn't it make sense to remove them as admins from the org?

If a user loses admin access they should lose access to classroom.

tarebyte:
Unless we do a check on every request we have no way of knowing if they are an admin after we have added them to Classroom.

Thoughts?

johndbritton:
Maybe there's a webhook that fires when a user is deleted from an organization.

You should go ahead with adding user management if you think that's the best way, don't let me hold you up. I was just trying to save some time.

tarebyte:
@johndbritton I'm going to table this for now, and come back to it at another time.

PR #97, opened 2015-08-16
Give users the ability to edit and destroy their assignments and group assignments

tarebyte:
Seems like a big oversight on my part.

* Change the name of the Assignment or Group Assignment
* Destroy it and all of repos that go with it.


tarebyte:
I'm going to also go ahead and the reset feature to the assignment, that way the teacher does't have to add it back in if they don't want to. This will also regenerate the assignment invitation.

app/views/assignments/edit.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    <div class="columns">
+      <div class="two-thirds column">
+        <%= render 'shared/error_messages', object: f.object %>
+        <%= render partial: 'assignments/assignment_form_options', locals: { f: f } %>
+      </div>
+
+      <div class="one-third column dangerzone-container">
+        <p class="text-emphasized">Delete this assignment</p>
+        <p class="text-muted">Once you delete this assignment, there is no going back. Please be certain.</p>
+        <p class="text-muted"><strong><%= pluralize(@assignment.assignment_repos.count, 'student assignment') %></strong> will be delete from github.com.</p>
```
johndbritton: > N repositories will be deleted from GitHub

app/views/assignments/edit.html.erb
```diff
@@ -0,0 +1,36 @@
...
+
+        <%= link_to 'Delete this assignment', [@organization, @assignment], class: 'btn btn-danger btn-block', method: :delete, data: { confirm: 'Are you positive that you want to delete your assignment?' } %>
```
johndbritton: > Are you sure you want to delete this assignment and it's N repositories?

PR #108, opened 2015-08-20
Change logout route to be post

### LINKED ISSUE #102: Logout should be a POST request
The `/logout` endpoint should require a POST request. Currently, someone could CSRF this endpoint by including an image like `` on another domain to log users out. This is quite low risk, but is an easy fix.

tarebyte:
Fixes #102

/cc @mastahyeti


mastahyeti:
:+1:

PR #109, opened 2015-08-20
Opt out of being logged as opposed to opting in

### LINKED ISSUE #103: Add ensure_logged_in before filter to ApplicationController
For an app in which most endpoints require authentication, it's a good idea to check authentication in the `ApplicationController` and have unauthenticated actions opt-out via `skip_before_filter`. This makes it harder to forget to add the authentication check when adding a new controller.

tarebyte:
Fixes #103 

/cc @mastahyeti


mastahyeti:
:+1:

Issue #111, opened 2015-08-20
eval() is a code smell


mastahyeti:
While it isn't a vulnerability in this case, `eval()` gives security people heart palpitations. Instead of

```ruby
invitation_path = eval("#{invitation.class.name.underscore}_path('#{invitation.key}')")
```

you could do

```ruby
invitation_path = send("#{invitation.class.name.underscore}_path", invitation.key)
```

or better yet

```ruby
invitation_path = case invitation
  when Assignment
    assignment_path invitation.key
  when GroupAssignment
    group_assignment_path invitation.key
end
```

Issue #112, opened 2015-08-20
Bit.ly discloses all assignment URLs


mastahyeti:
By taking one of the Bit.ly links generated by the application, I was able to use the Bit.ly [link info API](http://dev.bitly.com/links.html#v3_info) to find out whose account created the Bit.ly link. From there I was able to use the [user link history API](http://dev.bitly.com/user_info.html#v3_user_link_history) to find a listing of Bit.ly links generated by this user. This gives me the invitation URLs for all assignments.

johndbritton:
I'd suggest removing Bitly and just using the standard urls.

If you want to use a shortener, we've got `git.io`

PR #115, opened 2015-08-21
Remove use of eval, and bit.ly

tarebyte:
Found a much easier way for what I wanted to do, and I remove bit.ly in the hope of finding a better way in the future to shorten the urls.


tarebyte:
closes #111 #112

mastahyeti:
:+1:

Issue #117, opened 2015-08-21
Consider rack-canonical-host


johndbritton:
https://github.com/tylerhunt/rack-canonical-host

Require SSL and redirect all domains to one official domain.

PR #118, opened 2015-08-21
Add rack-canonical-host

tarebyte:
Also remove the bitly gem due to the fact I missed it before

To set this up on production

```
heroku config:add CANONICAL_HOST=classroom.github.com
```

/cc @johndbritton


tarebyte:
Closes #117

Issue #110, opened 2015-08-20
Add pagination to listing pages


johndbritton:
It looks like the listing pages don't have any pagination right now. It's probably fine at the moment, but once we see some larger classes or teachers using it semester after semester, it will become necessary.

tarebyte:
See #119

PR #101, opened 2015-08-19
New site design (WIP)

jasonlong:
This lays down the basics of the new styles proposed in https://github.com/education/classroom/issues/49#issuecomment-132236899. I'm hoping this gives @tarebyte the building blocks needed to get things wrapped up.

There were a couple of places I didn't get to yet (e.g. the assignment details page with the list of names and URLs comes to mind). I'm happy to jump back in and help wrap up these loose ends.

I also tweaked some copy here and there to better match our GitHub voice, but I didn't make a full audit or anything.

Here are a handful of screenshots.

screen shot 2015-08-19 at 9 24 09 am

screen shot 2015-08-19 at 9 58 00 am

screen shot 2015-08-19 at 11 02 58 am

screen shot 2015-08-19 at 11 20 45 am

screen shot 2015-08-19 at 1 32 59 pm

screen shot 2015-08-19 at 1 49 35 pm

screen shot 2015-08-19 at 1 48 38 pm

screen shot 2015-08-19 at 2 24 26 pm

Note the bit.ly links here are placeholders.

/cc @johndbritton 
/fyi @cameronmcefee @jglovier

PR #119, opened 2015-08-22
Add Pagination

tarebyte:
Right now this will be on the users organizations at `/dashboard`, and organizations assignments listing.

Needs styling, that can happen in #101 

/cc @johndbritton


tarebyte:
Closes #110

Issue #114, opened 2015-08-20
CSRF in AssignmentInvitationsController#accept_invitation


mastahyeti:
`AssignmentInvitationsController#accept_invitation` uses a GET requests. This should be a POST. Otherwise, an attacker could post an image like

```html
<img src="https://myclassroomapp.com/assignment_invitations/f772b522f9622307106fb91b7afe745d/accept_invitation">
```

on another site, forcing every Classroom user visiting the page to accept the invitation.

tarebyte:
See #120

PR #120, opened 2015-08-22
Require confirmation for accepting AssignmentInvitations

### LINKED ISSUE #113: AssignmentInvitationsController#show should require confirmation
Browsing to `AssignmentInvitationsController#show` fires JavaScript that accepts the invitation. It would be good to require some user interaction. Otherwise, any use can be sent to the URL and will be added to the org, added to the team, etc...

tarebyte:
When the user went to the page it automatically added them to the org
without a confirmation. This changes that so that the user has to click
on the accept button.

The styling still needs some :heart:, but the functionality is there.

I have also changed it so that when the user creates their personal team
instead of it being "Team 1" it just sets it to be there GitHub Login.

Closes #113 #114 

/cc @johndbritton @mastahyeti


tarebyte:
Just for a visual reference.

assignment_invitations

PR #127, opened 2015-08-24
Allow GitHub Staff members to be staff on Classroom for GitHub

tarebyte:
If the user is a staff member (the `site_admin` column) on GitHub then when they login they will
become staff on Classroom for GitHub.

Or if their uid is included in the NON_STAFF_GITHUB_ADMIN_IDS string
then they will also be updated to become an admin on sign in (for people like me :grin:)

/cc @johndbritton

### DIFF
--- a/app/services/auth_hash.rb
+++ b/app/services/auth_hash.rb
@@ -36,7 +36,7 @@ class AuthHash
 
   def non_staff_github_admins_ids
     return [] unless ENV['NON_STAFF_GITHUB_ADMIN_IDS'].present?
-    ENV['NON_STAFF_GITHUB_ADMIN_IDS'].split(',')
+    ENV['NON_STAFF_GITHUB_ADMIN_IDS'].split(',').compact.delete_if(&:empty?)
   end
 
   def raw_info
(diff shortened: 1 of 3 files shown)


tarebyte:
In your .env file it would looking something like this

`NON_STAFF_GITHUB_ADMIN_IDS=8675309,42`

app/services/auth_hash.rb
```diff
@@ -26,10 +27,16 @@ def token
...
+    raw_info[:site_admin]
+  end
+
   private
 
-  def info
-    user_hash.fetch('info', {})
+  def non_staff_github_admins_ids
+    return [] unless ENV['NON_STAFF_GITHUB_ADMIN_IDS'].present?
+    ENV['NON_STAFF_GITHUB_ADMIN_IDS'].split(',')
```
johndbritton: You probably wants a `compact` here to make sure `nil` never ends up in that array. I'd also add a test case to make sure setting the environment variable to `,,1` does not include a nil in the resulting array.

spec/services/auth_hash_spec.rb
```diff
@@ -1,8 +1,28 @@
...
+  let(:auth_hash) { AuthHash.new(OmniAuth.config.mock_auth[:github]) }
+
+  describe '#extract_user_info' do
+    it 'extracts the users information' do
+      expect(auth_hash.user_info[:token]).to eq('some-token')
+    end
+  end
+
+  describe '#non_staff_github_admins_ids' do
+    it 'returns an empty array if there the ENV is not set' do
```
johndbritton: typo :smile:

PR #129, opened 2015-08-24
Add sidekiq-peek to the status bar with link to the web path

### LINKED ISSUE #123: Move sidekiq link to peek bar
Having a link to Sidkiq is cool, but I'd prefer if we could keep all staff only links out of the main application. Can you add it to the peek bar instead?

screen shot 2015-08-24 at 3 28 42 pm

Here's what I did in another app:

screen shot 2015-08-24 at 3 30 20 pm

tarebyte:
Closes #123 

This adds sidekiq-peek to the top bar, and adds the link so that you can go to the webpath from the bar as well.

/cc @johndbritton for review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ gem 'peek-gc'
 gem 'peek-git'
 gem 'peek-performance_bar'
 gem 'peek-pg'
-gem 'peek-sidekiq', github: 'tarebyte/peek-sidekiq', branch: 'sidekiq-webpath-link'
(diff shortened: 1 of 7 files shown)


johndbritton:
Didn't look at your branch, but guessing you added the link there. You can actually override the view in this app by creating a file to replace the view in the correct folder within this app without modifying the gem.

tarebyte:
TIL, thanks!

johndbritton:
That's why we're here. :school_satchel:

johndbritton:
I think it's thanks to the magic of rails engines.

Issue #107, opened 2015-08-20
Org invitations are a spam vector


mastahyeti:
Organization invitations can be sent to any email address. The content of the invitation can be hijacked by modifying the Organization title, which can be quite long and even contain newlines:

*Creating an org with newlines in the title:*

```http
POST /orgs/free-viagra HTTP/1.1

utf8=%E2%9C%93&_method=patch&authenticity_token=[redacted]&organization%5Btitle%5D=Free+Viagra%21%21%21%21+https%3A%2F%2Fsketchy.site%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d%0a%0d&commit=Update+Organization
```

*Sending spam to any email address:*

```http
POST /orgs/free-viagra/invite-users HTTP/1.1

utf8=%E2%9C%93&_method=patch&authenticity_token=[redacted]&github_owners%5Bmastahyeti%5D=1144197&github_owner_emails%5Bmastahyeti%5D=[redacted]@gmail.com&commit=Invite
```

*Spam!!!*

<img width="903" alt="screen shot 2015-08-19 at 5 58 09 pm" src="https://cloud.githubusercontent.com/assets/1144197/9372459/e624e622-469b-11e5-96fd-05e20ae36b55.png">

This one is kind of hard to solve. Some options are:

1. Only send emails to addresses listed on the invited user's GitHub profile.
2. Implement some kind of rate limiting and spam detection.

tarebyte:
@mastahyeti number 1 works for me, it'll be annoying for users, but if they don't have a public facing GitHub email address there isn't a whole lot we can do.

Thoughts? /cc @johndbritton

johndbritton:
I think that's fine, as the organization invites are only for inviting teachers and admins.

@tarebyte is there a way for an org admin to access classroom without being invited?

tarebyte:
We could have it so anyone who is an admin who signs in to classroom automatically gets added to the org on Classroom

johndbritton:
> We could have it so anyone who is an admin who signs in to classroom automatically gets added to the org on Classroom

Yeah, I think that makes sense.

[...]

PR #132, opened 2015-08-25
Authorize users to to be added and access organizations

### LINKED ISSUE #105: OrganizationAuthorizedConstraint doesn't work
`OrganizationAuthorizedConstraint` searches the URL path for the first set of numeric characters and tries to look up an org with an id matching these digits. If no organization is found the check fails open. The `OrganizationsController#set_organization` method uses `Organization.find(params[:id])` to lookup organizations. Because of FriendlyId, this can take a slug instead of a numeric id. The result is that authorization is not checked for URLs using the org slug instead of a numeric ID.

The authorization check should be moved to the controller level and the organization should be looked up in the same way as elsewhere in the controller. The check should also fail closed if the organization can't be found.

tarebyte:
Closes #105 

This is the implementation that was discussed in #107.

When a user goes to an organization page on classroom if they are not already apart of it, Classroom will try to verify that they are in fact an organization admin. If they are it will add them to the org, then take them to the org page.

This is where I get unsure, right now when a user goes to an Assignment or GroupAssignment page if they aren't part of the org it will just 404. However this can be easily changed so that it will add the new user to the org if they go to any of the pages dealing with the org.

Thoughts?

/cc @johndbritton @mastahyeti


johndbritton:
> However this can be easily changed so that it will add the new user to the org if they go to any of the pages dealing with the org.

I don't see a problem with checking auth for any url under `/organization/ORG` and automatically adding the user if they have they are admin on the org.

app/controllers/organization_authorized_controller.rb
```diff
@@ -0,0 +1,19 @@
+class OrganizationAuthorizedController < ApplicationController
+  before_action :set_organization, :authorize_organization_access
+
+  decorates_assigned :organization
+
+  private
+
+  def authorize_organization_access
+    not_found if !@organization.users.include?(current_user) && current_user.staff?
```
mastahyeti: This looks like it only checks authorization if the user is staff. Maybe

```ruby
not_found unless @organization.users.include?(current_user) || current_user.staff?
```

johndbritton: @mastahyeti beat me to it, I had a comment box open for that line and was trying to parse it mentally.

tarebyte: @mastahyeti @johndbritton this logic went away with my latest commit

app/controllers/assignment_invitations_controller.rb
```diff
@@ -1,4 +1,6 @@
 class AssignmentInvitationsController < InvitationsController
+  skip_before_action :set_organization, :authorize_organization_access
```
johndbritton: :+1: Good move switching to an opt-out here.

tarebyte: This is actually leftover code from another idea I was spiking out. `set_organization` never actually happens in this controller.

johndbritton: Never mind me then.

tarebyte:
@johndbritton this is what the invite page for the organization looks like now. Any feedback on the copy would be much appreciated.

screen shot 2015-08-27 at 11 04 26

johndbritton:
Looks good. The copy could use some information about why someone would want to add another admin to classroom. Maybe explain a bit that admins are able to create assignments and get lists of submissions by students.

johndbritton:
If this is going to be an interstitial page in adding an org, I think the "Skip for now" button should be moved down to the bottom of the page to fit a more normal UX pattern.

cc/ @jasonlong

tarebyte:
@johndbritton something like this?

screen shot 2015-08-27 at 13 04 25

tarebyte:
screen shot 2015-08-27 at 13 10 56

johndbritton:
:+1: Works for me, we might want to revise at some point in the future but this is good.

PR #149, opened 2015-08-26
Move starter code cloning and pushing to a background job

### LINKED ISSUE #94: Creating new team causes application error
Visiting an assignment URL for a group assignment with starter code of `rails/rails` gives an application error after trying to join a new team.

URL: https://classroom-staging.herokuapp.com/group_assignment_invitations/5ba3f9f295292f846c56bcd72bc9ba56/accept_invitation

Don't really have any more insight into the issue.

tarebyte:
Closes #94

If the user has a really large repo, the application has the potential to time out while performing the action.

This moves the process out to a background job on it's own queue `starter_code` which has a higher weight than the default queue so that the assignments will be cloned and pushed faster.

The copy and style still needs some work, but I think that's for another PR.

/cc @johndbritton

### DIFF
--- a/app/jobs/push_starter_code_job.rb
+++ b/app/jobs/push_starter_code_job.rb
@@ -1,0 +1,10 @@
+class PushStarterCodeJob < ActiveJob::Base
+  queue_as :starter_code
+
+  def perform(creator, assignment_repo_id, starter_code_repo_id)
+    assignment_repository   = GitHubRepository.new(creator.github_client, assignment_repo_id)
+    starter_code_repository = GitHubRepository.new(creator.github_client, starter_code_repo_id)
+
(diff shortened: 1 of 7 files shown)


app/jobs/push_starter_code_job.rb
```diff
@@ -0,0 +1,10 @@
+class PushStarterCodeJob < ActiveJob::Base
+  queue_as :starter_code
+
+  def perform(creator, assignment_repo_id, starter_code_repo_id)
```
johndbritton: I usually try to call a single method on my model from `perform` where possible. That way you've got all of your logic in one place, you can call it synchronously if you need to, the job is only responsible for doing things in the background, and your tests are with the rest of your model code.

PR #154, opened 2015-08-27
Revisiting Invitation Pages after accepting the invitation

### LINKED ISSUE #143: Using a group invitation twice
If a user accepts a group assignment invitation and then goes back to the invitation page they are presented with the form to choose a team or create a new team but upon submitting, they're directed to the repo they created in the first pass.

I think we should probably just tell them that they already have a team / repo and direct them there without showing the form.

### LINKED ISSUE #144: Using an existing group of teams for a new group assignment
Say I create a group of teams for students to work in and an assignment to go with it.

The students self-select into teams and then submit their assignments. I create a second assignment and choose the same group of teams and pass out the invite link.

Students still see the form on the invitation page prompting them to create a team or choose one. Since I decided to use the same group of teams as before, the students should not get a choice and should go directly into the teams they were in in the previous assignment, unless they haven't yet participated.

tarebyte:
This closes #144 and closes #143.

When a student accepts an invitation to an assignment individual or group it will redirect them as expected.

But if they try to accept that same invitation again, it should automatically show them their assignment URL.

Looking for code feedback, and thoughts.

/cc @johndbritton


tarebyte:
This is what it currently looks like when a student accepts an individual assignment.

assignment_invitations

johndbritton:
The flow looks right to me. Planning to tackle the design issues separately?

johndbritton:
I think this resolves #143, but I don't know if it fully addresses #144. Since this is a new assignment, you might have the case where the group member previously accepted assignment A and formed teams, but has not accepted assignment B (which is using the same group of teams).

tarebyte:
May as well tackle the design here as well.

tarebyte:
@johndbritton how does this look for the individual assignment invitation page?

screen shot 2015-08-30 at 21 38 26

johndbritton:
@tarebyte I think you should include the full app chrome with "Classroom for GitHub," organization name, and assignment name heading sections and the info and actions inside that:



For the copy, how about this:

> Accepting this assignment will give you access to the `repository-name` repository in the `organization-name` organization on GitHub.

johndbritton:
I don't think it's necessary to explain that it will be created if it hasn't already been and that they will be given access if they haven't been. Telling them the end result should be clear enough.

tarebyte:
@johndbritton better?

screen shot 2015-08-31 at 10 48 06

screen shot 2015-08-31 at 10 48 55

johndbritton:
:heart: Looks great.

tarebyte:
@johndbritton merging this for now, shout if you have any feedback!

PR #158, opened 2015-08-29
Better Mobile Layout

### LINKED ISSUE #157: Mobile layout
We don't need anything fancy, but the app should work on mobile.

tarebyte:
Fixes #157 

This is to make Classroom for GitHub usable on mobile.

/cc @johndbritton 

It still needs a little work, but this is what I have so far

ios simulator screen shot aug 29 2015 10 39 31

ios simulator screen shot aug 29 2015 10 39 58


johndbritton:
Good progress. I think it would be ok to drop the display of invitation link text and just use a simple click to copy icon. You could replace the settings button with a gear icon or drop settings from mobile all together.

tarebyte:
@johndbritton I removed the invitation link from mobile for now. This can be put back in later, we really don't have a way for the copy button to work because we use flash :crying_cat_face:

tarebyte:
@johndbritton merging for now, shout if you have any other questions!

PR #174, opened 2015-09-03
Move creating of invitations and grouping to the front

### LINKED ISSUE #170: Use a spinner when waiting for an invite link
I've seen this case a couple times:

screen shot 2015-09-02 at 4 16 56 pm

> Hang tight your students invitation is on the way!

It'd be better to wait or display a spinner so that the link is immediately available.

tarebyte:
Closes #170 

We shouldn’t let these objects be created in the background.

This makes the invitations be created as an ActiveRecord callback, and
the Grouping in a Transaction

Looking for implementation feedback

/cc @johndbritton


app/services/group_assignment_service.rb
```diff
@@ -0,0 +1,26 @@
...
+    group_assignment = GroupAssignment.new(@new_group_assignment_params)
+
+    group_assignment.transaction do
+      if grouping.new_record?
+        grouping.save!
+        group_assignment.grouping = grouping
+      end
+
+      group_assignment.save!
+    end
```
johndbritton: Doesn't rails automatically do a transaction if you do this:

```ruby
group_assignment.grouping = grouping
group_assignment.save!
```

app/services/group_assignment_service.rb
```diff
@@ -0,0 +1,26 @@
...
+    @new_grouping_params         = new_grouping_params
+  end
+
+  # rubocop:disable MethodLength
+  def create_group_assignment
+    grouping = Grouping.find_by(id: @new_group_assignment_params[:grouping_id])
+    grouping = Grouping.new(@new_grouping_params) unless grouping.present?
+
+    group_assignment = GroupAssignment.new(@new_group_assignment_params)
+
```
johndbritton: It'd be really cool if you could do something like this:

```ruby
group_assignment = GroupAssignment.new(@new_group_assignment_params)
group_assignment.grouping = Grouping.where(id: @new_group_assignment_params[:grouping_id]).first_or_create(@new_group_params)
group_assignment.save!
```

johndbritton:
I think this is a good case for using the `build_*` methods. You can call `assignment.build_invitation` when you create the assignment and then both should be created when you persist the assignment.

app/services/group_assignment_service.rb
tarebyte: That would be pretty nice

app/controllers/assignments_controller.rb
```diff
@@ -7,7 +7,6 @@ class AssignmentsController < ApplicationController
   decorates_assigned :organization
   decorates_assigned :assignment
 
-  rescue_from ActiveRecord::RecordInvalid, with: :error
```
tarebyte: Pro tip never override this

tarebyte:
@johndbritton merging this for now, shout if you find any issues!

PR #183, opened 2015-09-05
Warn the users that they cannot change teams on classroom

### LINKED ISSUE #175: Users cannot change teams
Accepting a group assignment and choosing to create a new team or join an existing team is a permanent choice.

What should a participant do if they accidentally choose the wrong team. Is there any way to undo / correct a mistake?

If not, maybe we should warn them when they accept the assignment.

tarebyte:
Closes #175 

/cc @johndbritton


johndbritton:
:+1:

PR #192, opened 2015-09-10
Query strings begin with `?`

johndbritton:
Serves me right for showing up here and trying to actually do something. I'll stick to commentary from now on @tarebyte.


tarebyte:
:boom:

PR #204, opened 2015-09-11
Don't over cache API calls

### LINKED ISSUE #202: Org-list cache bug
Alright, this time I've got a real cache bug:

screen shot 2015-09-11 at 10 26 50 am

In four different browsers I'm logged in as four different users. @mastahyeti is the only one with access to the orgs listed in that screenshot, but the `/organizations/new` endpoint shows those orgs for all the users.

tarebyte:
Hopefully fixes #202 make sure that we aren't caching the users organizations when they create a new organization.

/cc @mastahyeti


app/models/git_hub_user.rb
```diff
@@ -15,8 +15,12 @@ def login
   # Public
   #
   def admin_organization_memberships
+    no_cache_headers = { 'Cache-Control' => 'no-cache, no-store' }
```
mastahyeti: Isn't that a server cache header? I assumed the problem was with the Octokit caching going on over in https://github.com/education/classroom/blob/master/config/initializers/octokit.rb#L2

tarebyte: > I assumed the problem was with the Octokit caching going on over in https://github.com/education/classroom/blob/master/config/initializers/octokit.rb#L2

It does deal with Octokit, the reason why I added a special header here is because most of the time we want `shared_cache: false` because requests against users and organization are private so normally they won't cache.

Is this not advisable?

mastahyeti: That sounds right, but I don't think the useragent can send the `Cache-Control` header. That's a header typically sent by the server to instruct the *useragent* not to cache the response. I'm not sure that the code change here does anything, but I could be totally wrong.

tarebyte: @mastahyeti this is what the log looks like on `master` with a clear cache:

```ruby
>> github_user.admin_organization_memberships
HTTP Cache: [GET /user/memberships/orgs?per_page=100&state=active] miss, store
```

This is what it looks like hitting it a second time.
```ruby
>> github_user.admin_organization_memberships
HTTP Cache: [GET /user/memberships/orgs?per_page=100&state=active] fresh
```

This is what it looks like with the new changes
```ruby
>> github_user.admin_organization_memberships
HTTP Cache: [GET /user/memberships/orgs?per_page=100&state=active] unacceptable
```

According to https://github.com/plataformatec/faraday-http-cache#instrumentation `:unacceptable` means that the request did not go through the cache at all.

Does this look right?

mastahyeti: TIL the user agent can send the `Cache-Control` header. It looks like the middleware *is* [handling this correctly](https://github.com/plataformatec/faraday-http-cache/blob/master/lib/faraday/http_cache/request.rb#L23). I stand corrected.

mastahyeti:
This looks good. Do you think there might be other places where we should be setting this header?

tarebyte:
I'll peruse through the code add and see if this is relevant everywhere else.

tarebyte:
@mastahyeti I went through the code, and added the no cache for when we check the users plan in order to create private assignments

PR #213, opened 2015-09-15
Give a helpful name for the reusable set of teams

### LINKED ISSUE #201: Reusable set of teams
I always have a hard time naming my "reusable set of teams". What if we don't let the user provide a name and just create the first set for them. The second time through we can ask them if they want to use the same teams as before, or create a different set.

Maybe give the set a name based on date or something like that.

tarebyte:
Fixes #201 

/cc @johndbritton


johndbritton:
:+1:

PR #223, opened 2015-09-18
Validate the titles by the slugs they create

tarebyte:
@johndbritton I came across an instance on staging where the assignment will use the UUID as part of the slug due to duplication. This is expected, but not what I think we should have.

This removes the use of `friendly.find` because we have finders turned on for the models that we use.

What do you think?


johndbritton:
@tarebyte What will the new urls look like?

tarebyte:
@johndbritton the urls won't change, this just ensures that the objects UUID doesn't end up as part of the URL

johndbritton:
This looks ok to me, but I'm not really familiar with this gem.

PR #236, opened 2015-09-22
Add rails admin for Staff

tarebyte:
Also add the git-branch octicon to the branch name

/cc @johndbritton


johndbritton:
I think we can get by without `rails_admin`. If we find a true need for this, we can consider adding it in the future.

PR #248, opened 2015-09-22
Validate assignment title length <= 60

tjhorner:
Right now, assignment names appear to be unlimited (or how much the machine running it can handle :wink:), this causes problems:

I used 100 as repo names can only be up to 100 characters long -- but it would append the student's username when creating an assignment. Should it validate the max at something like 85-90 characters so it gives room for the student username?

If this PR goes through I can add tests to the CI if needed.


tarebyte:
@tjhorner thank you for the PR, this is a great point!

What do you think about writing a custom maximum length validation, we can get a users login or a teams slug through the GitHub API. We can make that API call, and them make sure the that the title + the login/slug doesn't exceed 100 characters.

tarebyte:
@tjhorner we should also probably set a limit with how long an `Organization` title can be, probably around the neighborhood of 20-30 characters since we are going taking the time to look at assignments.

tjhorner:
@tarebyte Sounds like a good idea, so what you're saying is when the user accepts an assignment it will verify that the repo can be created with the character limit, but what will it do when it exceeds 100 characters? Chop off part of the repo name?

johndbritton:
Yeah, we should definitely add some sort of validations here. I'd go for the simplest thing possible, so just limit the number of characters or potentially change the naming pattern used for assignments with long titles. You could generate a slug to be used as the repo name and set the conditions for that slug to have a character limit.

I'd avoid hitting the API for the validation if you can.

app/models/assignment.rb
```diff
@@ -21,6 +21,7 @@ class Assignment < ActiveRecord::Base
 
   validates :title, presence: true
   validates :title, uniqueness: { scope: :organization }
+  validates :title, length: { maximum: 100 }
```
johndbritton: Let's make this 60 for now. Maximum username length on GitHub is 39, plus a character for the separator. We can always increase the limit later if we want.

tjhorner: Changed in 15f63a5

johndbritton:
Works for me.

@tarebyte How do you think we should handle the existing cases of titles that are over the limit?

tarebyte:
@tjhorner for the constraints let's do the following.

* `Assignment` and `GroupAssignment` titles: 60
* `Group` titles: 39

That way we will never hit the 100 character limit. We already know that GitHub won't let the user login name be more than 39 so we don't have to worry about that for individual assignments.

tjhorner:
Sure thing!

tjhorner:
What's with the 39 character limit by the way, why not 40?

johndbritton:
> What's with the 39 character limit by the way, why not 40?

I have no idea, but that's what it is.

johndbritton:
@tjhorner This needs one more validation for organization title length.

tjhorner:
Ah, thought it was safe to assume that it would technically never happen since org names are fetched through the API.

Adding now

johndbritton:
@tjhorner Sorry for being picky, can you make the organization maximum `60`. That is the title of the organization, not the login. It's not used in the repository name construction, so it can be longer without issue.

Also, if you can merge master into your branch, that'd be awesome.

tjhorner:
That's totally fine, haha.

johndbritton:
Thanks!

PR #274, opened 2015-09-24
Update gems, remove sha lock from Octokit

tarebyte:
/cc @johndbritton


johndbritton:
That was a quick release.

PR #281, opened 2015-09-26
Clarify the difference between Organization and Classroom (WIP)

### LINKED ISSUE #250: What's an "organization" in Classroom?
I was baffled by this page on my first test of Classroom:

screen shot 2015-09-22 at 7 55 24 pm

It says "New Organization," and then there's a box for the name, and a dropdown that says "Organization".  Am I selecting which organization I'm adding an organization to?  Or is it just that the organization is new to Classroom?  Why am I giving it a name, then?

What I expected was new, standalone organization within which I could test out Classroom stuff.  What I got was a course thing within an existing organization (happened to be brownplt).  Then when I created some test student repos, folks in the brownplt organization started getting emails that they were subscribed to the student repos, etc, which wasn't what I expected.

I didn't expect that any existing organization I had would need to be part of the workflow.  I think that perhaps what I should have done is create an organization through Github proper, then use that organization with Classroom.  In that case, the answer might be "RTFM, @jpolitz, this is documented elsewhere, go look" – where should I have found that workflow, if so?

Is there a reason each new course isn't just a brand-new organization within Github?  Why did my existing organizations need to be involved?

### LINKED ISSUE #252: Changing "Organization" to "Classroom"
@johndbritton there seems to be a lot of confusion what the difference between an Organization and a Classroom are. I think that is due to the fact that we intermingle the terms.

We say it's Classroom for GitHub, but we never actually use the term Classroom anywhere else.

What do you think using classroom in lieu of organization throughout the app.

tarebyte:
This is to help clarify for users the difference between a GitHub organization and a Classroom on Classroom for GitHub.

Closes #250, closes #252 

/cc @johndbritton for opinions and copy

### DIFF
--- a/app/views/organizations/_invite_users.html.erb
+++ b/app/views/organizations/_invite_users.html.erb
@@ -19,7 +19,7 @@
           <span class="input-group-button">
             <button aria-label="Copy to clipboard" id=<%= @organization.slug %> class="js-zeroclipboard btn tooltipped tooltipped-s" data-copied-hint="Copied!" type="button">
               <%= octicon('clippy') %>
-              Copy organization link
+              Copy classroom link
             </button>
           </span>
         </div>
--- a/app/views/organizations/_reset_and_remove_organization_modal.html.erb
+++ b/app/views/organizations/_reset_and_remove_organization_modal.html.erb
@@ -9,7 +9,7 @@
   <div class="remodal-description">
     <p>
       This action <strong>CANNOT</strong> be undone.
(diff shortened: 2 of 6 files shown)


tarebyte:
Just to help show the changes I'm proposing

screen shot 2015-09-26 at 00 14 48

screen shot 2015-09-26 at 00 14 55

johndbritton:
:shipit: This is an improvement, with more to come from a deeper copy audit. No sense in holding back incremental improvements.

PR #302, opened 2015-10-02
Gracefully handle invalid starter code input

### LINKED ISSUE #299: Using full URL in 'starter code from GitHub' caused server error

tarebyte:
Fixes #299 

If GitHub the repository format is not valid, GitHub sends us back the `ArgumentError`, we should handle this gracefully.

/cc @johndbritton 

Also I will add tests, just wanted to get an opinion on implementation.


johndbritton:
This looks reasonable to me. You still need to sanitize the user input though.

app/controllers/concerns/starter_code.rb
```diff
@@ -1,8 +1,19 @@
...
+
+    if repo_name =~ /[a-zA-Z0-9_-]+\/[a-zA-Z0-9_-]+/i
+      begin
+        github_repository = GitHubRepository.new(current_user.github_client, nil)
+        github_repository.repository(repo_name).id
+      rescue ArgumentError => err
+        raise GitHub::Error, err.message
+      end
+    else
+      raise GitHub::Error, 'Invalid repository format'
```
johndbritton: > Invalid repository name, use the format `owner/name`

tarebyte:
@johndbritton should be good to go

johndbritton:
Looks good. I'd recommend a test that asserts a 200 response from the action even when bad input is provided.

app/controllers/concerns/starter_code.rb
```diff
@@ -1,8 +1,21 @@
...
+  extend ActiveSupport::Concern
+
+  # rubocop:disable MethodLength
   def starter_code_repository_id(repo_name)
     return unless repo_name.present?
-    sanitized_repo_name = repo_name.gsub(/\s+/, '')
-    github_repository   = GitHubRepository.new(current_user.github_client, nil)
-    github_repository.repository(sanitized_repo_name).id
+
+    if repo_name =~ %r{/[a-zA-Z0-9_-]+\/[a-zA-Z0-9_-]+/}
```
johndbritton: I think you might need begin `^` and end `$` anchors in this regex. As it stands I think it might match anything containing `owner/name`.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -25,11 +25,29 @@
   end
 
   describe 'POST #create', :vcr do
+    before do
+      request.env['HTTP_REFERER'] = 'http://test.host/classrooms/new'
```
johndbritton: Why do you need this referrer?

spec/controllers/assignments_controller_spec.rb
```diff
@@ -25,11 +25,29 @@
...
+      it 'fails to create a new Assignment' do
+        expect do
+          post :create,
+               organization_id: organization.id,
+               assignment:      attributes_for(:assignment),
+               repo_name:       'https://github.com/rails/rails'
+        end.not_to change { Assignment.count }
+
+        expect(response).not_to have_http_status(:internal_server_error)
+        expect(flash[:error]).to eql('Invalid repository name, use the format owner/name')
```
johndbritton: Generally I try to have one expectation per `it` block. You could rephrase this as three different cases:

* It does not create an assignment
* It does not return an http error
* It provides a friendly error message

spec/controllers/assignments_controller_spec.rb
tarebyte: So we use `redirect_to :back` in the controllers on errors, and the test doesn't know where to go back to. Setting the `request.env['HTTP_REFERER']` solves this problem.

app/controllers/concerns/starter_code.rb
tarebyte: Yup, just confirmed!

tarebyte:
@johndbritton recreated the scenario described in #299, and it now behaves as expected.

:ship: :dash:

tarebyte:
@johndbritton going to :ship: :dash: 

I also added a test where it will successfully create an Assignment with the correct format.

spec/controllers/assignments_controller_spec.rb
johndbritton: I see, thanks for the explanation.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -30,6 +30,42 @@
...
+        expect(Assignment.count).to eql(0)
+      end
+
+      it 'does not return an internal server error' do
+        expect(response).not_to have_http_status(:internal_server_error)
+      end
+
+      it 'provides a friendly error message' do
+        expect(flash[:error]).to eql('Invalid repository name, use the format owner/name')
+      end
```
johndbritton: These tests look much cleaner now, thanks for updating.

johndbritton:
:sparkles:

PR #310, opened 2015-10-06
Add application monitoring endpoint at `/_ping`

johndbritton:
This adds an endpoint at `/_ping` for application monitoring using [Pinglish](https://github.com/jbarnette/pinglish).

Provides checks for DB, Memcached, and Redis, disabled by default and configurable with `ENV['PINGLISH_ENABLED]`.

Also adds documentation to the readme and a blank var to the `.env-example` file.


johndbritton:
This is out and working in production. Any glaring omissions here @tarebyte?

tarebyte:
@johndbritton, does this gem have value in dev? If not we can move it to the production only group

johndbritton:
@tarebyte production only is fine, I've updated the `Gemfile`.

PR #312, opened 2015-10-07
Use collaborators instead of teams for organization members

### LINKED ISSUE #311: Private group assignment repositories can still be cloned by anyone in the classroom organization
Result is that student groups can easily see each other's code in even private repo assignments.  Bug?

tarebyte:
Because of the new [GitHub organization permissions](https://github.com/blog/2020-improved-organization-permissions) users can now be apart of an org without being apart of a team.

This PR removes the need for a single user to be on their own team, and if the user accepts an individual assignment, it will add them as a collaborator to the repo.

Fixes #311


johndbritton:
@tarebyte how do you feel about doing the same for group assignments, I think it'd be an improvement as there is no chance that the team could be deleted.

We could even go so far as to rely on the collaborators list as the source of truth for who is in a group.

tarebyte:
@johndbritton do a lot of organizations enjoy using @team mentions and do they find it useful to keep their students in teams?

Also we lose the ability to reuse teams, not sure if we want students to remake their teams everytime.

Thoughts?

tarebyte:
@johndbritton if @team mentions don't matter, we could just store the "teams" on Classroom let the user select from the dropdown and then just add them as a collaborator.

johndbritton:
> do a lot of organizations enjoy using @team mentions and do they find it useful to keep their students in teams for better organization?

I think the team @mentions are useful to have, but we've seen a number of cases where people edit or delete the teams not knowing the negative consequences it will have in Classroom.

> Also we lose the ability to reuse teams for group assignments, not sure if we want students to rejoin their "teams" everytime.

> @johndbritton if @team mentions don't matter, we could just store the "teams" on Classroom let the user select from the dropdown and then just add them as a collaborator.

Yes, we could still have reusable groups by storing the data in classroom. Perhaps that's the way forward here, store the group info on classroom, use said info to make members into repository collaborators, and optionally create teams on GitHub for @mentions.

tarebyte:
> optionally create teams on GitHub for @mentions.

How do you see that happening? Would that be when you create the group assignment?

johndbritton:
We can shelve the optional teams for now and come back to it if we want later.

app/models/assignment_repo.rb
```diff
@@ -6,14 +6,36 @@ class AssignmentRepo < ActiveRecord::Base
...
+
+  before_destroy :destroy_github_repository
+
+  # Public
+  #
+  def add_user_as_collaborator
+    github_user = GitHubUser.new(user.github_client, user.uid)
+    repository  = GitHubRepository.new(organization.github_client, github_repo_id)
+
+    begin
```
johndbritton: This is the same pattern we used with pushing starter code. Is there a way to de-duplicate and share the behavior?

app/models/concerns/git_hub_repoable.rb
```diff
@@ -1,32 +1,11 @@
...
-  included do
-    before_validation(on: :create) do
-      create_github_repository if organization
-    end
-
-    before_create :add_team_to_github_repository
-    before_create :push_starter_code
-
-    before_destroy :destroy_github_repository
-  end
```
johndbritton: What's your reasoning behind removing this. I think this code still exists directly on the class, so it's more just being moved from the module, right?

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -44,17 +44,11 @@
...
-      RepoAccess.destroy_all
     end
 
     it 'redeems the users invitation' do
       patch :accept_invitation, id: invitation.key
-
-      assert_requested :post, github_url("/organizations/#{organization.github_id}/teams")
-      assert_requested :post, github_url("/organizations/#{organization.github_id}/repos")
-
       expect(assignment.assignment_repos.count).to eql(1)
```
johndbritton: Does it make sense to use `expect(user.assignment_repos.count).to eq(1)` now?

johndbritton:
This is coming along nicely :sparkles:

app/models/concerns/git_hub_repoable.rb
tarebyte: I want to keep the methods in the module, but they called differently for `AssignmentRepo` and `GroupAssignmentRepo`

spec/controllers/assignment_invitations_controller_spec.rb
tarebyte: I think so, it verifies that not only we successfully created the `AssignmentRepo` we created it for the right user.

app/models/concerns/git_hub_repoable.rb
johndbritton: :cool:

tarebyte:
@johndbritton when you get a chance this could use some :eyes: I just want to make sure I didn't break anything.

I will also be updating the tests soon to reflect the changes

app/models/assignment_repo.rb
```diff
@@ -6,14 +6,22 @@ class AssignmentRepo < ActiveRecord::Base
...
   validates :github_repo_id, presence:   true
   validates :github_repo_id, uniqueness: true
 
-  validates :repo_access, presence:   true
-  validates :repo_access, uniqueness: { scope: :assignment }
+  before_validation(on: :create) do
+    if organization
+      create_github_repository
+      push_starter_code
+      add_user_as_collaborator
```
johndbritton: If `create_github_repository` succeeds but `push_starter_code` fails, we need to remove the repository from GitHub. I think this is handled correctly here since we're getting these methods from the `GitHubRepoable` module, right?

app/models/assignment_repo.rb
```diff
@@ -30,13 +38,13 @@ def private?
   # Public
   #
   def github_team_id
-    repo_access.github_team_id
+    repo_access.present? ? repo_access.github_team_id : nil
```
johndbritton: I cringe at returning `nil` in cases like this. Is there a case when you are calling this method and expecting `nil`? If not, I'd raise an exception here so it goes into our logs and we can fix those calls.

If there are cases where you're expecting `nil`, you'd be better served by returning an instance of a specific class.

johndbritton:
The use of `delete` and `destroy` in method names is inconsistent. My suggestion is to update all method names dealing with `ActiveRecord` to use `destroy` and anything to do with API requests to use `delete`.

Totally open to other patterns, but we should have some consistency.

app/models/assignment_repo.rb
johndbritton: Also, this looks like code for backwards compatibility. If that's the case and we'll :fire: it soon after we ship this, then I think it's ok.

app/models/git_hub_organization.rb
```diff
@@ -85,6 +91,16 @@ def plan
     end
   end
 
+  def remove_organization_member(github_user_id)
+    github_user_login = GitHubUser.new(@client, github_user_id).login
+
+    with_error_handling do
+      @client.remove_organization_member(@id,
+                                         github_user_login,
+                                         headers: { 'Accept' => 'application/vnd.github.ironman-preview+json' })
```
johndbritton: can we put those headers in a helper method?

app/models/git_hub_repository.rb
```diff
@@ -10,6 +10,14 @@ def initialize(client, id)
 
   # Public
   #
+  def add_collaborator(collaborator)
+    with_error_handling do
+      @client.add_collaborator(@id, collaborator, headers: { accept: 'application/vnd.github.barred-rock-preview' })
```
johndbritton: same here

app/models/repo_access.rb
```diff
@@ -19,29 +16,40 @@ class RepoAccess < ActiveRecord::Base
 
   before_validation(on: :create) do
     if organization
-      add_member_to_github_team
-      accept_membership_to_organization
+      add_membership_to_github_organization
+      accept_membership_to_github_organization
```
johndbritton: Is there ever a case where we should handle a failing `accept_membership_to_github_organization`?

johndbritton: For example, should we remove the user's organization membership and show them an error to try again?

johndbritton: If this code is going to be :fire: soon, it might not be worth handling that edge case, but if this will stick around we should probably handle it.

db/migrate/20151009033035_add_user_to_assignment_repo.rb
```diff
@@ -0,0 +1,6 @@
+class AddUserToAssignmentRepo < ActiveRecord::Migration
+  def change
+    add_column :assignment_repos, :user_id, :integer
```
johndbritton: It looks like you're modifying the DB schema in a way that the code in this pull request won't work correctly without having the migrations run.

If that's the case, users will see errors in production when you deploy this code until the migrations have been run.

As a workaround, you can pull your db migrations out into another pull request that we can deploy first without the functionality you're adding in this code. Then once we run the migrations and merge that pull request you can easily deploy this code without users ever seeing errors.

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -83,7 +77,7 @@
       it 'removes the repository on GitHub' do
         patch :accept_invitation, id: invitation.key
 
-        assert_requested :delete, github_url('/repositories/43894381')
+        assert_requested :delete, github_url('/repositories/44262216')
```
johndbritton: I really wish we didn't have to do this. ref: https://github.com/education/classroom/pull/316#discussion_r41522602

johndbritton:
Left some comments regarding style. Tried this branch out locally and everything seems to be working.

Let me know if you need help pulling the migrations to another PR.

app/models/assignment_repo.rb
tarebyte: Yup, that's exactly right

tarebyte: For reference: https://github.com/education/classroom/blob/collaborators/app/models/concerns/git_hub_repoable.rb#L54

app/models/assignment_repo.rb
tarebyte: Yes this is for backwards compatibility

app/models/repo_access.rb
tarebyte: We should handle this, this is a change that is hanging around for now

tobiasmayrtfo:
Hello, was this changed again later? Because when my students join a Classroom Team, they are added as member of the github organization, and if I change them to outside collaborators on github, they are removed from the Classroom Team again.

Is this intended? Having them as members of the organization is not ideal, as the students tend to create organization-wide projects and repositories ;-)

johndbritton:
As I remember it, group assignments still use teams, but individual assignments use outside collaborators.

vinobstin:
keep me tip note for nroadcast, how to integrated  profile organisation

PR #316, opened 2015-10-07
Move code import to GitHub Importer

### LINKED ISSUE #283: Stuck `PushStarterCodeJob` in Sidekiq
There's a `PushStarterCodeJob` stuck in Sidekiq with the error `GitHub::NotFound: Resource could not be found on github.com`.

We should gracefully handle losing access to the starter code repository due to permissions or repository deletion.

johndbritton:
This removes the `PushStarterCode` background job and replaces it with GitHub Importer.

No more failed background jobs and as an added benefit, the user will see a progress bar while the code is being imported into their repository.

Resolves #283, as the job is :fire:

screen shot 2015-10-07 at 6 45 25 pm


app/models/git_hub_repository.rb
```diff
@@ -17,14 +17,12 @@ def full_name
...
   def get_starter_code_from(source)
-    Dir.mktmpdir do |dir|
-      Dir.chdir(dir) do
-        system("git clone --bare https://#{@client.access_token}@github.com/#{source}.git")
-        Dir.chdir("#{source.split('/').last}.git") do
-          system("git push --mirror https://#{@client.access_token}@github.com/#{full_name}.git")
-        end
-      end
-    end
+    @client.put(
```
johndbritton: @tarebyte Should this go inside of a `with_error_handling` block?

johndbritton:
Since this is happening in a `before_action` I'm concerned about failure states. Here are the possibilities:

* Repo creation fails
* Repo creation succeeds, call to importer fails
* Repo creation succeeds, call to importer succeeds

The happy case works fine and the case where repo creation fails (for example a repo already exists with this name) we don't create the AR object.

The middle case, where repo creation succeeds, but the call to the importer fails is not good. The importer fail will currently stop the record from being created in the DB, even though the repo exists on GitHub. Then, if the user tries to accept the invitation again, it will fail because the repo can't be created (already exists).

I'm open to suggestions on how to resolve that problem.

app/models/git_hub_repository.rb
tarebyte: Yes, most definitely

tarebyte:
@johndbritton what do you think about creating a kind of API transaction, where is one part fails the model isn't created.

For the `AssignmentRepo` and `GroupAssignmentRepo` if the repo fails to be created the models won't be created.

If we make the import process part of the model creation we handle delete the repository and then the model won't be created.

tarebyte:
@johndbritton I took a stab at this my transaction idea, let me know what you think.

This will definitely need tests.

johndbritton:
@tarebyte I like it, this is pretty much what I had in mind as well.

app/models/concerns/git_hub_repoable.rb
```diff
@@ -43,7 +45,18 @@ def destroy_github_repository
...
+
+    client = creator.github_client
+
+    assignment_repository   = GitHubRepository.new(client, github_repo_id)
+    starter_code_repository = GitHubRepository.new(client, starter_code_repo_id)
+
+    begin
+      assignment_repository.get_starter_code_from(starter_code_repository)
+    rescue
+      destroy_github_repository
```
johndbritton: @tarebyte what if we get an exception here?

We should probably do something smart like tell the user what happened and ask them to report a bug or give them a way to fix the situation.

app/models/concerns/git_hub_repoable.rb
```diff
@@ -43,7 +45,18 @@ def destroy_github_repository
...
-    PushStarterCodeJob.perform_later(creator, github_repo_id, starter_code_repo_id)
+
+    client = creator.github_client
+
+    assignment_repository   = GitHubRepository.new(client, github_repo_id)
+    starter_code_repository = GitHubRepository.new(client, starter_code_repo_id)
+
+    begin
+      assignment_repository.get_starter_code_from(starter_code_repository)
+    rescue
```
johndbritton: I don't really like to `rescue` without specifying the type of exception. Maybe this should be `rescue GitHub::Error` and we should make all of the other error classes inherit from `GitHub::Error`

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -56,5 +58,33 @@
...
+      before do
+        allow_any_instance_of(GitHubRepository)
+          .to receive(:get_starter_code_from)
+          .and_raise(GitHub::Error)
+      end
+
+      it 'removes the repository on GitHub' do
+        patch :accept_invitation, id: invitation.key
+
+        assert_requested :delete, github_url('/repositories/43894381')
```
johndbritton: This controller test is making sure that if the API call to GitHub Importer fails that we delete the repository we created on GitHub so we don't leave the user in an inconsistent state (a repository on GitHub without a corresponding record in Classroom).

In this case, the repository on GitHub was successfully completed and a non-persisted `ActiveRecord` object has the `github_repository_id` in one of it's attributes, but because this test is in a different process, I don't think I have any way to access that ID until it has been persisted.

Since we're using VCR, I ran the tests once and then pulled the ID out of the cassette and hard coded it here so that I can safely assert that the correct `DELETE` request is being made.

I'd really love to get rid of this hard coded id so that if someone needs to make a new cassette in the future they won't have to know to run it and then grab the ID manually and update this.

Anyone have any ideas on how I can improve this?

cc/ @jonmagic

johndbritton: I'm going to merge this, but still interested in better solutions so I can fix this up.

mastahyeti:
With this change, it seems like the student needs to have access to the starter-repo. Otherwise, they get a Importer error asking for credentials when they try to go to their repo.

spec/controllers/assignment_invitations_controller_spec.rb
johndbritton: @bkeepers do you have any suggestions on how to get around hard coding this value in the spec?

johndbritton:
> With this change, it seems like the student needs to have access to the starter-repo. Otherwise, they get a Importer error asking for credentials when they try to go to their repo.

@mastahyeti This was resolved in https://github.com/education/classroom/pull/323. Private starter code should work as long as the assignment creator has access to the repository.

mastahyeti:
Yep. I've been following along. Thanks for getting the fix together so quickly!

spec/controllers/assignment_invitations_controller_spec.rb
bkeepers: I think you can pass a regex to assert_requested. Maybe something like this?

```ruby
assert_requested %r{\A#{github_url('/repositories')}/\d+\z}
```

johndbritton: Interesting, that's better than manually updating this.

> In this case, the repository on GitHub was successfully completed and a non-persisted ActiveRecord object has the github_repository_id in one of it's attributes, but because this test is in a different process, I don't think I have any way to access that ID until it has been persisted.

Is that explanation correct? No way to grab the actual repository id?

PR #318, opened 2015-10-08
Silent destroy callbacks

johndbritton:
This makes versions of GitHub resource destructors that always return true as to not cause `ActiveRecord` transactions to fail.

This way, we can still find out of the destroy worked or not if we need to.


tarebyte:
LGTM, :ship: :dash:

PR #320, opened 2015-10-08
Remove starter_code background job queue

tarebyte:
Since we don't push code in the background anymore (#316) we don't need the queue.

/cc @johndbritton


johndbritton:
:shipit:

PR #325, opened 2015-10-15
Set migrations up for new collaborator changes

tarebyte:
These are the migrations needed for #312. These changes are to help facilitate

* Create `RepoAccesses` without `github_team_ids` we can't require them to be present
* Add the `user_id` column onto `AssignmentRepos` because `AssignmentRepos` use outside collaborators.

/cc @johndbritton


johndbritton:
Ran the migrations in production,

PR #326, opened 2015-10-15
Report exceptions to Airbrake

johndbritton:
This adds the `airbrake` gem to report exceptions.

We only send exceptions if the `AIRBRAKE_API_KEY` environment variable is set.


tarebyte:
:shipit:

Issue #263, opened 2015-09-23
Request fewer permissions on invitation


johndbritton:
When a student accepts an invitation, we might be able to request less access than we do currently. We need `repo` scope to create repositories, but we could get that from the teacher when they authorize.

PR #328, opened 2015-10-15
Spike for reducing student OAuth scopes

mastahyeti:
A quick spike to address https://github.com/education/classroom/issues/263. Does this look like a reasonable approach @johndbritton? Not sure how to test this. It also doesn't address the case of a student becoming an instructor.


mastahyeti:
Bummer... Looks like you need org-admin scope to [accept an invitation](https://github.com/education/classroom/blob/f1b3835fbfcbe5d77672f571e1032f3df24b082a/app/models/git_hub_organization.rb#L13).

johndbritton:
I think it makes sense. Once #312 ships we won't need to accept the invitation for individual assignments as they'll be added as outside collaborators.

mastahyeti:
Oh, neat!

johndbritton:
> Not sure how to test this.

I'd run it locally with the testing environment variables set and then actually do the oauth dance and record the interactions with VCR and make some assertions about what scopes you're getting.

johndbritton:
Check out https://github.com/education/classroom#testing-environment-variables for more info on that.

johndbritton:
@mastahyeti https://github.com/education/classroom/pull/312 has shipped, so we can reduce the needed scopes for students when they accept an individual assignment for now. Group assignments still require the ability to accept an organization membership invitation on behalf of the student since we're adding them to the org and a team for their group.

mastahyeti:
:cool: I can do some more work on this PR later this week, unless someone else wants to take it over. I think we'll want to track what states the user has in the session, so if they go from individial->group or student-> instructor, we can know that we need to send them back into the OAuth dance to get more scopes.

johndbritton:
We've got a few other bugs to fix and features prioritized ahead of this so if you've got the cycles / motivation we'd appreciate the contribution.

mastahyeti:
@johndbritton I got this PR fixed up and wrote some tests. Care to :eyes: it when you have a chance?

mastahyeti:
It works well locally fwiw. The auth upgrade when you switch from student->teacher or individual->group is smooth.

app/controllers/sessions_controller.rb
```diff
@@ -14,6 +17,9 @@ def create
     session[:user_id] = user.id
 
     url = session[:pre_login_destination] || organizations_path
+
+    session[:current_scopes] = user.github_client_scopes
+
     redirect_to url
   end
```
tarebyte: @mastahyeti can you add `# rubocop:enable Metrics/LineLength` at the end of the method? Otherwise it disables it for the rest of the file

mastahyeti: Fixed in a6753dd948483a43343076f728ccf93e83e36df4

tarebyte:
@mastahyeti thank you so much for this! I've also tested this locally as well and everything seems to work for me.

@johndbritton can we branch deploy this to try it out?

app/controllers/sessions_controller.rb
```diff
@@ -2,9 +2,12 @@ class SessionsController < ApplicationController
   skip_before_action :authenticate_user!, :set_organization, :authorize_organization_access
 
   def new
-    redirect_to '/auth/github'
+    scopes = session[:required_scopes] || 'user:email,repo,delete_repo,admin:org'
```
johndbritton: @mastahyeti Maybe we should put these scopes behind a method like `default_scopes` or `all_scopes` to make this a bit more clear.

There shouldn't be cases where `session[:required_scopes]` isn't set, so this seems to be just a backup for safety.

spec/support/vcr.rb
```diff
@@ -8,7 +8,7 @@
     serialize_with: :json,
     preserve_exact_body_bytes:  true,
     decode_compressed_response: true,
-    record: ENV['TRAVIS'] ? :none : :once
+    record: ENV['TRAVIS'] ? :none : :new_episodes
```
johndbritton: Do we actually want to change this VCR setting permanently?

Using `:new_episodes` means that we can inadvertently introduce new requests without getting a error/warning: https://relishapp.com/vcr/vcr/v/2-8-0/docs/record-modes/new-episodes

spec/integration/authentication_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
+
+RSpec.describe 'OAuth scope requirements', type: :request do
+  let(:organization) { GitHubFactory.create_owner_classroom_org }
+  let(:user)         { organization.users.first                 }
+
+  describe 'unauthenticated organization#show' do
```
johndbritton: Suggested modifications:

```ruby
describe 'organizations#show' do
  context 'unauthenticated request' do
    it 'redirects to sessions#new' do
      get url_for(organization)
      expect(response).to redirect_to(login_path)
    end

    it 'sets required scopes in session' do
      get url_for(organization)
      expect(session[:required_scopes]).to eq('user:email,repo,delete_repo,admin:org')
    end
  end

  context 'authenticated request' do
    before(:each) do
      get url_for(organization)
      get response.redirect_url # http://www.example.com/login
      get response.redirect_url # http://www.example.com/auth/github?scope=user%3Aemail%2Crepo%2Cdelete_repo%2Cadmin%3Aorg
      get response.redirect_url # http://www.example.com/auth/github/callback
    end

    it 'renders organizations#show' do
      get response.redirect_url
      expect(response.status).to eq(200)
      expect(response).to render_template('organizations/show')
    end
  end
end

describe 'sessions#new' do
  before(:each) do
    get url_for(organization)
  end

  it 'redirects to omniauth' do
    get response.redirect_url
    expect(response).to redirect_to('/auth/github?scope=user%3Aemail%2Crepo%2Cdelete_repo%2Cadmin%3Aorg')
  end
end

describe 'OAuth dance', :vcr do
  before(:each) do
    get url_for(organization)
    get response.redirect_url # http://www.example.com/login
    get response.redirect_url # http://www.example.com/auth/github?scope=user%3Aemail%2Crepo%2Cdelete_repo%2Cadmin%3Aorg
  end

  it 'redirects back to organizations#show' do
    get response.redirect_url # http://www.example.com/auth/github/callback
    expect(response).to redirect_to(url_for(organization))
  end
end
```

johndbritton:
This looks great @mastahyeti, thanks for the contributions. Left some inline notes to fix up, happy to deploy this once those few things are addressed.

johndbritton:
@mastahyeti Once this ships, we should consider updating existing authorizations to only have as few scopes as needed: https://developer.github.com/v3/oauth_authorizations/#update-an-existing-authorization

johndbritton:
Closing in favor of #341. Thanks @mastahyeti.

PR #330, opened 2015-10-20
fix NoMethodError for NullGitHubUser

### LINKED ISSUE #329: paginating to page2 does not work We're sorry, but something went wrong.
We're sorry, but something went wrong.
If you are the application owner check the logs for more information.

Airbrake Error 1534922897515050480

Some student did delete their account, maybe it broke something?

tarebyte:
Fixes #329 

/cc @johndbritton 

we definitely need tests for these at some point.

### DIFF
--- a/app/models/null_git_hub_user.rb
+++ b/app/models/null_git_hub_user.rb
@@ -6,4 +6,8 @@ class NullGitHubUser < NullGitHubResource
   def login
     'ghost'
   end
+
+  def name
+    'Deleted user'
+  end
 end


app/models/null_git_hub_user.rb
```diff
@@ -6,4 +6,8 @@ def html_url
   def login
     'ghost'
   end
+
+  def name
+    'ghost'
```
johndbritton: I prefer to use "Deleted user" here as that's the name that is associated with https://github.com/ghost and makes more sense when the user sees that.

tarebyte:
@johndbritton updated the PR based on your suggestion

PR #331, opened 2015-10-20
Migrate to Collab

tarebyte:
This PR creates a rake task that will move all users to collaborators on their assignments and then deleting the unnecessary teams.

/cc @johndbritton for :eyes:


lib/tasks/collab_migration.rake
```diff
@@ -0,0 +1,10 @@
+namespace :collab_migration do
+  task :migrate do
+    User.all.each do |user|
```
tarebyte: @johndbritton what would be the best way to break this up into chunks?

johndbritton:
This looks correct, but should really have some tests.

lib/tasks/collab_migration.rake
```diff
@@ -0,0 +1,14 @@
...
+task collab_migration: :environment do
+  User.find_in_batches(batch_size: 100) do |users|
+    users.each do |user|
+      user.repo_accesses.each do |repo_access|
+        collab_migrator = CollabMigration.new(repo_access)
+        collab_migrator.migrate
+      end
+    end
+
+    sleep(60)
```
johndbritton: Why are you `sleep`ing here?

PR #336, opened 2015-10-29
Replace zeroclipboard with clipboard.js

tarebyte:
I found https://github.com/zenorocha/clipboard.js the other day, and thought it was worth the time to migrate from using flash to a JavaScript only implementation of the clipboard.

This way more people can use the copy button.

/cc @johndbritton


johndbritton:
:+1:

PR #338, opened 2015-11-01
Create a new workflow for creating Classrooms

tarebyte:
As we had talked about in chat I'm working on a new flow for users to create their classrooms.

This allows the users to select their organization, then change the name of the Classroom if they want, which then takes them to the page to invite other users.

new classroom flow

This is still rough in terms of styling and wording, ~~it also doesn't have tests yet.~~


tarebyte:
This also fixes a bug where is the user changed the name of the classroom, the slug wouldn't update.

tarebyte:
This is what the select screen looks like after discussion, we now show all of the orgs and if they can't be added or have already been added we show the user.

disabled tooltips

app/views/organizations/_disabled_organization_select.html.erb
```diff
@@ -0,0 +1,6 @@
+<% aria_label = org[:role] != 'admin' ? 'You are not an admin of this org' : 'This org has already been added to Classroom' %>
```
johndbritton: > You are not an admin of this org

Make this more clear: "You cannot add this organization to classroom because you are not an owner"

Now the user will know to go get admin access from someone.

> This org has already been added to Classroom

What can the user do in this situation? Maybe better to say "Another owner has already added this organization to Classroom for GitHub"

app/controllers/organizations_controller.rb
```diff
@@ -1,8 +1,9 @@
 class OrganizationsController < ApplicationController
   include OrganizationAuthorization
 
-  before_action :authorize_organization_addition, only: [:create]
-  before_action :set_users_github_organizations,  only: [:new, :create]
+  before_action :authorize_organization_addition,     only: [:create]
+  before_action :set_users_github_organizations,      only: [:new, :create]
+  before_action :paginate_users_github_organizations, only: [:new, :create]
```
johndbritton: This is fine, since you've been using Rails callbacks for a lot of stuff in this app. Normally I'd do this in the controller action.

johndbritton:
This looks great, just suggested a couple copy tweaks.

PR #340, opened 2015-11-02
Add info box to alert users to change their permissions

### LINKED ISSUE #322: Private group assignment advisory
On the private group assignment page we need to add an info box telling the user to set their "Default repository permission" to "None" so that students can't see each other's assignments.

The info box should link to the exact settings page for that organization.

tarebyte:
Closes #322 

screen shot 2015-11-02 at 11 20 52 am

/cc @johndbritton

### DIFF
--- a/app/views/group_assignments/show.html.erb
+++ b/app/views/group_assignments/show.html.erb
@@ -1,4 +1,5 @@
 <%= render 'organizations/organization_banner' %>
+<%= render partial: 'permission_info_box' %>
 
 <div class="site-content">
   <div class="site-content-cap">
--- a/app/assets/javascripts/application.coffee
+++ b/app/assets/javascripts/application.coffee
@@ -26,4 +26,6 @@
 #
 #= require clipboard/dist/clipboard
 #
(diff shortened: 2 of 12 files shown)


johndbritton:
Is there a way for us to show this message conditionally depending on the default repository permission setting? If so, we should only show this in the case that it's not already set to "None."

app/views/group_assignments/show.html.erb
```diff
@@ -1,5 +1,10 @@
 <%= render 'organizations/organization_banner' %>
 
+<div class="flash flash-warn">
+  <%= octicon('broadcast') %>
+  Please go to your organization <%= link_to 'settings', "https://github.com/organizations/#{organization.login}/settings/member_privileges", target: "_blank" %> and update your <b>Default repository permission</b> to none.
```
johndbritton: We should tell the user why they need to do this.

tarebyte:
This is the new box and the working functionality.
permission box

johndbritton:
:+1: :shipit:

PR #343, opened 2015-11-03
Assignment/GroupAssignment slugs should only be unique to the organization

### LINKED ISSUE #342: Project names include some kind of key
Example "myFakeGroup/SuperAwseomeProject-34235-9160-4e50-ba243-545435-cool-team created by mattfxyz

tarebyte:
Closes #342 

~~This still needs tests~~

/cc @johndbritton @mattfxyz

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -2,7 +2,7 @@ class Assignment < ActiveRecord::Base
   include GitHubPlan
 
   extend FriendlyId
-  friendly_id :title, use: [:slugged, :finders]
+  friendly_id :slug_candidates, use: [:slugged, :finders]
 
   default_scope { where(deleted_at: nil) }
 
@@ -41,8 +41,19 @@ class Assignment < ActiveRecord::Base
 
   private
 
+  def organization_slug
+    organization.slug
+  end
+
   def should_generate_new_friendly_id?
     title_changed?
+  end
+
+  def slug_candidates
+    [
+      [:title],
+      [:title, :organization_slug]
+    ]
   end
 
   def uniqueness_of_title_across_organization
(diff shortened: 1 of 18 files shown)


app/models/assignment.rb
```diff
@@ -41,10 +41,21 @@ def starter_code?
...
+  end
+
   def should_generate_new_friendly_id?
     title_changed?
   end
 
+  def slug_candidates
+    [
+      [:title],
+      [:title, :organization_slug]
```
johndbritton: Why do we need to include the organization slug here?

johndbritton: Assignment names shouldn't need to be unique across the app.

tarebyte:
@johndbritton @mattfxyz I fixed the problem, I stripped `friendly_id` out of Classroom and we use our own [slugging method](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-parameterize)

I think this also makes the controllers more clear as to what we are looking for in our parameters now that we explicitly state we are using the slugs.

johndbritton:
:+1: I like this a lot better.

tarebyte:
@johndbritton I added some tests to verify that slugs will be unique when necessary. I think this just needs one more :eyes: and then we can :shipit:

spec/models/assignment_spec.rb
```diff
@@ -3,6 +3,17 @@
 RSpec.describe Assignment, type: :model do
   it_behaves_like 'a default scope where deleted_at is not present'
 
+  describe 'slug uniqueness' do
+    let(:organization) { create(:organization) }
+
+    it 'verifes that the slug is unique even if the titles are unique' do
```
tarebyte: @johndbritton this tests seems to satisfy your test case we were discussing.

tarebyte:
@johndbritton this is what the uniqueness check looks like on the form
uniqueness

tarebyte:
This is what title presence validation looks like

presence

johndbritton:
:shipit:

Are you going to follow up with the ID based routing for organizations in another pull request?

tarebyte:
Yup!

PR #354, opened 2015-11-04
Gracefully handle when a student rejects their authentication token

tarebyte:
If a student rejects their auth token it raises a `Octokit::Unauthorized` error which we now handle.

/cc @johndbritton


johndbritton:
:+1:

Issue #356, opened 2015-11-04
Add stafftools for management


johndbritton:
Let's add a `StafftoolsController` with restricted access for managing the application. Here's a wishlist of features (we won't necessarily add them all):
- [x] Fake login (#359)
- [x] Search for all app models (#380)
- [ ] Statistics
- [ ] Link to exceptions in airbrake for current user
- [x] User activity (last access / login)
- [ ] Staff notes / comments for all resources
- [x] Enabled features when we get Flipper (#474)

tarebyte:
@johndbritton what kind of statistics would be most helpful?

johndbritton:
We might want to use this as a project tracking issue and have a separate issue for each of the main features in stafftools.

Counts for all resources:
* Users
* Organizations
* Assignments
* GroupAssignments
* Groupings
* Groups

It'd be great if we could show a visual representation of the growth of these resources on a chart or graph of some sort

johndbritton:
There should be a resource overview for each of the models as well that gives a full diagnostic overview of the resource, for example the user overview would include data on related models.

You might consider building on top of `rails_admin` or similar, but I'm not sure if we need the full set of tools included in that library.

johndbritton:
@tarebyte How do you feel about breaking out the last few remaining tasks here to individual issues with a bit more explanation of what they should be. Maybe with more info we can attract some community contributions.

tarebyte:
@johndbritton that sounds fine to me, do you want me to tackle it?

johndbritton:
Sure, if you could write a short issue for each and close this one that'd be :100:

PR #359, opened 2015-11-06
Create Stafftools with impersonate user capabilities

tarebyte:
This is the start of #356 currently this allows users with staff privilege to impersonate themselves as another user for support purposes.

This is the general idea of how it works right now (there is a pause because the js alert wouldn't work with licecap.

impersonate

/cc @mastahyeti if you have time to give this a look over for :lock: reasons or if you have advice on how to improve this implementation it would be really appreciated.


app/controllers/application_controller.rb
```diff
@@ -50,8 +52,22 @@ def decorated_current_user
...
+
   def current_user
-    @current_user ||= User.find(session[:user_id]) if session[:user_id]
+    user_id = session[:impersonated_user_id] || session[:user_id]
+
+    if @current_user.present?
+      if @current_user.id != user_id
+        @current_user = User.find(user_id)
+      else
+        @current_user ||= User.find(user_id)
```
mastahyeti: Why is this line needed? In this case `@current_user.id` is already correct, right?

app/controllers/application_controller.rb
```diff
@@ -50,8 +52,22 @@ def decorated_current_user
...
+      end
+    else
+      @current_user ||= User.find(session[:user_id]) if session[:user_id]
+    end
```
mastahyeti: I think the logic in this method could be simplified a lot.

```ruby
def current_user
  return @current_user if defined?(@current_user)
  
  @current_user = if true_user.try(:admin?) && session[:impersonated_user_id]
    User.find session[:impersonated_user_id]
  else
    true_user
  end
end
```

app/controllers/application_controller.rb
```diff
@@ -84,4 +100,8 @@ def redirect_to_root
   def staff?
     logged_in? && current_user.staff?
   end
+
+  def true_user
+    @true_user ||= User.find(session[:user_id]) if session[:user_id]
```
mastahyeti: Using a `defined?` check on `@true_user` instead of `||=` would save some AR calls.

app/controllers/stafftools_controller.rb
```diff
@@ -0,0 +1,7 @@
+class StafftoolsController < ApplicationController
```
mastahyeti: I might have missed it elsewhere, but shouldn't this have some before-filter checking that the user is staff?

config/routes.rb
```diff
@@ -43,4 +43,17 @@
...
+
+  namespace :stafftools, constraints: StaffConstraint.new do
+    root to: :impersonate
+  end
+
+  scope module: 'stafftools' do
+    resource :users, only: [] do
+      member do
+        post :impersonate, constraints: StaffConstraint.new
+        get  :stop_impersonating
```
mastahyeti: This should be a POST as well.

app/controllers/stafftools_controller.rb
tarebyte: I set it as a route constraint here: https://github.com/education/classroom/pull/359/files#diff-21497849d8f00507c9c8dcaf6288b136R47 would it be more clear if these checks were on the controller itself?

config/routes.rb
```diff
@@ -43,4 +43,17 @@
       resources :group_assignments, path: 'group-assignments'
     end
   end
+
+  namespace :stafftools, constraints: StaffConstraint.new do
```
mastahyeti: Ah, I knew I was missing the authorization check :smile:. I feel like these kind of checks are better in the controller than in the routes.

app/controllers/stafftools/users_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Stafftools
+  class UsersController < StafftoolsController
+    before_action :set_user
+
+    def impersonate
+      if @user
+        impersonate_user(@user)
```
mastahyeti: Where is `impersonate_user` defined?

app/controllers/stafftools/users_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+        redirect_to root_path
+      else
+        flash[:error] = 'This user does not exist on Classroom for GitHub'
+        redirect_to :back
+      end
+    end
+
+    def stop_impersonating
+      stop_impersonating_user
```
mastahyeti: Where is `stop_impersonating_user` defined?

tarebyte: https://github.com/ankane/pretender/blob/master/lib/pretender.rb#L39

app/controllers/stafftools/users_controller.rb
tarebyte: https://github.com/ankane/pretender/blob/master/lib/pretender.rb#L34

app/views/stafftools/impersonate.js.erb
```diff
@@ -0,0 +1 @@
+$('#js-user-search-results').html("<%= escape_javascript(render partial: 'stafftools/impersonate/user_results', locals: { users: @users }) %>");
```
mastahyeti: Dynamic JavaScript is kind of an antipatern. Also, this could be abused by another site. On attacker.com, if I have a script tag pointing at this URL and can trick a staff user into visiting my site, the results would get loaded on attacker.com and I could collect them.

Also, I don't see where this script is getting loaded.

app/controllers/stafftools/users_controller.rb
mastahyeti: Is a gem really needed for this? It seems like we just need to set `session[:impersonated_user_id]`.

app/controllers/stafftools_controller.rb
mastahyeti: Yeah, it's a personal-preference thing, but I definitely like seeing authorization checks at the controller level.

app/controllers/stafftools/users_controller.rb
tarebyte: probably not, it was an easy first grab

app/views/stafftools/impersonate.js.erb
tarebyte: it's not really loading any scripts all this is doing it replacing the partial with the new set of results.

This is the partial https://github.com/education/classroom/pull/359/files#diff-0eb563d2788c58b421c0979a6f4a4c04R1

tarebyte: What would be the proper way to do a partial page update?

mastahyeti: You can have an endpoint that renders the partial without a layout. In JavaScript you can GET that endpoint and update the DOM.

tarebyte:
@mastahyeti thank you for all of the feedback! I think I've made all of the changes you had suggested.

@johndbritton this still needs tests but let me know what you think.

tarebyte:
Here is how it looks right now, I think I'm going to create a modal to confirm instead of the `data-confirm` since turbolinks keeps double binding it
impersonate_again

app/controllers/stafftools_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+  end
+
+  private
+
+  def authorize_access
+    not_found unless current_user.try(:staff?)
+  end
+
+  def set_users
```
tarebyte: @johndbritton I'm open to suggestions on how to make this search better, although it seems to work pretty well right now.

app/controllers/stafftools_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    match_phrase_prefix = { match_phrase_prefix: { login: params[:query] } }
+    wildcard            = { wildcard:            { login: '*' }            }
+
+    user_query = params[:query].present? ? match_phrase_prefix : wildcard
+    @users = UsersIndex::User.query(user_query).page(params[:page]).per(10)
+  end
```
tarebyte: @johndbritton I'm open to suggestions on how to make this search better, although it seems to work pretty well right now.

app/views/shared/_header.html.erb
```diff
@@ -16,8 +16,17 @@
           </li>
 
           <li>
-            <%= link_to decorated_current_user.github_url, class: 'tooltipped tooltipped-s', 'aria-label' => 'View GitHub profile' do %>
-              <%= image_tag decorated_current_user.avatar_url(40), class: 'avatar', alt: "@#{decorated_current_user.login}", height: 20, width: 20 %>
+            <% if current_user != true_user %>
```
mastahyeti: Seems like `#impersonating?` would be a useful helper method.

config/routes.rb
```diff
@@ -1,10 +1,7 @@
-require 'staff_constraint'
-
 require 'sidekiq/web'
 
 Rails.application.routes.draw do
-  mount Peek::Railtie => '/peek',    constraints: StaffConstraint.new unless Rails.env.test?
-  mount Sidekiq::Web  => '/sidekiq', constraints: StaffConstraint.new
+  mount Peek::Railtie => '/peek'
```
mastahyeti: With the routing constraint removed, is this endpoint still being authenticated?

config/routes.rb
```diff
@@ -43,4 +40,20 @@
       resources :group_assignments, path: 'group-assignments'
     end
   end
+
+  namespace :stafftools do
+    mount Sidekiq::Web  => '/sidekiq'
```
mastahyeti: Same question here.

mastahyeti:
I posted a few more comments inline, but this is looking really good.

app/views/shared/_header.html.erb
tarebyte: that seems more clear to me as well, I'm going to go ahead and make that change.

config/routes.rb
tarebyte: @mastahyeti I have it set to staff only here: https://github.com/education/classroom/blob/master/app/controllers/application_controller.rb#L18

But I can put it under the `stafftools` namespace which does verify that the user is properly authenticated.

config/routes.rb
tarebyte: Everything under the `stafftools` namespace inherits from the stafftools controller.

This is the check: https://github.com/education/classroom/pull/359/files#diff-22dc66320c0cd133d26c6fa203093eecR4

config/routes.rb
mastahyeti: > @mastahyeti I have it set to staff only here: https://github.com/education/classroom/blob/master/app/controllers/application_controller.rb#L18

:cool: just double checking.

> But I can put it under the stafftools namespace which does verify that the user is properly authenticated.

How does that work, if it doesn't inherit from the `StafftoolsController` class?

config/routes.rb
mastahyeti: Really? This Rails magic is new to me.

config/routes.rb
tarebyte: > How does that work, if it doesn't inherit from the StafftoolsController class?

It doesn't, that's a :derp: on my part

config/routes.rb
tarebyte: It doesn't for some reason I thought it did, ignore me

tarebyte: @mastahyeti would it be a good idea just to put back the staff constraint?

mastahyeti: Hmmm. I'm really not a fan of specifying auth logic at the route level, but that seems to be the recommendation from Sidekiq: https://github.com/mperham/sidekiq/wiki/Monitoring#restful-authentication-or-sorcery

tarebyte: I saw that as well, if you think it's ok to use it for just `sidekiq` an nothing else I'm :cool: with that.

If not I can try and explore ways to not use the route constraint

mastahyeti: Yeah, let's use it for Sidekiq, but keep the controller logic where possible.

tarebyte: :+1:

johndbritton:
@mkcode Can you take a look at the usage of `elasticsearch` and `chewy` in here?

app/controllers/concerns/organization_authorization.rb
```diff
@@ -6,7 +6,7 @@ module OrganizationAuthorization
   end
 
   def authorize_organization_access
-    return if @organization.users.include?(current_user) || current_user.staff?
```
johndbritton: :heart: that we're removing this.

johndbritton:
I don't love that you're updating the `users` index from so many different models, is there a way we can keep that up to date without adding it in so many places?

config/routes.rb
johndbritton: @mastahyeti Out of curiosity, what's your reason for preferring controller level auth rather than route constraints?

app/controllers/stafftools/users_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        session[:impersonated_user_id] = @user.id
+        redirect_to root_path
+      else
+        flash[:error] = 'This user does not exist on Classroom for GitHub'
+        redirect_to :back
+      end
+    end
+
+    def stop_impersonating
+      not_found unless true_user.try(:staff?)
```
johndbritton: I know returning `404` helps with avoiding disclosure, but since this app is open source is there any reason not to just return unauthorized here?

app/controllers/stafftools_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  def users
+    respond_to do |format|
+      format.html { render partial: 'stafftools/impersonate/user_results', locals: { users: @users } }
+    end
+  end
+
+  private
+
+  def authorize_access
+    not_found unless current_user.try(:staff?)
```
johndbritton: Same question about disclosure here. Does it make sense to return unauthorized since this app is open source?

spec/controllers/stafftools/users_controller_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  describe 'POST #stop_impersonating', :vcr do
+    context 'as an unauthorized user' do
+      it 'returns a 404' do
+        expect { post :impersonate, id: student.id }.to raise_error(ActionController::RoutingError)
+      end
+    end
+
+    context 'as an authorized user' do
+      before do
+        user.update_attributes(site_admin: true)
```
johndbritton: You should also make sure to start impersonating a user here so that you can make sure the call to `#stop_impersonating` works.

Right now, I think these tests will pass even if you remove the implementation from the `#stop_impersonating` method.

spec/controllers/stafftools_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      before do
+        user.update_attributes(site_admin: true)
+      end
+
+      it 'returns a success status' do
+        get :impersonate
+        expect(response).to have_http_status(:success)
+      end
+
+      it 'as an array of user search results' do
```
johndbritton: typo?

spec/spec_helper.rb
```diff
@@ -91,4 +91,12 @@
   config.profile_examples = 10
 =end
   # rubocop:enable Style/BlockComments
+
+  def sign_in(user)
```
johndbritton: Glad to have these methods now.

spec/spec_helper.rb
```diff
@@ -91,4 +91,12 @@
...
+    session[:user_id] = user.try(:id)
```
johndbritton: I'm not a big fan of `try`. Is there ever a case where `sign_in` should be called with a nil user? Should that be an exception?

app/controllers/application_controller.rb
```diff
@@ -50,8 +50,18 @@ def decorated_current_user
...
 
+  def decorated_true_user
+    @decorated_true_user ||= true_user.decorate
+  end
+
   def current_user
-    @current_user ||= User.find(session[:user_id]) if session[:user_id]
+    return @current_user if defined?(@current_user)
+
+    @current_user = if true_user.try(:staff?) && session[:impersonated_user_id]
```
johndbritton: Is there ever a case where `true_user` would be `nil`? Can we avoid the call to `try`?

johndbritton:
This is looking great. I'd like to take another pass once you've had a chance to address all the comments so far so just let me know when you're ready for that.

app/controllers/application_controller.rb
mastahyeti: > Is there ever a case where true_user would be nil? 

If the user is unauthenticated.

config/routes.rb
mastahyeti: It feels weird to authenticate the user twice. The controller will always need to lookup `current_user` and any auth logic (like impersonation) would need to be duplicated in the controller and the constraint. You also don't have as much context in the constraint as you do in the controller, leading to authz issues like https://github.com/education/classroom/issues/105. I also feel like doing authn/authz in the controller is idiomatic and putting it elsewhere makes it unclear how/if authentication is working.

johndbritton: > It feels weird to authenticate the user twice.

Makes sense, thanks for giving some me some context.

mkcode: Just went through this same process in education-web. The solution we did was to extend `ActionDispatch::Request` and delegate to in in `ApplicationController`. Very nice for removing duplication and authorizing once in both Route Constraints and Controller

app/request/user_session_methods.rb
```ruby
module UserSessionMethods
  def current_user
    @current_user ||= User.find_by_id session[:user_id]
  end

  def current_user=(user)
    @current_user = user
    session[:user_id] = user.id
    session[:valid_until] = 1.days.from_now
  end

  def signed_in?
    !!(current_user && session[:valid_until] &&
      session[:valid_until] > Time.now)
  end
  ...
end

ActiveSupport.on_load(:action_controller) do
  ActionDispatch::Request.send(:include, UserSessionMethods)
end
```

app/controllers/application_controller
```ruby
class ApplicationController < ActionController::Base
  ...
  delegate :current_user, :current_user=, :signed_in?, :admin?,
           to: :request
  helper_method :current_user, :signed_in?, :admin?
  helper_method :admin_discount_request_path
  ...
```

```ruby
  mount Sidekiq::Web  => '/sidekiq', constraints: { admin?: true }
```

app/controllers/stafftools_controller.rb
mkcode: This looks fine to me.

config/chewy.rb
```diff
@@ -0,0 +1,25 @@
+Chewy.logger = Rails.logger
+
```
mkcode: @tarebyte - Did you evaluate using other Chewy strategies?

I generally prefer using an async Job strategy to get the ind updating time out of the request. It's as simple as adding this. Also, the `chewy` job type should added to the sidekiq entry in `Procfile` with this.

```ruby
# Use ActiveJob config for async index updates.
Chewy.strategy(:active_job)

# Chewy wraps controller actions in the atomic strategy by default.
# We change this to use ActiveJob here. Makes index updates async, outside
# of the request.
Chewy.request_strategy = :active_job
```

mkcode:
With chewy searches, I generally recommend adding integration tests that ensure that the expected results for search queries are returned. I like we can save this for a separate PR though.

config/chewy.rb
tarebyte: @mkcode I did not, I'm still diving into Chewy to see what it can do.

This looks great, thank you for the pointer.

tarebyte:
@mkcode @johndbritton I'be been doing a little bit of digging over the weekend, and wanted an opinion of using `elasticsearch-rails` over `chewy`.

My main interest is that it seems that I can do [search over multiple types of resources at the same time](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/README.md#searching-multiple-models) which I think would be quite advantageous.

My goal is that the stafftools search looks like the search on GitHub or something close to it.

screen shot 2015-11-12 at 6 42 57 am

@mkcode it doesn't really look like `chewy` is designed for this kind of thing out of the box, am I correct in this line of thinking?

johndbritton:
> My main interest is that it seems that I can do search over multiple types of resources at the same time which I think would be quite advantageous.

I haven't done tons of elasticsearch stuff, but I'd bet there's a way to do it with Chewy. @mkcode is a better person to ask in this case.

app/controllers/stafftools/users_controller.rb
tarebyte: To me it looks better to return a `404` than a `unauthorized` it's just a personal preference though.

tarebyte: @mastahyeti do you have an opinion on this?

mkcode:
> wanted an opinion of using elasticsearch-rails over chewy

Having experience with both, I strongly recommend chewy over elasticsearch-rails. Elasticsearch-rails is really poorly designed and a very incomplete integration. I could talk for an hour about how bad elastisearch-rails is, and how chewy does a great job with the same problem.

> My main interest is that it seems that I can do search over multiple types of resources at the same time which I think would be quite advantageous.

You can search across multiple object types and even indexes with either integration. This point actually speaks to the biggest problem with elasticsearch-rails; it doesn't separate the concepts of AR models from elastic search indexes. Elastic search indexes allow for multiple index types. In chewy you would add another `define_type` entry to you index. 

> My goal is that the stafftools search looks like the search on GitHub or something close to it.

Nice. :-) So a few recommendations in working with elastic search:

 * Design search queries first. Index design follows from the queries.
 * Download marvel / sense and use it to inactively create indexes and test out search queries: https://www.elastic.co/downloads/marvel. Be sure to find the `sense` part - basically an web IDE for building ES queries.
 * Reference the ES documentation, not any rails gem. Here is the relevant searching part: https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl.html
 * Stay away from elastic search 2.0 for now. Still to early for it.

Since ES queries are all json at the end of the day, it's pretty trivial to make class that builds the exact query that you want, without using any rails gem query dsl. So jump into the real stuff. It's fun and really powerful.

To start, I would rename the `UsersIndex` to `StafftoolsIndex` and define all the types that you would search across there.

app/chewy/stafftools_index.rb
```diff
@@ -0,0 +1,77 @@
+class StafftoolsIndex < Chewy::Index
+  define_type User do
```
tarebyte: @johndbritton what other resource indexes do you think would be nice to have?

tarebyte:
@mkcode thank you so much for the pointers they were really helpful! Especially the elasticsearch tips.

If you could just :eyes: over the [search implementation](https://github.com/education/classroom/pull/359/files#diff-22dc66320c0cd133d26c6fa203093eecR22) it would be much appreciated.

I've also come up with a way to [render indexes by type](https://github.com/education/classroom/pull/359/files#diff-f66816e295fe37f52bcb449c1323babaR1) so that when we start to search through other types it fits nicely rails model render convention.

@johndbritton I still need to update the tests, but let me know what you think!

app/controllers/stafftools_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def set_resources
+    resource_query = params[:query].present? ? match_phrase_prefix(params[:query]) : {}
+    @resources     = StafftoolsIndex::User.query(resource_query).page(params[:page]).per(20)
```
mkcode: Shouldn't this just be `StafftoolsIndex` an not `StafftoolsIndex::User` so you can get organization results as well?

app/helpers/stafftools_resource_render_helper.rb
```diff
@@ -0,0 +1,6 @@
+module StafftoolsResourceRenderHelper
+  def render_stafftools_resource(resource)
+    type = resource._data['_type']
+    render partial: "stafftools/#{type.pluralize}/#{type}", locals: { type.to_sym => resource }
+  end
+end
```
mkcode: Beautiful. Take advantage of the hidden `_type` field and automatically render a partial base on it. Nice job. This is simple and easily scalable.

mkcode:
@tarebyte - left some inline comments. Nice job with all this.

Looks like your search results would include organizations though, which could have been caught with some integration tests. Here is how I setup the tests:

in config/initializers/chewy.rb: https://gist.github.com/mkcode/eb7dcdb161643b1dd770

Replace spec/support/chewy.rb with this: https://gist.github.com/9f3dc0a36b12eabe6643
Then add `type: 'chewy' to any test that you would like to use real ES, create objects which will build the index, and execute a real search.

tarebyte:
@mkcode I wasn't sure what I wanted to do with organizations yet so I left them out of the search for right now.

I think I'll add the ability to search for organizations in another PR. Do you think I should remove the `OrganizationIndex` from the `StafftoolsIndex` for now?

mkcode:
> Do you think I should remove the OrganizationIndex from the StafftoolsIndex for now?

You only have the one StafftoolsIndex. I don't see any issue with leaving it in.

tarebyte:
> You only have the one StafftoolsIndex. I don't see any issue with leaving the organization object type in it.

Thank your for the clarification, my apologies on the terminology.

app/chewy/stafftools_index.rb
```diff
@@ -0,0 +1,142 @@
+class StafftoolsIndex < Chewy::Index
+  define_type Assignment do
```
tarebyte: @johndbritton I've added all of the models as objects for the `StafftoolsIndex` let me know if there other attributes that you think would be useful here.

johndbritton: This looks pretty good to me.

johndbritton:
This looks great.

johndbritton:
@tarebyte Can you document the Bonsai addon dependency in `README.md`

johndbritton:
Branch deployed this in production and ran `rake chewy:update`

johndbritton:
@tarebyte Can you also update instances of `staff-tools` to `stafftools`, it's one word in our other projects and I like consistency.

johndbritton:
Rubocop has some warning output in CI:

> /home/travis/build/education/classroom/app/chewy/stafftools_index.rb: Metric/ClassLength has the wrong namespace - should be Metrics
/home/travis/build/education/classroom/app/chewy/stafftools_index.rb: Metric/ClassLength has the wrong namespace - should be Metrics

https://travis-ci.org/education/classroom/builds/92851259#L855

johndbritton:
The popup confirmation box is double prompting.

tarebyte:
That's probably turbolinks, what do you think about it being a modal?

app/views/shared/_header.html.erb
```diff
@@ -16,8 +16,17 @@
           </li>
 
           <li>
-            <%= link_to decorated_current_user.github_url, class: 'tooltipped tooltipped-s', 'aria-label' => 'View GitHub profile' do %>
-              <%= image_tag decorated_current_user.avatar_url(40), class: 'avatar', alt: "@#{decorated_current_user.login}", height: 20, width: 20 %>
+            <% if impersonating?(current_user, true_user) %>
+              <%= link_to stop_impersonating_stafftools_user_path(current_user), class: 'tooltipped tooltipped-s', method: :delete, 'aria-label' => "@#{decorated_true_user.login} impersonating @#{decorated_current_user.login}" do %>
```
johndbritton: The tooltip text should read:

> Stop impersonating @username

The other tooltips are actions, this one should be as well.

johndbritton:
> what do you think about it being a modal?

I'm fine with whatever, but I do think there should be a confirmation step.

johndbritton:
Index creation complete:

> Imported StafftoolsIndex::User for 1074.59s, documents total: 10980

tarebyte:
@mkcode do you have an example of `elasticsearch` in the `Procfile.dev`? I'm currently running into weird issues.

mkcode:
@tarebyte - I just simply have: `elasticsearch: elastic search` in Profile.dev

tarebyte:
Weird it works for me now, but it didn't before, thanks @mkcode!

tarebyte:
@johndbritton this should be good to go. Let me know what you think

tarebyte:
:ship: :dash:

Issue #366, opened 2015-11-20
Errors with repository creation


jakesorce:
Hello,

My students are getting various errors while trying to accept the invitations to assignments. The first couple of assignments seemed to work ok but today they are getting the following errors:

1) repository already exists but they aren't able to see it or push to it
2) permission denied
3) forbidden
4) repository is empty

Each time they try to click the invite they get different errors.

We are currently running classes right now and would love to know what is happening / when a fix will be implemented.

Thanks

tarebyte:
@jakesorce I'm going to take some time on Sunday and see if I can get it resolved. I'm seeing the errors on our end I'm just not sure what's causing them yet.

Thanks for your patience!

jakesorce:
Much appreciated @tarebyte

jakesorce:
@tarebyte - any luck with this one? Trying to decide if we should continue running class with github classroom.

tarebyte:
@jakesorce I'm making headway, did you by chance reject your auth token for your organization?

The error I'm receiving is:

```ruby
NoMethodError: undefined method `private_repos' for nil:NilClass
```

The error [points to this line](https://github.com/education/classroom/blob/master/app/models/git_hub_organization.rb#L98)

This happens when we retrieve an organization through the API and through a token that can't retrieve the private information about an org.

tarebyte:
We should handle this error better, so I'm going to open a PR for it

tarebyte:
@jakesorce I'm closing this issue since it seems to be resolved, please feel free to reopen if you experience this again.

PR #369, opened 2015-11-23
Raise an error if the token cannot retrieve the orgs plan

tarebyte:
Working through #366 if the user has rejected their token for the org, when we try and retrieve the plan it errors org.

/cc @johndbritton @jakesorce


johndbritton:
:+1:

PR #380, opened 2015-11-30
Make all other resources on Classroom for GitHub searchable

tarebyte:
This makes all of the resources on GitHub available to us in the stafftools search

/cc @johndbritton


app/controllers/stafftools/assignment_invitations_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Stafftools
+  class AssignmentInvitationsController < StafftoolsController
```
johndbritton: You can write this as `class Stafftools::AssignmentInvitationsController < StafftoolsController`.

tarebyte: I had that before, but rubocop didn't like it. I can add a rule to ignore that if `class Stafftools::AssignmentInvitationsController < StafftoolsController` is preferable

johndbritton: If rubocop is complaining, it's fine like this. It's really just a preference.

tarebyte:
@johndbritton this is what the organization view looks like, the rest of them will be similar
screen shot 2015-12-03 at 11 52 50 am

johndbritton:
> this is what the organization view looks like, the rest of them will be similar

Looks good. I'd recommend adding a title to that page to indicate that this is an organization being viewed.

app/views/stafftools/assignment_repos/_assignment_repo.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <span class="assignment-icon assignment-icon-individual left">
+      <%= octicon('repo') %>
+    </span>
+
+    <div class="student-information">
+      <%= link_to stafftools_assignment_repo_path(assignment_repo.id) do %>
+        <strong class="css-truncate-target login">Assignment Repo: <%= assignment_repo.id %></strong>
+      <% end %>
+    </div>
+
```
johndbritton: This view should link to the repository on GitHub and show the resource status. If the repo has been deleted it'd be nice to have the ability to recreate it.

johndbritton: When you're working on the other views, try and provide status info to show that things are in a good (expected state) or bad state. For example showing that a user's token is invalid on the user stafftools page.

tarebyte: @johndbritton I think you're looking for the this page: https://github.com/education/classroom/blob/371c676f25264760c0fcf31bf48008eb2d62086b/app/views/stafftools/assignment_repos/show.html.erb

The `show` view should give all the details about it right?

johndbritton:
This looks good. When it comes to making the views, try to show status info where there might be problems and provide functionality to fix those problems. At the start, there won't be a ton of functionality, but we'll add to this over time as we need the features.

tarebyte:
> Looks good. I'd recommend adding a title to that page to indicate that this is an organization being viewed.

do you mean `<title>` or something more obvious on the page itself?

johndbritton:
Something on the page, right now when you look at the page you don't know what you're looking at. It could be a user, organization, assignment, or any other resource.

johndbritton:
Just tried this out on staging, it looks pretty great.

I think the main search page could be improved a bit, some thoughts:

* maybe don't show any results until the query input is populated
* Update the index for items without titles to be more descriptive "Group Assignment Repo: 44" isn't as useful as `#{group}'s group assignment repo for #{assignment.title}`. This will also make searching better.
* add tooltip text to result icons in the list, hovering over a :link: icon should tell me what it actually is
* Add object type to results `User: @johndbritton`
* Show `created / joined on #{date}` for users in result list

jasonlong:
What do you guys think of something like this? I added some spacing between each section and reduced the number of divider lines throughout the listings. To make the CSS more straightforward, I needed to update the Ruby for creating the result groupings. This may not be the best implementation and it probably busted some tests, so feel free to refactor/replace/whatever.

screen shot 2015-12-13 at 11 00 26 am

tarebyte:
@jasonlong I'll pull it down and check it out, but from the screenshot it looks great

tarebyte:
@jasonlong I really like the changes, if you have any thoughts about the stafftools view for a search results 

screen shot 2015-12-13 at 11 15 21 am

It would be very welcome

jasonlong:
Thoughts on this direction? I removed the outer box and kept the individual `.boxed-group`s to help with the visual clutter. I also added the striped headings we used in the regular stafftools.

screen shot 2015-12-15 at 1 53 46 pm

screen shot 2015-12-15 at 1 48 33 pm

tarebyte:
@jasonlong I like it, it feels cleaner and still has the information we need it to have.

Thoughts @johndbritton?

jasonlong:
:thumbsup: Just FYI, I have these changes locally if we decide to go this direction. I'll can commit these and get the other views updated as well.

johndbritton:
The styles look great, thank @jasonlong.

Regarding the results list in https://github.com/education/classroom/pull/380#issuecomment-164271036, how does it behave with pagination? Do we only show the top results or is there a pager?

tarebyte:
@johndbritton, we are paginating based on the search, and just group results on the page

johndbritton:
> we are paginating based on the search, and just group results on the page

:cool: That makes sense to me. Not sure if it's a common UI pattern or not but we don't have a huge amount of usage of these pages so I think it's probably fine.

tarebyte:
@jasonlong I think we are in agreement on this then, when you get a chance if you can push your local changes that would be :sparkles:

There are a lot of views so if you don't feel like doing all of them I'm more than happy to help out.

jasonlong:
Alrighty, I think the remaining stafftools `show` views are updated now. Let me know if you see anything broken.

:pencil2: Also, a couple of copy things to keep in mind:

- We always use sentence case for headings, labels, etc. (e.g. "Group assignment repositories" and not "Group Assignment Repositories").

- We always use "repository" and not "repo" for user facing content.

I updated the instances of these that I came across, but a more thorough check might be necessary.

tarebyte:
@jasonlong I found some other instances the only ones we can't change without a lot of ruby :tophat: are the ones that are created by the types since the model name actually uses repo instead of repository.

Some examples are things like

screen shot 2015-12-16 at 4 00 53 pm

screen shot 2015-12-16 at 4 01 01 pm

But I've fixed the rest that I could find.

@johndbritton after I push these changes let me know if there is anything else you'd like to see different.

Otherwise I think we're good to ship these changes.

jasonlong:
That all sounds good and I meant to ask about those access types. I
wouldn't sweat it too much, especially since these are admin views. Those
tips were meant as much for you future knowledge since you'll be doing more
GitHub work in the future now. :)

tarebyte:
:sparkles:

johndbritton:
Looks like you added the yellow hatch image, but aren't using it.

tarebyte:
@johndbritton we are, I'm just using the wrong image url in the css and it's not showing up.

PR #382, opened 2015-12-01
adding code coverage analysis tool

jethroo:
Proposal to add **simplecov** as analysis tool for tests, hence on long run it helps to increase quality and will encourage contributors to increase it as well ;) 

```
Coverage report generated for RSpec to /Users/cwirth/Workspace/classroom/coverage. 933 / 1264 LOC (73.81%) covered.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -78,6 +78,7 @@ group :test do
   gem 'database_cleaner'
   gem 'factory_girl_rails'
   gem 'faker'
+  gem 'simplecov', require: false
   gem 'vcr'
   gem 'webmock'
 end
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 /log/*
 !/log/.keep
 /tmp
+/coverage
 
 # Ignore local environment variables
 .env
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -131,6 +131,7 @@ GEM
     database_cleaner (1.5.0)
     debug_inspector (0.0.2)
     diff-lcs (1.2.5)
+    docile (1.1.5)
     dotenv (2.0.2)
     dotenv-rails (2.0.2)
       dotenv (= 2.0.2)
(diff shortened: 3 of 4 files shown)


Gemfile
```diff
@@ -75,6 +75,7 @@ group :production do
 end
 
 group :test do
+  gem 'simplecov', '~> 0.11.0', require: false
```
johndbritton: Is there a reason you're using `~>` and specifying a version? We generally don't lock to a version unless it's necessary.

johndbritton:
Thanks for the contribution @jethroo!

@tarebyte What do you think about adding code coverage to the build?

Gemfile
jethroo: @johndbritton I'm somewhat used to fixate version numbers (at least in larged projects) to have a controlled updated process in which one can bump versions as needed step by step thus reducing the propability of introducing errors due to the version changes. But one may argue that with 99% coverage this is not really needed since one will spot introduced errors right away (with less coverage these might be go through unnoticed).

Anyhow I'll stick to your project policies so just let me know if I should remove the version and I'll will update the PR accordingly ;)

johndbritton: > to have a controlled updated process in which one can bump versions as needed step by step thus reducing the propability of introducing errors due to the version changes

`Gemfile.lock` keeps our gems at the same version from machine to machine, and `bundle update <gem>` can be used to update gems individually without updating others.

> let me know if I should remove the version and I'll will update the PR accordingly ;)

Please do, thanks!

tarebyte:
@johndbritton I think that's a pretty good idea, let's put it under the linters in terms of order.

Also can we create a `script/coverage` to fit the pattern of the app?

jethroo:
@johndbritton removed version and paid respect to alphabetically order in test group

tarebyte:
Thanks @jethroo!

PR #384, opened 2015-12-04
adding test for staff_constraint.rb

jethroo:
... to increase overall test coverage a bit ;)

### DIFF
--- a/spec/lib/staff_constraint_spec.rb
+++ b/spec/lib/staff_constraint_spec.rb
@@ -1,0 +1,33 @@
+require 'rails_helper'
+
+describe StaffConstraint do
+  subject { described_class.new }
+
+  describe '#matches?' do
+    context 'when without user session' do
+      let(:request) { double('request', session: {}) }
+
+      it { expect(subject.matches?(request)).to be_falsey }
+    end
+
+    context 'when with user session' do
+      let(:id) { double('id') }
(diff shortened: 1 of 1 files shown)


tarebyte:
Looks great thanks @jethroo!

spec/lib/staff_constraint_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+describe StaffConstraint do
+  subject { described_class.new }
+
+  describe '#matches?' do
+    context 'when without user session' do
+      let(:request) { double('request', session: {}) }
+
+      it { expect(subject.matches?(request)).to be_falsey }
```
johndbritton: The preference in this project is to use the more verbose `do...end` syntax for `it` blocks. The extra information provided makes the test output look really nice and easy to parse.

```ruby
it 'does not match the constraint' do
  expect(subject.matches?(request)).to be_falsey
end
```

Can you update this and the other `it` blocks to use this format?

johndbritton: Just noticed this was already merged. If you're feeling extra nice, bonus :sparkles: for a follow up PR.

jethroo: sure will do ;)

Issue #288, opened 2015-09-28
Group Assignment - Maximum number of members


ChristianMurphy:
This is a feature request, to add an optional field in the settings for the maximum number of members per team.

This is a setting would either: only be set on initial creation and be immutable afterwards, or could only be edited to a higher value.
Otherwise, if the value could be edited to a lower number of members per team, there would be the problems of removing members from a team, which members would be removed, and where those members would go.

johndbritton:
Thanks for the suggestion @ChristianMurphy. I think this is a totally reasonable request and something that wouldn't be difficult to implement.

We'd welcome a pull request adding such a feature.

nixpulvis:
Along the same lines, it would be nice to allow staff to assign groups.

johndbritton:
> Along the same lines, it would be nice to allow staff to assign groups.

@nixpulvis Could you open another issue for that feature request? I'd be especially interested in your ideas as to how the teacher could assign the groups. When creating the assignments, we don't necessarily have the GitHub account info for all of the students yet.

franktank:
Should the input be a field or a select list? Also, should the column for team_size or group_size be added to group_assignments?

johndbritton:
> Should the input be a field or a select list?

It should be an optional field. Empty would mean no limit, enter an integer and we'll enforce that limit.

> Also, should the column for team_size or group_size be added to group_assignments?

Yes, the value should also be editable on the settings page for the group assignment.

PR #388, opened 2015-12-06
Maximum number of team members for group assignments

beydogan:
This PR adds ability to set a maximum number of team members for group assignments. #288 

screenshot from 2015-12-06 14 28 50
screenshot from 2015-12-06 14 31 09


tarebyte:
@beydogan can you make the error message a little more descriptive?

Maybe something like "You cannot join this team because it already has 5 members".

beydogan:
@tarebyte done :+1:

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -52,6 +52,10 @@ def authorize_group_access
     group_id = group_params[:id]
 
     return unless group_id.present?
+    group = Group.find(group_id)
```
tarebyte: @beydogan I this logic is getting to be a bit much, I think we need to start extracting this out into different methods.

What do you think about something like this?

```ruby
def authorize_group_access
  group_id = group_params[:id]

  return unless group_id.present?
  return unless group_apart_of_grouping?(group_assignment, group_id)
  return unless group_has_reach_max_members?(group_assignment, group_id)
end
```

beydogan: @tarebyte I was thinking the same, on it now.

spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -58,6 +58,17 @@
...
+        allow_any_instance_of(RepoAccess).to receive(:silently_remove_organization_member).and_return(true)
+        group = Group.create(title: 'The Group', grouping: grouping)
+        group_assignment.update(max_members: 1)
+        user_1 = GitHubFactory.create_classroom_student
+        group.repo_accesses << RepoAccess.create(user: user_1, organization: organization)
+
+        expect_any_instance_of(ApplicationController).to receive(:flash_and_redirect_back_with_message)
+        patch :accept_invitation, id: invitation.key, group: { id: group.id }
+      end
     end
```
tarebyte: It would also be nice to have a test showing that user is added to the group if the max number has not been met

spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -58,6 +58,17 @@
...
+      it 'does not allow if group has maximum number of members' do
+        allow_any_instance_of(RepoAccess).to receive(:silently_remove_organization_member).and_return(true)
+        group = Group.create(title: 'The Group', grouping: grouping)
+        group_assignment.update(max_members: 1)
+        user_1 = GitHubFactory.create_classroom_student
+        group.repo_accesses << RepoAccess.create(user: user_1, organization: organization)
+
+        expect_any_instance_of(ApplicationController).to receive(:flash_and_redirect_back_with_message)
+        patch :accept_invitation, id: invitation.key, group: { id: group.id }
+      end
```
tarebyte: It would also be nice to have a test showing that user is added to the group if the max number has not been met

johndbritton:
Thanks for this contribution @beydogan. I think the idea is right, but the way it's exposed to the user could be improved.

What do you think about making the accept assignment page work a bit more like the add classroom organization page. We could remove the dropdown menu to choose the team and instead display a list of teams and associated avatars for team members. That way, it's easier for users to join the correct team. Additionally, we could show an empty space avatar on teams to indicate how many seats are left on that team.

On the assignment creation side, there might be a better way to expose this than a free form text entry box.

Also, we should probably set some limitations for sanity sake. Should a teacher be able to set the team limit to 1,000? How would that impact the join a group page?

tarebyte:
Closed in favor of #511

PR #393, opened 2015-12-08
Don't use a cached resource for repo creation

### LINKED ISSUE #360: Repositories occasionally created with wrong user's name
I have had two cases where a user has clicked to join an assignment and a repo has been created correctly, but with the wrong username.

The is the link to the assignment page in case it helps you investigate:
https://classroom.github.com/classrooms/9050108-mpags-python-2015/assignments/coursework
and this is the corresponding organisation:
https://github.com/mpags-python

On the first occasion, although a student was logged in and clicked the invitation link, the repo was created with my (the assignment creator's) username.  I had previously joined my own assignment (to test the process) and then deleted the repo.  I'd seen some other discussion in the issues about odd behaviour when deleting repos, and that work was ongoing, so I didn't report it at the time.

However, I've just seen a second case, in which a student joined the assignment, but the repo that was created was named after a completely different user.  The erroneous user doesn't appear to be associated with the course at all. The correct user was assigned access to the repo (even though the names didn't match).

Both times I have simply renamed the repo to fix the problem, and haven't noticed any further issues.

Is there any way that the repo creation code could accidentally get the wrong username?

tarebyte:
Closes #360 

I still need to work out how to test this, but I think this will fix the issue.

/cc @johndbritton


johndbritton:
Looks reasonable. Do you have any more info on how caching could be happening for this call?

It'd be great if you could narrow down the case where one request is made and cached and then used inappropriately in a subsequent request.

Since the creation is being done on behalf of the teacher, maybe that's the root cause. Would love to see a failing test for this case.

tarebyte:
@johndbritton I've updated the specs for `GitHubUser` to enforce that we use the id as part of the `GET` request.

This should be good to :ship:

johndbritton:
:shipit: Thanks for adding those tests.

PR #395, opened 2015-12-09
Design tweaks

jasonlong:
While getting back up to speed in preparation to review the upcoming staff tools changes, I noticed a few minor unrelated design things I could clean up.

First, I updated the CSS rule ordering based on our existing, preferred order. I've added an `.scss-lint.yml` which includes this ordering.

I added a subtle drop shadow to the text in the heading to help it stand out better against the patterned backgrounds. This matches what we do on https://guides.github.com and https://github.com/explore.

**Before**

screen shot 2015-12-09 at 12 33 35 pm

**After**

screen shot 2015-12-09 at 12 33 11 pm

I removed the indentation before this text field:

screen shot 2015-12-09 at 1 04 08 pm

Now:

screen shot 2015-12-09 at 1 06 48 pm

The longer group assignment description here makes this UI unbalanced. 

screen shot 2015-12-09 at 1 29 57 pm

I'm proposing this change which I hope is still clear enough.

screen shot 2015-12-09 at 1 30 53 pm

And finally, the current footer seems really ginormous.

screen shot 2015-12-09 at 1 11 27 pm

I made this smaller to match our other sites (like [Octicons](https://octicons.github.com/)) and updated it to use Octicons + text rather than an image.

screen shot 2015-12-09 at 1 26 56 pm

/cc @tarebyte @johndbritton

### DIFF
--- a/app/assets/stylesheets/.scss-lint.yml
+++ b/app/assets/stylesheets/.scss-lint.yml
@@ -1,0 +1,406 @@
+# This is the lint file for .scss files. It uses https://github.com/brigade/scss-lint
+# to search through .scss files and point out errors.
+# You can view these errors in your editor.
+#
+# Here's a link to all the default configurations
+# https://github.com/brigade/scss-lint/blob/master/config/default.yml
+# below is our settings.
+
+linters:
+  BangFormat:
+    enabled: true
+    space_before_bang: true
+    space_after_bang: false
+    severity: error
+
+  BemDepth:
+    enabled: false
+
+  BorderZero:
+    enabled: true
+    convention: zero # or `none`
+    severity: error
+
+  ColorKeyword:
+    enabled: true
+    severity: error
+
+  ColorVariable:
+    enabled: false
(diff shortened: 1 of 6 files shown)


tarebyte:
This looks great! @jasonlong is there a reason why the GitHub Octicon is higher than the other ones?

screen shot 2015-12-09 at 3 19 20 pm

tarebyte:
I was about to open up an issue with octicons, but realized I had already commented on one similar https://github.com/github/octicons/issues/59

jasonlong:
Yeah, that one is funky. There was already a 1px adjustment to it (copied from the Octicons footer), but 2px seems to look better (which I just committed).

tarebyte:
I just pulled the change down, yeah the extra pixel really does make a difference.

I'll let this sit so that @johndbritton can take a look before we :ship: :dash:

johndbritton:
:heart::heart::heart:

:shipit:

PR #397, opened 2015-12-10
Closes #396 autofocus user input textbox

### LINKED ISSUE #396: Focus text box on display
When clicking the "reset and remove" classroom button, the confirmation box requires typing the organization name. That text box should be focused immediately.

That box should get focus in every case where we use this pattern.

tarebyte:
/cc @johndbritton


johndbritton:
:shipit:

PR #409, opened 2015-12-19
Add Notes to stafftools

tarebyte:
This adds notes to the stafftools page by adding markdown files into the app and then parsing them.

This is what it looks like right now.

screen shot 2015-12-19 at 12 46 55 am

screen shot 2015-12-19 at 12 47 47 am

The listing needs some styling, but I wanted to get some initial feedback.

/cc @johndbritton @jasonlong

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem 'draper'
 gem 'faraday-http-cache'
 
 gem 'geo_pattern'
+gem 'github-markup'
 
 gem 'jbuilder'
 gem 'jquery-turbolinks'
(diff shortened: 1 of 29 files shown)


johndbritton:
Looks like there might have been a misunderstanding about the note feature. We don't want to store notes in the repository, but want to be able to display and write notes on resources in stafftools that would be persisted in the database.

You could create a StaffNote model with a polymorphic association to be used with the various resources.

Thinking about this even more, we might just want to punt on this functionality until we really need it.

tarebyte:
Yup totally misunderstood this, thanks for the clarification!

PR #410, opened 2015-12-19
Support special characters for Assignments

### LINKED ISSUE #266: The members in my organization cant enter the github classroom page
i just invite them into my organization
after they accepted, they cant enter the classroom page.

tarebyte:
Closes #266


tarebyte:
@johndbritton the more I think about this, the more I believe we should not support special characters.

We should instead handle them how GitHub does, for example an Ü becomes a -. This means that we should also add some client side warnings so that the user knows what they're assignment name will be.

What do you think?

johndbritton:
We shouldn't change user input, so if the title includes special characters, we should display them (as GitHub does). I'm fine with replacing special characters in URL slugs, though.

tarebyte:
I agree about still displaying special characters, the thing is though GitHub repos don't support special characters. It just replaces everything with one dash.

This is an example of that scenario right now.

screen shot 2015-12-21 at 6 51 59 am

Maybe we should warn the user when they make the title, something similar to this.

screen shot 2015-12-21 at 6 54 04 am

johndbritton:
> the thing is though GitHub repos don't support special characters.

Names of assignments aren't GitHub Repo names. This is a perfect example, the assignment name should be able to use special characters even if the repository cannot.

If anything, I think this is another reason why we shouldn't put the assignment name in the URL and potentially another reason to get rid of `title` all together.

johndbritton:
> Names of assignments aren't GitHub Repo names.

Except for the fact that we create the repo names based on the name of the assignment. Is there another way we could make these names?

johndbritton:
Perhaps we create a `slug` field that is used to generate the repo name on GitHub and allow teachers to set that. If we wanted to get fancy, you could display two fields `title` and `slug` and have the slug field auto-populated but editable by the user.

tarebyte:
> Perhaps we create a slug field that is used to generate the repo name on GitHub and allow teachers to set that.

We already have the `slug` column on those models so we could add those to the form.



> If we wanted to get fancy, you could display two fields title and slug and have the slug field auto-populated but editable by the user.

That could be very easy to do, I'm :+1: for it.

PR #418, opened 2016-01-04
Give users the ability to recreate submission repositories

tarebyte:
I wanted to start a discussion about how we want to give users a better set of tools to manage their ~~assignments~~ submissions on a per repo case.

There are a lot of times where the user deletes a repo on GitHub and then wants to be able to create a new ~~assignment~~ submission for a student for whatever reason.

My thought here is that on the assignment page

screen shot 2016-01-04 at 9 41 46 am

We can give the organization owners the ability to delete one or more ~~assignments~~ submissions.

I see the selection following this pattern:

screen shot 2016-01-04 at 9 45 09 am

Thoughts?

/cc @johndbritton


tarebyte:
also /cc @jasonlong for design insight

johndbritton:
> We can give the organization owners the ability to delete one or more assignments.

Do you mean delete one or more submission? What happens to the team, does that get deleted too?

My thought was to show a "recreate repo" button whenever we display a NullGitHubRepo

tarebyte:
> Do you mean delete one or more submission? 

Yes that's what I mean, sorry for the poor phrasing.

> What happens to the team, does that get deleted too?

Possibly, I wasn't sure which way to go with that. Do you have a preference?

> My thought was to show a "recreate repo" button whenever we display a `NullGitHubRepo`

I like that concept, that should be pretty easy to do.

jasonlong:
On the listing view there, it looks like the number of avatars and the NWO link are both truncated fairly aggressively leaving a void in the middle that would never be utilized.

Without mocking it up, I'm wondering if it might feel better to move the NWO link to the left below the name and making the truncation length fairly long. Then the avatars could be put on the right, perhaps making them a bit smaller. This would let you show a few more and maybe even utilize 2 rows before truncating.

These are a quick 2 cents. I can take a closer :eyes: if you like.

tarebyte:
@jasonlong something like this?

screen shot 2016-01-04 at 8 57 23 pm

jasonlong:
Yeah, I feel like that flows a little better. What do you think?

tarebyte:
I like that change as well,

What do you think about moving from this?

screen shot 2016-01-05 at 7 44 11 am

To this?

screen shot 2016-01-05 at 7 45 26 am

jasonlong:
That feels better to me as well. If you wanted to keep both names, you could show the regular name next to the username. e.g.

**tarebytetest** (Tarebyte Test Account)

With that treatment, we usually go with a lighter font as well on the non-bold text.

tarebyte:
@johndbritton it needs some UI polish, but this is the general feel for how this feature works for `AssignmentRepos`.

deleted assignment_repo

tarebyte:
@johndbritton what do you think about this for the `group_assignment` submissions, I'm not super happy with it.

screen shot 2016-01-06 at 12 22 42 am

johndbritton:
> what do you think about this for the group_assignment submissions

Yeah, not a fan of that either. What about stacking the avatars up on the left side?

johndbritton:
Is there a way for the student to recreate the repository from the invitation after it's been accepted and then the repo deleted?

tarebyte:
not as of right now, the `AssignmentRepo` still exists which means that the student can't reaccept the invitation.

However we can add some logic to verify the GitHub repo still exists, and if it doesn't we can allow them to accept the invitation again and recreate the repo.

johndbritton:
> However we can add some logic to verify the GitHub repo still exists, and if it doesn't we can allow them to accept the invitation again and recreate the repo.

I think we should do that, but also make it clear to the student what's happening before creating the new repository.

On the invitation page we could say something like

> You previously accepted this assignment, but we can't find the repository we created for you. It's possible it was deleted, would you like us to create a new repository for you to start fresh?

drbeco:
I don't like too much the **recreate** button. It raises question of what is "reused" and what is not. Leaving students with questions that may delay the action until next class, where the teacher would need to explain.

I prefer a simple **delete**, and then go back to the first page and create a new repository like if it never happened.

My 2 cents.

tarebyte:
We added Webhooks so now if a repository needs to be recreated we can delete it on GitHub and then re accept the assignment.

Closing.

PR #370, opened 2015-11-23
Install bullet

mkcode:
Install and configures bullet: https://github.com/flyerhzm/bullet

This will find N+1 queries, Unused Eager loading, and needed counter caches.

This configures bullet in a strict way, where bullet will raise errors in test mode. This will prevent any N+1 (and more) queries from being merged into master.

/cc @johndbritton

PR #419, opened 2016-01-05
Make N+1's more noticeable

tarebyte:
I was running the app and realized that there are n+1 queries happening but I just missed them

What do we think about adding these notices to airbrake so that they are brought to our attention in a more obvious way?

/cc @johndbritton @mkcode

### DIFF
--- a/app/views/group_assignment_repos/_group_assignment_repo.html.erb
+++ b/app/views/group_assignment_repos/_group_assignment_repo.html.erb
@@ -7,7 +7,7 @@
     <% end %>
 
     <ul>
-      <% group_assignment_repo.group.repo_accesses.limit(4).each do |repo_access| %>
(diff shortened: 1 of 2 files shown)


mkcode:
Great. Getting the bullet errors in airbrake is a good idea. 

Bullet.raise - https://github.com/education/classroom/pull/370/files#diff-412a4fba00cba5ff704676f839824c0fR15 - in #370 should be raising errors for detected n+1 queries in test mode - and therefore failing CI builds and now allowing deploys. So an N+1 in production should indicate that there is no test coverage of that code. @tarebyte - can you verify this?

tarebyte:
You got it!

johndbritton:
> What do we think about adding these notices to airbrake so that they are brought to our attention in a more obvious way?

:+1:

> So an N+1 in production should indicate that there is no test coverage of that code.

Yep, means we need some more test coverage.

tarebyte:
Will revisit this another time, closing due to staleness.

PR #422, opened 2016-01-06
Verify that the title is present

tarebyte:
Not sure if this is the best way, but this fixes the issue we are having for recreating the stafftools index.

/cc @johndbritton


tarebyte:
Taking a new direction with this

PR #425, opened 2016-01-09
Set Chewy to log level info

johndbritton:
Our logs are full of Chewy debug info, this will reduce the noise.


tarebyte:
:sparkles:

PR #454, opened 2016-02-02
Cache bundler to speed up travis builds

kylekthompson:
Caching bundler will speed up your travis builds (mine dropped from 2:30 to about 1:00).

More info can be found [here](https://docs.travis-ci.com/user/caching).

Small note: in their docs it says that it is only available for private repos; however, in my public repos it works well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ script:
 
 bundler_args: --without development production --deployment --jobs=3 --retry=3
 
+cache: bundler
+
 notifications:
   email: false
 


kylekthompson:
> Small note: in their docs it says that it is only available for private repos; however, in my public repos it works well.

Looks like it worked in the build for this branch as well. (the build speed would be noticed in subsequent builds now that it is cached.

johndbritton:
The original build time:

> Elapsed time 2 min 52 sec

After re-triggering the build:

> Elapsed time 1 min 25 sec

Looks good to me, leaving this to @tarebyte to merge.

tarebyte:
Thanks for the PR @kylekthompson!

PR #473, opened 2016-02-28
Make the bootstrapping for Classroom easier

tarebyte:
This adds some nice things for running Classroom.

* For OSX users with Homebrew
  * No longer have to run `brew install`
  * Can now access `http://classroom.dev` instead of localhost
* For all users
  * No longer have to open separate windows for the server and for workers.

/cc @johndbritton @mkcode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ dump.rdb
 
 # RubyMine
 /.idea
+
+# Nginx
+config/dev/nginx.conf
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,11 @@
+# Helpers
+tap "homebrew/versions"
+tap "github/bootstrap"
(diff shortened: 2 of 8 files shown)


johndbritton:
:+1: Looks great. Assuming you've done a fresh setup to validate that everything works as expected.

mkcode:
Looking over this: it ran well for me - save for the postgres upgrade issues I experienced that are not the fault of this PR though. That github/bootstrap tap repo is cool as well. I definitely would like to take some these ideas to education-web.

I'm a bit confused as to how the nginx portion of this works though. The config it generates is not actually used when running `script/server`. Looks like nginx is autoloaded through homebrew - killing the process on my machine yielded no changes - I could still access classroom.dev:5000 the same way.

PR #479, opened 2016-03-03
environment variable names corrected

hackerkid:
The variable names in the readme was not the same as in the .env file. A grammar error is also fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,12 @@ If you are recording new cassettes, you need to make sure all of these values ar
 
 ENV Variable | Description |
 :-------------------|:-----------------|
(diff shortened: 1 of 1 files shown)


johndbritton:
Looks like you also need to regenerate the VCR cassette since the test description is used there.

hackerkid:
@johndbritton I have updated the commit to include changes only in Environment variables. I will send another PR for the second problem so that you can merge this.

johndbritton:
:+1: from me, @tarebyte?

tarebyte:
Thanks @hackerkid!

PR #484, opened 2016-03-08
Documentation suggestions

ksuderman:
A minor reworking of the _Invite other administrators_ page. Also fixes a broken link in the CONTIBUTING.md page.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ Here are a few things, that will help us help you when opening an issue.
 ## Submitting a pull request
 
 0. [Fork][fork] and clone the repository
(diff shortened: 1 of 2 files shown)


tarebyte:
Thanks for the Pull Request @ksuderman :sparkles: 

For convenience this is what your change looks like on Classroom

screen shot 2016-03-08 at 8 38 03 am

I think I'd like to see more whitespace between the url button and the explanation, what do you think?

ksuderman:
Yes, more whitespace would be good.  I wasn't able to run/view the classroom project before submitting the PR or I likely would have added more whitespace myself.  However, the idea is to make is clearer what URL should be sent; apparently my attention span is only two lines long and I was always reading "this URL" as https://classroom.github.com and not the URL to our actual classroom.

PR #485, opened 2016-03-08
Request more, but not less scopes than we have for the user

### LINKED ISSUE #460: Student not given write permission to repo after accepting assignment
First of all, sorry if this has already been made an issue - I didn't see anything open, and I still see this problem happen in production.

At [Make School](http://makeschool.com) we use Classroom for Github a lot.  See our classroom organization here: https://github.com/makeschool-17

Very commonly, students won't be granted write permission to their repositories after accepting assignments.  It seems to happen somewhere between 10 to 20% of the time.  [Here's](https://github.com/MakeSchool-17/ruby-koans-cmaher92) an example repository where the student wasn't granted write permission, so I manually added him this morning.

johndbritton:
This adds a new method `ApplicationController#request_scopes` which makes it so that we request more scopes when necessary, but if we need fewer scopes for a given action we don't ask GitHub to reduce the scopes for the user token.

@mastahyeti for a quick once over.

Fixes #460

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -34,13 +34,17 @@ class ApplicationController < ActionController::Base
     required_scopes.all? { |scope| current_scopes.include?(scope) }
   end
 
+  def request_scopes
+    required_scopes | current_scopes
+  end
+
   def authenticate_user!
     auth_redirect unless logged_in? && adequate_scopes?
   end
 
   def auth_redirect
     session[:pre_login_destination] = "#{request.base_url}#{request.path}"
-    session[:required_scopes] = required_scopes.join(',')
+    session[:required_scopes] = request_scopes.join(',')
     redirect_to login_path
   end
 


tarebyte:
@johndbritton this doesn't work for me, I just tried it out and still bumped down my token.

app/controllers/application_controller.rb
```diff
@@ -34,13 +34,17 @@ def adequate_scopes?
     required_scopes.all? { |scope| current_scopes.include?(scope) }
   end
 
+  def request_scopes
+    required_scopes | current_scopes
+  end
```
tarebyte: So when you sign into an invitation this is what this method looks like

```ruby
required_scopes
#=> ["user:email"]

# Because we aren't logged in yet.
current_scopes
#=> []

required_scopes | current_scopes
#=> ["user:email"]
```

johndbritton: Aha, so this is because the user isn't logged in. Any ideas on how to ask for the same scopes we have plus some more?

tarebyte: Not sure yet, because we don't know what permissions the user already has because we don't know who the user is signing in before hand.

tarebyte:
This is not needed, issue fixed by #604

PR #486, opened 2016-03-08
Use start_source_import from octokit

tarebyte:
After the Source Import API preview shipped https://developer.github.com/changes/2016-02-19-source-import-preview-api/ I added it to Octokit.rb https://github.com/octokit/octokit.rb/pull/722.

This makes use of the official Octokit version.

/cc @johndbritton


johndbritton:
:shipit:

PR #490, opened 2016-03-09
Add credentials to source_import request

### LINKED ISSUE #489: Importer Failing on New Private Assignments
@johndbritton This issue appears to be similar to https://github.com/education/classroom/pull/323. We are using Classroom to distribute activities to Trainer candidates. It worked well until this week when we started seeing a recurring issue with the importer.

1. The candidate accepts the assignment. The assignment template is a private repo and the assignment creates a new private repo for the candidate.
2. At the importer stage the candidate is being asked to authenticate. They cannot authenticate with their own credentials and the import fails.
3. I am able to log in as GitHubTeacher (the assignment owner) and enter my credentials to complete the import.

Ideas?

tarebyte:
Closes #489 

This needs to have tests, but I'd like to follow up with that when I have more time.

/cc @johndbritton @crichID

### DIFF
--- a/app/models/git_hub_repository.rb
+++ b/app/models/git_hub_repository.rb
@@ -26,7 +26,8 @@ class GitHubRepository
   #
   def get_starter_code_from(source)
     with_error_handling do
-      @client.start_source_import(@id, 'git', "https://github.com/#{source.full_name}")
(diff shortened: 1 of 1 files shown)


tarebyte:
I just confirmed that this works on production, sorry for the trouble @crichID

PR #492, opened 2016-03-10
Make sure that we are getting the AssignmentRepos user for indexing

tarebyte:
There are quite a few models that didn't transition from using `RepoAccess` for `AssignmentRepo` therefore when we reran the index it would fail because we weren't actually grabbing the user and therefore the `github_client` was not present.

I deployed this to branch to production, and the reindexing of Classroom was successful.

/cc @johndbritton

### DIFF
--- a/app/chewy/stafftools_index.rb
+++ b/app/chewy/stafftools_index.rb
@@ -40,7 +40,7 @@ class StafftoolsIndex < Chewy::Index
     field :assignment_title, value: ->(assignment_invitation) { assignment_invitation.assignment.title }
 
     field :user_login, value: (lambda do |assignment_repo|
-      user = assignment_repo.user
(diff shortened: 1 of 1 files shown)


johndbritton:
> There are quite a few models that didn't transition from using `RepoAccess` for` AssignmentRepo`.

@tarebyte Can you elaborate on this a bit? Do we have some cleanup left to do from the transition?

tarebyte:
@johndbritton sure, a while ago when we did the migration in #331 we couldn't move people from having a personal team to being an outside collaborator on a repository because the tokens in use were invalid :cry: 

For the sake of not deleting data that may be used again we decided to leave that small batch and force new users to be an outside collaborator for individual assignments.

So I don't think we can really do a "cleanup" unless you want to remove users who don't have valid access tokens anymore.

PR #501, opened 2016-03-16
Automatically add users to any existing classrooms

### LINKED ISSUE #259: Auto join any existing classrooms
If a user is an admin on an organization that exists on Classroom, they should auto join the organization and it should be listed on `/organizations`.

There have been a couple cases where other administrators have added the organization to Classroom without sending out the direct invite link and they didn't get the access they should have.

cc/ @mks-instructors

dinever:
Hi guys,

Fixes #259. It automatically looks up the organizations and add the current user to the corresponding classroom if he/she has the admin privilege. 

I found it a little bit hard to add test cases for this functionality. We only have 2 users(a student and a teacher) set in the test environment. However, to test this feature we need a 3rd user: The user who has admin privilege to the organization but was not invited to the classroom.

I'm a GSoC 2016 candidate and this is my first pull request for the Classroom. Any advice/feedback would be greatly appreciated! :smile: 

Thanks!


dinever:
Sorry for the failing check. Seems that the method has too many lines, I will extract those lines into another method.

app/controllers/organizations_controller.rb
```diff
@@ -125,6 +126,17 @@ def set_users_github_organizations
   end
   # rubocop:enable AbcSize
 
+  # Check if the current user has any organizations with admin privilege, if so add the user to the corresponding
+  # classroom automatically.
+  def auto_scan_organizations_access
+    @users_github_organizations.each do |organization|
+      next unless organization[:classroom].present?
+      next if organization[:classroom].users.include?(current_user) || current_user.staff?
```
tarebyte: Why would we want to skip over the user if they are a staff member? Staff members can be part of the organization.

app/controllers/organizations_controller.rb
```diff
@@ -125,6 +126,17 @@ def set_users_github_organizations
...
+      github_org = GitHubOrganization.new(current_user.github_client, organization[:github_id])
+      github_org.admin?(decorated_current_user.login) ? organization[:classroom].users << current_user : not_found
```
tarebyte: What would be the case we raise a `404` here?

tarebyte:
Thanks for the contribution @dinever!

I would :sparkling_heart: some tests around this new functionality. I also left a couple of comments.

dinever:
Hi Mark @tarebyte,

I have added two test cases for this one and have solved the issues mentioned in the comments you left. You are right, we should not pass the staffs and should not raise a 404 page when the user is not an admin to the orginization.

Thanks!

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,23 @@
       end
     end
 
+    context 'auto add user to classroom' do
+      before do
+        sign_out
+        sign_in(student)
+      end
+
+      it 'user with admin privilege' do
```
johndbritton: I think this test could be improved:

```ruby
context 'user with admin privilege on organization' do
  it 'adds the user to the organization' do
    get :index
    expect(assigns(:current_user).organizations.first.id).to eq(organization.id)
  end
end
```

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,23 @@
       end
     end
 
+    context 'auto add user to classroom' do
```
johndbritton: 'add user to classroom' isn't a context

spec/controllers/organizations_controller_spec.rb
```diff
@@ -5,6 +5,7 @@
 
   let(:organization)  { GitHubFactory.create_owner_classroom_org }
   let(:user)          { organization.users.first                 }
+  let(:student)       { GitHubFactory.create_classroom_student   }
```
tarebyte: @dinever is your student an admin of the organization?

dinever: @tarebyte yes, the student here is the admin of the organization in the VCR record of the test case.

tarebyte: Ah that explains it, so the student is not supposed to be a member of the organization. It is supposed to be just another GitHub user

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,22 @@
...
+        get :index
+        expect(assigns(:current_user).organizations.first.id).to eq(organization.id)
+      end
+
+      it 'will not be added if he is not admin of the GitHub organization' do
+        get :index
+        expect(assigns(:current_user).organizations.length).to eq(0)
+      end
+    end
+
```
tarebyte: @dinever this is a good start, what do you think about something like this?

```ruby
    context 'user with admin privilege on the organization but not apart of the classroom' do
      before(:each) do
        organization.users = []
      end

      it 'will add the user to the classroom' do
        expect(user.organizations).to be_empty

        get :index

        expect(user.organizations).not_to be_empty
        expect(user.organizations).to include(organization)
      end
    end

    context 'user without admin privilege on the organization' do
      before(:each) do
        sign_out
        sign_in(student)
      end

      it 'will not add the user to the classroom' do
        expect(student.organizations).to be_empty

        get :index

        expect(student.organizations).to be_empty
      end
    end
```

johndbritton: "apart" should be "part"

```ruby
      it 'will add the user to the classroom' do
        expect(user.organizations).to be_empty

        get :index

        expect(user.organizations).not_to be_empty
        expect(user.organizations).to include(organization)
      end
```

one expectation is enough here: `expect(user.organizations).to include(organization)`

```ruby
      before(:each) do
        sign_out
        sign_in(student)
      end
```

I really don't like that we're signing out and then back in here. Can we modify the nesting so the sign out isn't necessary?

"will not add the user to the classroom" should be "does not" and only one expectation is needed here.

dinever:
Hi guys, @tarebyte @johndbritton

I just had the test cases improved according to the comments, which really helped me a lot.

Thanks!

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,34 @@
...
+        get :index
+
+        expect(user.organizations).to include(organization)
+      end
+    end
+
+    context 'user without admin privilege on the organization' do
+      before(:each) do
+        sign_in(student)
+      end
```
dinever: We don't need to `sign_out` before `sign_in` another user, as `sign_in` simply updates `session[:user_id]`.

```ruby
def sign_in(user)
  session[:user_id] = user.id
end
```

So I removed the `sign_out`, just put a `sign_in` here.

tarebyte:
There doesn't seem to be a performance hit like I thought there would be.

@johndbritton with this request though we're only getting the first 100 organization for a user

```
HTTP Cache: [GET /user/memberships/orgs?per_page=100&state=active] unacceptable
```

Is this something we need to be concerned about?

johndbritton:
> with this request though we're only getting the first 100 organization for a user

@tarebyte I believe we're using Octokit's [auto-pagination](https://github.com/octokit/octokit.rb#auto-pagination) feature, meaning that it will grab the other pages automatically if they exist.

app/controllers/organizations_controller.rb
```diff
@@ -125,6 +126,23 @@ def set_users_github_organizations
...
+  # Check if the current user has any organizations with admin privilege, if so add the user to the corresponding
+  # classroom automatically.
+  def auto_scan_organizations_access
+    @users_github_organizations.each do |organization|
+      organization = organization[:classroom]
+      next unless organization.present?
+      next if organization.users.include?(current_user)
+      create_user_organization_access(organization)
+    end
+  end
```
johndbritton: I think this is easier to understand:

```ruby
  def auto_scan_organizations_access
    @users_github_organizations.each do |organization|
      organization = organization[:classroom]
      if organization.present? && organization.users.include?(current_user)
        create_user_organization_access(organization)
      end
    end
  end
```

app/controllers/organizations_controller.rb
```diff
@@ -2,7 +2,8 @@ class OrganizationsController < ApplicationController
   include OrganizationAuthorization
 
   before_action :authorize_organization_addition,     only: [:create]
-  before_action :set_users_github_organizations,      only: [:new, :create]
+  before_action :set_users_github_organizations,      only: [:index, :new, :create]
+  before_action :auto_scan_organizations_access,      only: [:index]
```
johndbritton: Is there a better name for this method? Suggestions @tarebyte?

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,34 @@
       end
     end
 
+    context 'user with admin privilege on the organization but not part of the classroom' do
+      before(:each) do
+        organization.users = []
+      end
+
+      it 'adds the user to the classroom' do
+        expect(user.organizations).to be_empty
```
johndbritton: This `expect` is not necessary

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,34 @@
...
+      end
+    end
+
+    context 'user without admin privilege on the organization' do
+      before(:each) do
+        sign_in(student)
+      end
+
+      it 'does not add the user to the classroom' do
+        expect(student.organizations).to be_empty
```
johndbritton: this `expect` is not necessary

spec/controllers/organizations_controller_spec.rb
```diff
@@ -34,6 +35,34 @@
       end
     end
 
+    context 'user with admin privilege on the organization but not part of the classroom' do
+      before(:each) do
+        organization.users = []
```
johndbritton: since the factory creates an org with a user, I believe this needs a call to `organization.save!`

tarebyte: @johndbritton I thought so too at first, but then I confirmed that this is actually correct in console

Here is my pry terminal

```ruby
>> o = Organization.first
  Organization Load (0.6ms)  SELECT  "organizations".* FROM "organizations" WHERE "organizations"."deleted_at" IS NULL  ORDER BY "organizations"."id" ASC LIMIT 1
=> #<Organization:0x007fc3acf97118
 id: 4,
 github_id: 12439714,
 title: "tatooine-moisture-farmers",
 created_at: Mon, 07 Mar 2016 06:13:25 UTC +00:00,
 updated_at: Mon, 07 Mar 2016 06:13:25 UTC +00:00,
 deleted_at: nil,
 slug: "12439714-tatooine-moisture-farmers">

>> o.users
  User Load (0.6ms)  SELECT "users".* FROM "users" INNER JOIN "organizations_users" ON "users"."id" = "organizations_users"."user_id" WHERE "organizations_users"."organization_id" = $1  [["organization_id", 4]]
=> [#<User:0x007fc3b48ec7c0
  id: 1,
  uid: 564113,
  token: "token-here-nice-try",
  created_at: Sun, 28 Feb 2016 17:17:56 UTC +00:00,
  updated_at: Thu, 17 Mar 2016 03:29:02 UTC +00:00,
  site_admin: true,
  role: 0>]

>> o.users = []
   (0.2ms)  BEGIN
  SQL (0.2ms)  DELETE FROM "organizations_users" WHERE "organizations_users"."organization_id" = $1 AND "organizations_users"."user_id" = 1  [["organization_id", 4]]
   (2.7ms)  COMMIT
=> []

>> Organization.first.users
  Organization Load (0.4ms)  SELECT  "organizations".* FROM "organizations" WHERE "organizations"."deleted_at" IS NULL  ORDER BY "organizations"."id" ASC LIMIT 1
  User Load (0.3ms)  SELECT "users".* FROM "users" INNER JOIN "organizations_users" ON "users"."id" = "organizations_users"."user_id" WHERE "organizations_users"."organization_id" = $1  [["organization_id", 4]]
=> []
```

So this works as expected

tarebyte:
@johndbritton that call actually uses the `application_client` which doesn't have auto pagination turned on. https://github.com/education/classroom/blob/master/app/models/concerns/git_hub.rb#L6-L9

We just need to update it to look like this.

```ruby
def application_github_client
  client_id        = Rails.application.secrets.github_client_id
  client_secret = Rails.application.secrets.github_client_secret
  Octokit::Client.new(client_id: client_id, client_secret: client_secret, auto_pagination: true)
end
```

app/controllers/organizations_controller.rb
tarebyte: @johndbritton I think your line needs to be

```diff
- if organization.present? && organization.users.include?(current_user)
+ if organization.present? && !organization.users.include?(current_user)
```

What do you think about this?

```ruby
def auto_scan_organizations_access
  @users_github_organizations.each do |organization|
    classroom = organization[:classroom]
    if classroom.present? && !classroom.users.include?(current_user)
      create_user_organization_access(classroom)
    end
  end
end
```

app/controllers/organizations_controller.rb
tarebyte: How about `join_organizations`?

spec/controllers/organizations_controller_spec.rb
tarebyte: agreed

spec/controllers/organizations_controller_spec.rb
tarebyte: agreed

tarebyte:
@dinever, @johndbritton and I left some comments for you to chew on.

Let us know if you have any questions!

dinever:
Hi Mark @tarebyte and John @johndbritton,

Thank you for that feedback! Just had the code modified according to them.

spec/controllers/organizations_controller_spec.rb
johndbritton: @dinever You should remove this line.

johndbritton:
@dinever Thanks for your contribution. I just left one comment about an `expect` that you don't need, but aside from that this looks good to me.

@tarebyte Can you give this a go locally and then merge when you're satisfied?

dinever:
Thanks @johndbritton, just got the unnecessary expectation line removed.

spec/controllers/organizations_controller_spec.rb
johndbritton: > the student here is the admin of the organization in the VCR record of the test case.

Maybe we should change the name from `:student` to `:teaching_assistant` to make this more clear.

tarebyte: @johndbritton but that's not supposed to be the role of this user. The student is someone who isn't apart of the organization.

dinever: @johndbritton, that comment from me is outdated. This `student` user is only used in the context `user without admin privilege on the organization` now. So I think `student` should be enough.

tarebyte:
@dinever just one last comment and I think this will be good to ship.

Would you mind updating this line https://github.com/education/classroom/blob/master/app/controllers/organizations_controller.rb#L119

To the following

```ruby
classroom: Organization.unscoped.includes(:users).find_by(github_id: membership.organization.id)
```

That way we don't have to make an extra SQL call to grab all of the users.

Thanks!

dinever:
Hi @tarebyte,

Just got that line modified.

Thanks!

tarebyte:
Thanks @dinever!

spec/controllers/organizations_controller_spec.rb
johndbritton: > his student user is only used in the context user without admin privilege on the organization now. So I think student should be enough.

Sorry, pardon my confusion.

Issue #442, opened 2016-01-22
Octospinner gets stuck when opening in new tab


johndbritton:
If you go to an assignment listing page and cmd+click on an assignment link to open it in a new tab, the octospinner starts, and never stops.

franktank:
Hi John,

I see that in the organization index view, @organizations is being rendered, but I am not sure where the octospinner is called.

tarebyte:
Hi @franktank :wave:

The markup for the spinner is here:
https://github.com/education/classroom/blob/master/app/views/layouts/application.html.erb#L16

The coffeescript is here:
https://github.com/education/classroom/blob/master/app/assets/javascripts/navigation.coffee

Hope this helps!

franktank:
Can a temporary fix be to put a timer on the loading spinner?

johndbritton:
> Can a temporary fix be to put a timer on the loading spinner?

I'd prefer that it not be triggered at all in this case.

franktank:
So in the event that the user is trying to open a new tab, disable the octospinner feature?

johndbritton:
That's the desired behavior.

tarebyte:
> So in the event that the user is trying to open a new tab, disable the octospinner feature?

But in this case @johndbritton @franktank we lose the spinner entirely for the page load.

It would be :cool: if we could use https://github.com/turbolinks/turbolinks-classic#progress-bar to make this change.

franktank:
Okay so I was thinking of checking for event of ctrl key being pressed, and if it is hide the octospinner.

johndbritton:
Fixed in https://github.com/education/classroom/pull/506

PR #506, opened 2016-03-19
[fix] Hide Octospinner when opening in a new tab/new window.

dinever:
Hi guys,

This is a fix to #442.

Any feedback would be greatly appreciated.

Thanks! :smile:

### DIFF
--- a/app/assets/javascripts/navigation.coffee
+++ b/app/assets/javascripts/navigation.coffee
@@ -1,5 +1,12 @@
 ready = ->
-  $('.js-navigation').on('click', -> $('.loading-indicator').show())
+  $('.js-navigation').on('click', (event) ->
+    if (event.ctrlKey or # New tab (windows)
+    event.shiftKey or # New window
+    event.metaKey or # New tab (mac)
+    (event.button and event.button == 1) # New tab (middle click)
+    )
+      return
+    return $('.loading-indicator').show())
 
 $(document).ready(ready)
 $(document).on('page:load', ready)
--- a/app/assets/javascripts/navigation.coffee
+++ b/app/assets/javascripts/navigation.coffee
@@ -1,12 +1,16 @@
+# A new window/tab event can be one of the following
+# ctrlKey (Windows)
+# metaKey (OSX)
+# shiftKey (OS agnositc)
+# New tab with middle click
+is_new_window_or_tab_event = (event) ->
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/navigation.coffee
```diff
@@ -1,5 +1,12 @@
 ready = ->
-  $('.js-navigation').on('click', -> $('.loading-indicator').show())
+  $('.js-navigation').on('click', (event) ->
+    if (event.ctrlKey or # New tab (windows)
+    event.shiftKey or # New window
+    event.metaKey or # New tab (mac)
+    (event.button and event.button == 1) # New tab (middle click)
+    )
+      return
+    return $('.loading-indicator').show())
```
tarebyte: So this is a good start, and I like the documentation, but what do you think about something like this?

I think extracting it out into it's own method makes it a little more readable.

Just a style note I'm a fan of guard clauses

```coffee
return if foo
# continue on
```

Also coffeescript much like Ruby the last line of the method is returned as the result. So the last result is not necessary.

```coffee
# A new window/tab event can be one of the following
# ctrlKey (Windows)
# metaKey (OSX)
# shiftKey (OS agnositc)
# New tab with middle click
new_window_or_tab_event = (event) ->
  middle_click_event = (event.button && event.button == 1)
  event.ctrlKey || event.metaKey || event.shiftKey || middle_click_event

ready = ->
  $('.js-navigation').on('click', (event) ->
    return if new_window_or_tab_event(event)
    $('.loading-indicator').show())
```

dinever:
Hi @tarebyte,

You are right. I had the style improved.

Thanks!

johndbritton:
This looks good to me. Thanks for the contribution @dinever.

@tarebyte If you verified this, I say go for it. Maybe file another issue about the Turbolinks 3 progress bar / spinner you wanted to upgrade to for the future.

tarebyte:
I can confirm this works locally, thanks @dinever

johndbritton:
> This is a fix to #442.

@dinever Just an FYI, if you use the term `Fixes #442`, `Closes #442` or similar in your pull request, the issue will automatically be closed when the pull request is merged. https://github.com/blog/1506-closing-issues-via-pull-requests

dinever:
Thanks @johndbritton, I'll remember that.

PR #508, opened 2016-03-20
Autocomplete for starter code repo

### LINKED ISSUE #480: Autocomplete for starter code repo
It'd be great if the starter code repo textbox autocompleted based repo names (instead of having the type it exactly correct).

cyhsutw:
Hello guys,

This should fix #480. 

Here's a short demo:

autocomplete

- By default, it shows the repositories that the user can access using the [list your repos API](https://developer.github.com/v3/repos/#list-your-repositories).
- Or, if the user put type some string like `education/t`, it will search the the repositories under namespace education (in this case, it will show `education/teachers_pet` and `education/teachers`).
- It utilizes the [repository search API](https://developer.github.com/v3/search/#search-repositories) directly for other queries like `rails`.

### Issues with Testing

Previously, I have the `AutocompleteController` generate JSON responses and use CoffeeScript to create suggestion elements.
After digging into the code base, I found using the `render_partial` might be more extensible for future additions like searching `users`.

After this change, it becomes hard to write tests since we will deal with the HTML contents which is relatively harder to validate.

One possible solution is making the controller support both HTML and JSON and the tests will check the JSON responses instead of HTML responses. 

Is there any good way to make the `AutocompleteController` more testable?

------

I would love to participate GSoC 2016 and this is my first PR for the Classroom. I would be very grateful if you can provide any advice/feedback :smile: 

Thank you!


config/routes.rb
```diff
@@ -12,6 +12,9 @@
   match '/auth/:provider/callback', to: 'sessions#create',  via: [:get, :post]
   match '/auth/failure',            to: 'sessions#failure', via: [:get, :post]
 
+  # autocomplete
```
tarebyte: This comment isn't really needed, would you mind taking it out?

app/views/group_assignments/_group_assignment_form_options.html.erb
```diff
@@ -19,7 +19,10 @@
 
 <dl class="form">
   <dt><label>Add your starter code from GitHub <small>(optional)</small> </label></dt>
-  <dd><input class="textfield" type="text" name="repo_name" value="<%= group_assignment.full_name %>" placeholder="<%= group_assignment.full_name || 'e.g. education/classroom'%>"</dd>
+  <dd><input id="autocomplete-textfield" class="textfield" type="text" name="repo_name" value="<%= group_assignment.full_name %>" placeholder="<%= group_assignment.full_name || 'e.g. education/classroom'%>"></dd>
+
+  <!-- autocomplete suggestions -->
```
tarebyte: I'm not really a big fan of comments in the markup could you also remove this comment?

app/views/assignments/_assignment_form_options.html.erb
```diff
@@ -19,5 +19,8 @@
 
 <dl class="form">
   <dt><label>Add your starter code from GitHub <small>(optional)</small> </label></dt>
-  <dd><input class="textfield" type="text" name="repo_name" value="<%= assignment.full_name %>" placeholder="<%= assignment.full_name || 'e.g. education/classroom'%>"</dd>
+  <dd><input id="autocomplete-textfield" class="textfield" type="text" name="repo_name" value="<%= assignment.full_name %>" placeholder="<%= assignment.full_name || 'e.g. education/classroom'%>" autocomplete="off"></dd>
+
+  <!-- autocomplete suggestions -->
```
tarebyte: Ditto https://github.com/education/classroom/pull/508/files#r56770268

app/assets/stylesheets/application.scss
```diff
@@ -14,3 +14,6 @@
 @import "stafftools";
 
 @import "remodal/dist/remodal";
+
+// autocomplete
```
tarebyte: The `@import` is pretty self explanatory here, would mind removing this comment?

tarebyte:
@cyhsutw would you mind using this spinner instead?

octocat-spinner-32

app/assets/javascripts/autocomplete.coffee
```diff
@@ -0,0 +1,69 @@
...
+update_textfield = (list_element) ->
+  # remove focused state
+  $(list_element).removeClass('suggestion-focused')
+  $('#autocomplete-textfield').val($(list_element).find('.res-name')[0].innerHTML)
+
+ready = ->
+  # TODO: test with IE 8
+  # onpropertychange -> oninput for IE 8
+  # https://msdn.microsoft.com/en-us/library/ms536956(v=vs.85).aspx
+  $('#autocomplete-textfield').on('focus input propertychange', ->
```
tarebyte: GitHub doesn't support IE 8 https://help.github.com/articles/supported-browsers/, so we don't have to worry about testing for that.

app/controllers/autocomplete_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    keyword = query.gsub(%r{^[a-zA-Z0-9_-]+\/}, '') + ' in:name fork:true'
+
+    # add namespace criteria if needed
+    if query.include? '/'
+      namespace = query.split('/').first
+      keyword += " user:#{namespace}"
+    end
+    current_user.github_search_client.search_repos(keyword, per_page: 10, page: 1)[:items]
+  end
+end
```
tarebyte: So I think this controller can be cleaned up a little bit, we can remove some of the logic away from the `#search_repos` action.

```ruby
class AutocompleteController < ApplicationController
  def search_repos
    repos = search_github_repos(params[:query])

    respond_to do |format|
      format.html { render partial: 'autocomplete/repository_suggestions', locals: { repos: repos } }
    end
  end

  private

  def github_search_client
    @github_search_client ||= current_user.github_search_client
  end

  def search_github_repos(query)
    return github_search_client.repos(nil, sort: 'updated', per_page: 10, page: 1) unless query.present?

    keyword = query.gsub(%r{^[a-zA-Z0-9_-]+\/}, '') + ' in:name fork:true'

    # add namespace criteria if needed
    if query.include? '/'
      namespace = query.split('/').first
      keyword += " user:#{namespace}"
    end

    github_search_client.search_repos(keyword, per_page: 10, page: 1)[:items]
  end
end
```

config/routes.rb
```diff
@@ -12,6 +12,9 @@
...
+  get '/autocomplete/repo_name', to: 'autocomplete#search_repos'
```
tarebyte: What do you think about

```ruby
get '/autocomplete/github_repos', to: 'autocomplete#search_github_repos'
```

app/assets/javascripts/autocomplete.coffee
tarebyte: I'm also not a huge fan of `id` based javascript bindings.

Generally what I do is add another class to the DOM object and add a `js-` prefix to it.

Example.

```html
<dd><input class="textfield js-autocomplete-textfield" ...>...</dd>
```

So then your coffee would look like
```coffee
$('.js-autocomplete-textfield').on('focus input propertychange', ->
```

app/assets/javascripts/autocomplete.coffee
```diff
@@ -0,0 +1,69 @@
...
+
+  $('#autocomplete-textfield').on('blur', ->
+    # check if there's any selected item
+    selected_items = $('.autocomplete-suggestion-item.suggestion-focused')
+
+    # if so, update the value of the textfield
+    update_textfield(selected_items.first()) if selected_items.length == 1
+
+    # hide suggestion list
+    $('#autocomplete-suggestions-container').hide()
```
tarebyte: All of these comments aren't needed, the code gives a pretty clear picture of what is happening.

app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,71 @@
...
+  background: #fff;
+  border: 1px solid #c1c1c1;
+  border-radius: 3px;
+  box-shadow: 0 0 5px rgba(0, 0, 0, 0.3);
+
+  // behaviors
+  overflow-x: hidden;
+  overflow-y: scroll;
+
+  // list
```
tarebyte: This comment isn't really needed, it doesn't add anymore context to what is going on.

app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,71 @@
...
+  ul {
+    list-style: none;
+
+    // item
```
tarebyte: ditto https://github.com/education/classroom/pull/508/files#r56771369

app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,71 @@
...
+      cursor: pointer;
+
+      &.suggestion-focused {
+        background-color: $brand-blue;
+        color: #fff;
+      }
+    }
+  }
+
+  // octicon
```
tarebyte: ditto https://github.com/education/classroom/pull/508/files#r56771369

app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,71 @@
...
+  .octicon {
+    vertical-align: text-top;
+  }
+
+  .octicon-lock {
+    color: #e9dba5;
+  }
+
+  // spinner
+  .autocomplete-loading-indicator {
```
tarebyte: ditto https://github.com/education/classroom/pull/508/files#r56771369

app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,71 @@
...
+    height: 72px;
+    background: image-url("octocat-spinner-32.gif") no-repeat center center;
+  }
+}
+
+
+// blank status
+.autocomplete-suggestions-empty {
```
tarebyte: https://github.com/education/classroom/pull/508/files#r56771369

app/assets/javascripts/autocomplete.coffee
```diff
@@ -0,0 +1,69 @@
...
+
+        # hide indicator
+        $('#autocomplete-loading-indicator').hide()
+
+        # Resolve an issue with `onBlur`.
+        # When clicking the item in the suggestion list,
+        # the `onBlur` event will take place before `onClick`.
+        # So when a user hovers the item, we can mark that item first
+        # and retrieve the hovered item when `onBlur` is fired.
+        $('.autocomplete-suggestion-item').hover(->
```
tarebyte: Can we use a `js-` prefix class here so we don't mix javascript with css styling?

app/assets/javascripts/autocomplete.coffee
```diff
@@ -0,0 +1,69 @@
...
+          $(this).addClass('suggestion-focused')
+        , ->
+          $(this).removeClass('suggestion-focused')
```
tarebyte: You defined `$(this)` here https://github.com/education/classroom/pull/508/files#diff-a34de611c45045bee0ea980cd2c0471cR19, but it looks like it never gets used.

tarebyte:
@cyhsutw this is looking really good so far, I left some comments.

Regarding the testing we don't really do any markup testing right now.

What I would do is make controller tests that verify that you get search results from the controller action.

cyhsutw:
Hi @tarebyte thanks for reviewing my code.

Your comments are really helpful for me to improve my coding styles.

I've updated my code according to your suggestions, please review them when you are available.

Thank you! :smile:

app/views/assignments/_assignment_form_options.html.erb
```diff
@@ -19,5 +19,6 @@
 
 <dl class="form">
   <dt><label>Add your starter code from GitHub <small>(optional)</small> </label></dt>
-  <dd><input class="textfield" type="text" name="repo_name" value="<%= assignment.full_name %>" placeholder="<%= assignment.full_name || 'e.g. education/classroom'%>"</dd>
+  <dd><input class="textfield js-autocomplete-textfield" type="text" name="repo_name" value="<%= assignment.full_name %>" placeholder="<%= assignment.full_name || 'e.g. education/classroom'%>" autocomplete="off" data-autocomplete-search-endpoint="github_repos"></dd>
```
cyhsutw: Previously, we used the `name` attribute to determine the endpoint to send request to. 

As we changed the `routes.rb`, I then add a custom [`data-*`](https://www.w3.org/TR/html5/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes) attribute to specify the endpoint.

app/assets/javascripts/autocomplete.coffee
```diff
@@ -0,0 +1,54 @@
+delay = do ->
+  timer = 0
+  (callback, ms) ->
+    clearTimeout timer
+    timer = setTimeout(callback, ms)
+    return
+
+update_textfield = (list_element) ->
+  $(list_element).removeClass('suggestion-focused')
+  $('.js-autocomplete-textfield').val($(list_element).find('.res-name')[0].innerHTML)
```
johndbritton: What do you think about modifying the form to submit `repository_id` when using autocompletion instead of sending `name/owner`?

I know it's an unlikely case, but it's possible that a user could rename the repository in between the time of choosing the item from the list and submitting the page. By using `repository_id` we avoid that

cyhsutw: :+1: 

This is a possible case, I would change the name if I am not satisfied with it :stuck_out_tongue:

What do you think about adding a hidden `input` with `repository_id` as `value` and we can use the value if it's present?

johndbritton: And it also just feels cleaner to reference by `repository_id`

app/models/user.rb
```diff
@@ -39,6 +39,11 @@ def github_client_scopes
     GitHubUser.new(github_client, uid).client_scopes
   end
 
+  # for searching, disable pagination so we can control the number of suggestions
+  def github_search_client
```
johndbritton: I don't love that we're adding another method to the `User` here. Why is this the right place for this code?

tarebyte: I think the reasoning here is because we need the users `access_token` to view their private repositories

johndbritton: Yeah, I see that we need the access token from the user, but why does this client belong on the user? Couldn't you also do something in the controller like this

```ruby
def search_client
  Octokit::Client.new(access_token: current_user.token, auto_paginate: false)
end
```

I'm not sure why, but I don't really like adding this method here to use it in just one case. Alternatively, what about adding an options hash argument to `User#github_client` that would allow you to disable pagination with that method?

app/assets/javascripts/autocomplete.coffee
johndbritton: > What do you think about adding a hidden input with repository_id as value and we can use the value if it's present?

Yeah, I like this idea. By default send `owner/name` but if it's selected via autocomplete also send a hidden `repository_id` and use that if it's supplied. You're going to need to make sure you have tests for this as well.

I'd check with @tarebyte to make sure he likes this plan before implementing.

app/models/user.rb
cyhsutw: Actually we can access the `token` attribute outside the class, so we can create the client in the controller.

My concerns were about concurrency, but I forgot that HTTP is stateless :sweat_smile:

> Alternatively, what about adding an options hash argument to User#github_client that would allow you to disable pagination with that method?

I thinks this is more reasonable :+1:

tarebyte: @johndbritton that sound fine to me

@cyhsutw if you could memoize the `seach_client` that'd be ✨ 

```ruby
def search_client
  @search_client ||= Octokit::Client.new(access_token: current_user.token, auto_paginate: false)
end
```

cyhsutw: @johndbritton @tarebyte 

Hi @johndbritton 

What about declaring `github_client` like this:

```ruby
def github_client(opts = {})
  @github_client ||= Octokit::Client.new(opts.merge(access_token: token, auto_paginate: true))
end
```

Then we can have our `search_client` like this:

```ruby
def search_client
  @search_client ||= current_user.github_client(auto_paginate: false)
end
```

cyhsutw: Oops! Please ignore my comment :arrow_up: 

Looks like it doesn't work since we have conditional assignment on `@github_client`, which is assigned before the `search_client` is called.

cyhsutw:
Hi @johndbritton @tarebyte 

I've updated the code based on your suggestions, please review them when you are available.

Thank you! :smile:

cyhsutw:
> What I would do is make controller tests that verify that you get search results from the controller action.

@tarebyte does this mean that we only test if the endpoint works?

I tried to write some tests for the `AutocompleteController`, here's what I got so far:

```ruby
RSpec.describe AutocompleteController, type: :controller do
  let(:organization) { GitHubFactory.create_owner_classroom_org }
  let(:user)         { organization.users.first                 }

  before do
    sign_in(user)
  end

  describe 'GET #github_repos', :vcr do
    it 'returns success and returns repos of the user' do
      get :github_repos
      expect(response).to have_http_status(:success)
      expect(response).to render_template(partial: 'autocomplete/_repository_suggestions')
    end

    it 'returns success and returns search results of a query' do
      get :github_repos, query: 'rails'
      expect(response).to have_http_status(:success)
      expect(response).to render_template(partial: 'autocomplete/_repository_suggestions')
    end
  end
end
```

I don't know if it's a proper way to test this controller.

Could you please give me some advice?

Thanks! :smile:

tarebyte:
@cyhsutw sorry for not being more clear, I meant test that a known search gives you back the results you expect.

Like `rails/rails` returns the ruby on rails project.

app/assets/javascripts/autocomplete.coffee
tarebyte: That sounds fine to me, we have validations that the repo belongs to the user on the controller so I'm not worried about anyone spoofing us.

app/controllers/autocomplete_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    @repos = search_github_repos(params[:query])
+
+    respond_to do |format|
+      format.html { render partial: 'autocomplete/repository_suggestions', locals: { repos: @repos } }
+    end
+  end
+
+  private
+
+  def required_scopes
```
johndbritton: How does this behave if we don't have the right scopes? Since we're requesting this endpoint via ajax, there's no opportunity to ask for more scopes.

cyhsutw: @johndbritton 

Currently, there isn't any error handling in this controller (which is bad), I will try to add some of them to make it more robust.

johndbritton: the `#required_scopes` method is used to request more scopes for the user if necessary. In this case it doens't really make sense, since it's happening in the context of an ajax request.

cyhsutw: Sorry for misunderstanding the meaning of the method. I'll pushed a fix later tonight. 

Thank you for pointing this out.

cyhsutw:
@johndbritton @tarebyte 

Thank you very much for taking time to review my code :smile: 

I've update my code based on your suggestions.

Short summary:

- Add some specs to test the functionality of `AutocompleteController`. 
- Ability to pass/handle `repo_id` when using autocomplete (as suggested [here](https://github.com/education/classroom/pull/508#discussion-diff-56806759R10))
- Remove misuse of `#required_scopes`.

cyhsutw:
> johndbritton [had a problem deploying](https://heaven.githubapp.com/apps/classroom/deployments/539749/log) to production 23 hours ago

@johndbritton I saw this message in this PR, I'd like to know if there's anything I can help with?
(I've check the [log](https://heaven.githubapp.com/apps/classroom/deployments/539749/log) but it shows `Not Found`)

tarebyte:
@cyhsutw that's because the log it GitHub employee only 😄 

We were trying to see if we could use our internal tools to branch deploy a fork, it doesn't look like we can though

cyhsutw:
@tarebyte Thanks for the explanation! :smile:

tarebyte:
@cyhsutw can you merge master back in one more time? I think this is good to 🚢

spec/controllers/assignments_controller_spec.rb
```diff
@@ -66,6 +66,23 @@
         expect(flash[:error]).to eql('Invalid repository name, use the format owner/name')
       end
     end
+
+    context 'repo_id for starter_code is passed' do
```
johndbritton: You need one more context here `invalid repo_id for starter_code is passed`, this could be the case of a repository the user doesn't have access to or an invalid input for `repo_id` like the string `abc` or `0`.

spec/controllers/autocomplete_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+RSpec.describe AutocompleteController, type: :controller do
+  let(:organization) { GitHubFactory.create_owner_classroom_org }
+  let(:user)         { organization.users.first                 }
+
+  before do
+    sign_in(user)
+  end
+
+  describe 'GET #github_repos', :vcr do
+    it 'returns success and returns a subset of repos of the user' do
```
johndbritton: best practice here is to have a single `expect` per `it` block. That way the test output is more descriptive when something fails.

johndbritton: So you could split this up into multiple `it` blocks and extract any repeated setup to a `before` block.

spec/controllers/autocomplete_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(response).to have_http_status(:success)
+
+      returned_repo_ids = assigns(:repos).map { |repo| repo[:id] }.to_set
+      actual_repo_ids = user.github_client.repos.map { |repo| repo[:id] }.to_set
+      expect(returned_repo_ids).to be_subset(actual_repo_ids)
+
+      expect(response).to render_template(partial: 'autocomplete/_repository_suggestions')
+    end
+
+    it 'returns success and returns search results of a query' do
```
johndbritton: Same thing here, you can split up your `expect`s into separate `it` blocks.

johndbritton:
@cyhsutw Left a few more inline notes, but this is looking very good. Once those last few things are addressed I think this is ready to deploy.

cyhsutw:
@johndbritton @tarebyte Thanks for spending time reviewing my code.

I've updated the code based on your suggestions.

> You need one more context here invalid repo_id for starter_code is passed, this could be the case of a repository the user doesn't have access to or an invalid input for `repo_id` like the string `abc` or `0`.

Actually, the controller DOES create assignments with those invalid `repo_id`s!
To address this, I added a [method](https://github.com/education/classroom/pull/508/files#diff-063449d33fb1ea54624f576bba84869f) in `concerns/starter_code.rb` to check if the user has proper access to that repo.

johndbritton:
@cyhsutw Thank you for your contribution, this has been deployed to production.

cyhsutw:
@johndbritton @tarebyte Thank you for taking time reviewing my code, helped me a lot!

PR #510, opened 2016-03-21
Custom 404 page

OrthoDex:
I added a custom 404 page, which notifies the user of 

- A general page not found; or

- An incorrect invitation link

I have run into a hurdle writing specs for this, majorly because the authenticate_user! function runs before any action and is returning a 302 which is failing the tests.


app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ErrorsController < ApplicationController
+	def error404
+		render template: "errors/simple_404.html.erb", status: :not_found
+	end
+end
```
tarebyte: @OrthoDex for this project we use 2 spaces for the formatting files and we have a newline on the end.

Also what do you think about this?

```ruby
class ErrorsController < ApplicationController
  def not_found
    render template: "errors/not_found.html.erb", status: :not_found
  end
end
```

app/views/errors/simple_404.html.erb
```diff
@@ -0,0 +1,18 @@
...
+						<% end %>
+						<% else %>
+						<p id="error">Page Could not be Found.</p>
+						<p>You may have mistyped the address or the page may have moved.</p>
+					<% end %>
+					</h2>
+			</div>
+		</div>
+	</div>
+</div>
```
tarebyte: Same commend with file ending and 2 spaces instead of tabs

config/routes.rb
```diff
@@ -12,6 +12,8 @@
   match '/auth/:provider/callback', to: 'sessions#create',  via: [:get, :post]
   match '/auth/failure',            to: 'sessions#failure', via: [:get, :post]
 
+  match "/404", :to => "errors#error404", :via => :all, as: :not_found
```
tarebyte: I'd like to see this look like `match "/404", to: 'errors#not_found`, via: :all, as: :not_found.

lib/not_found_exception.rb
```diff
@@ -0,0 +1,11 @@
...
+	def call(env)
+    	status = env["PATH_INFO"][1..-1]
+
+    	if status == "404"
+      		Rails.application.routes.call(env)
+    	else
+      		super
+    	end
+  	end
+end
```
tarebyte: Same comment with 2 spaces and line endings

config/application.rb
```diff
@@ -66,5 +66,10 @@ class Application < Rails::Application
         end
       end
     end
+
+    #Setting invoked exceptions application for custom 404 error pages
+    require Rails.root.join("lib/not_found_exception")
+    config.exceptions_app = NotFoundException.new(Rails.public_path)
```
tarebyte: This can actually be done with this line.

```ruby
config.exceptions_app = self.routes
```

So the entire lib file is uneeded.

lib/not_found_exception.rb
tarebyte: From my comment https://github.com/education/classroom/pull/510/files#r56814616 we don't need this file anymore

tarebyte:
@OrthoDex I can look at the specs for you later today, I left some comments that I think will be helpful.

Can you throw some screenshots in this PR to show us what this looks like?

app/views/errors/simple_404.html.erb
tarebyte: Also from my comment https://github.com/education/classroom/pull/510/files#r56814029 we can rename this to `not_found.html.erb`

tarebyte:
Also it doesn't look like the `octocat.png` in the public assets is being used.

spec/integration/errors_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    it 'shows general page not found message' do
+    	get '/not-existing-page'
+    	expect(response.status).to eq(404)
+    	within '#error' do
+    		has_content? 'Page Could not be Found.'
+    	end
+    end
+  end
+end
```
tarebyte: Same comments about line anding and 2 spaces vs tabs

OrthoDex:
@tarebyte 
Here is the general 404:

screenshot 404

And this is the invalid link 404:

404 invalid link

One thing worth mentioning is that the invalid link takes quite some time to render, I had to click on the 'you are being redirected' link to get the page.

Thank you so much for your feedback!

tarebyte:
@OrthoDex I'm not a fan of the flash with the status page, can we just have the text on the page?

OrthoDex:
@tarebyte I removed the flash message, on a side note I am still getting the issue of the invalid link 404 rendering really slow. 
Would love to know your thoughts on the matter :smile:

tarebyte:
@OrthoDex which page is causing it to render slowly?

OrthoDex:
@tarebyte The redirection from an invalid link to the 404 page is rendering slowly.
shot 1

shot 2

PR #516, opened 2016-03-23
Settings buttons look too similar

### LINKED ISSUE #513: Identical 'Settings' buttons in the same view
screenshot 2016-03-23 10 07 15

There's two identical `Settings` buttons. I know, I know – if you pay enough attention, they are in different contexts / ui boxes:
- the 'Kipos' classroom header, and
- the ui box owned by the 'test' assignment.

Still, it doesn't feel like a good solution, it's not _awesome_ nor it's straightforward to understand. The fact that the buttons are so apart from their modifying elements doesn't make it any easier.

Renaming the first one's label to `Classroom Settings` helps, but it's just a patch, not a solution.

---

Also, any reason not to add the octicon gear icon?

tarebyte:
Closes #513 

screen shot 2016-03-23 at 10 30 57 am

screen shot 2016-03-23 at 10 31 16 am

/cc @jasonlong @johndbritton

### DIFF
--- a/app/views/assignments/show.html.erb
+++ b/app/views/assignments/show.html.erb
@@ -16,7 +16,7 @@
       </div>
 
       <div class="flex-table-item settings">
-        <%= link_to edit_organization_assignment_path(@organization, @assignment), class: 'btn btn-green js-navigation right' do %>
+        <%= link_to edit_organization_assignment_path(@organization, @assignment), class: 'btn js-navigation right' do %>
           <%= octicon('gear') %>
           Assignment settings
         <% end %>
--- a/app/views/assignments/show.html.erb
+++ b/app/views/assignments/show.html.erb
@@ -16,7 +16,10 @@
       </div>
 
       <div class="flex-table-item settings">
-        <%= link_to 'Settings', edit_organization_assignment_path(@organization, @assignment), class: 'btn btn-green js-navigation right' %>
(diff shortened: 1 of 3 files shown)


jasonlong:
Looking at it, I might try making the assignment settings button gray. Either that or a regular link with the icon.

tarebyte:
Option 1
screen shot 2016-03-23 at 10 40 47 am

Option 2
screen shot 2016-03-23 at 10 44 40 am

@jasonlong here are your two suggestions

jasonlong:
I think I prefer the gray button. What do you guys think? Regardless, I think this direction solves the problem of the similar looking buttons pretty well.

tarebyte:
I'm good with the button

dasilvacontin:
@jasonlong @tarebyte All buttons in `site-content-cap` look like actual buttons, with a border. The color difference may be good since 'Settings' is not a 'creation' action – easier to associate with the outcome.

dasilvacontin:
Still, 'Update your permissions' is not a 'creation' action, yet the button is green. But it's the main action of that dialog. Settings is not even an action, links to a secondary view..

Gray button and text changes LGTM. Looks nice and it's much easier to understand the buttons' purpose. :) :+1:

johndbritton:
Looks great, I'm happy with either.

tarebyte:
alrighty gray buttons it is, thanks everyone!

PR #520, opened 2016-03-24
git_hub -> github

tarebyte:
I don't know why I never did this before, this adds the acronym `GitHub` so that we don't have to make file names like `git_hub` anymore.

/cc @johndbritton


config/initializers/inflectors.rb
```diff
@@ -0,0 +1,3 @@
+ActiveSupport::Inflector.inflections do |inflect|
+  inflect.acronym 'GitHub'
```
johndbritton: I didn't even know this was a thing. So good.

johndbritton: @mkcode FYI

mkcode:
This is smart. I wonder if core/rails would accept a PR for this. Surely it more useful than 'sheep' :sheep: 

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflections.rb

mkcode:
Never mind. Core infections are frozen.

PR #521, opened 2016-03-25
Organize GitHub lib

tarebyte:
We've had the `/app/models/concerns/github.rb` file hanging around for a while now.

I feel like it's dumping ground for logic that we (I) didn't know where to put.

This moves that logic into the lib folder and separates the where it is more understandable.

/cc @johndbritton


config/initializers/github_client.rb
```diff
@@ -0,0 +1,11 @@
+module Classroom
+  def self.github_client(options = {})
+    client_options = {
+      client_id:     Rails.application.secrets.github_client_id,
+      client_secret: Rails.application.secrets.github_client_secret,
+      auto_paginate: true
```
johndbritton: Doesn't look like https://github.com/education/classroom/pull/521/files#diff-9955ca00c1bcd788fb13273ca8ace816L243 had `auto_paginate` enabled. Do you expect any issues with that?

johndbritton:
:+1:

config/initializers/github_client.rb
tarebyte: I don't just because most of the resources we get are singlular

Issue #319, opened 2015-10-08
Handle repository already exists on GitHub


johndbritton:
We've seen a few cases where something strange happened and a user was in a state where they were unable to accept an invitation because a repository already existed within an organization with the same name as the repository we were trying to create.

Let's handle this case gracefully for the user. Maybe give them the option to connect to the existing repository or just create a new repository with a slightly modified name.

PR #522, opened 2016-03-25
Fix double flashes in `group_assignment_invitations/accept.html.erb`

cyhsutw:
While trying to reproduce #319, I found that `group_assignment_invitations/accept.html.erb` would render 2 flash messages :zap: :zap: when the repo name already exists.

double_flash

The extra flash message (the lower one) is removed in the PR.

### DIFF
--- a/app/views/group_assignment_invitations/accept.html.erb
+++ b/app/views/group_assignment_invitations/accept.html.erb
@@ -12,8 +12,6 @@
 </div>
 
 <div class="site-content-body">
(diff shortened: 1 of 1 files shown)


johndbritton:
:+1:

tarebyte:
Thanks @cyhsutw can you merge master in and then I can ship it?

cyhsutw:
@tarebyte Just rebased and passed the checks.

PR #524, opened 2016-03-26
Extract search logic to GitHub::Search

tarebyte:
We didn't wrap the search client in error handling and it's blowing up Airbrake.

This still needs tests, but I wanted to see what you all thought of this approach.

/cc @johndbritton @cyhsutw


app/controllers/autocomplete_controller.rb
```diff
@@ -1,29 +1,10 @@
 class AutocompleteController < ApplicationController
   def github_repos
```
johndbritton: I have to say, I like this controller a lot more now.

lib/github/search.rb
```diff
@@ -0,0 +1,35 @@
...
+      @search_client = Octokit::Client.new(options)
+    end
+
+    def search_github_repositories(query, options = { per_page: 10, page: 1 })
+      GitHub::Errors.with_error_handling do
+        if query.present?
+          search_query = build_github_repositories_query(query)
+          search_client.search_repos(search_query, options)[:items]
+        else
+          search_client.repos(@github_id, sort: 'updated', per_page: 10, page: 1)
```
johndbritton: Should this use the `options` that are passed in above?

tarebyte: yup, just a copy paste error

johndbritton:
:shipit:

app/assets/javascripts/autocomplete.coffee
```diff
@@ -11,7 +11,7 @@ update_textfield = (list_element) ->
   $('.js-autocomplete-resource-id').val($(list_element).data('res-id'))
 
 ready = ->
-  $('.js-autocomplete-textfield').on('focus input', ->
+  $('.js-autocomplete-textfield').on('change keyup', ->
```
cyhsutw: Hi @tarebyte 

Would you mind elaborating a little on using `keyup` instead of `input`? :pray: 

The reason why I used `input` is to ignore control events like `command + left arrow` (jumps to the beginning of the line).

cyhsutw:
@tarebyte 

Sorry about missing the error handling :disappointed_relieved:

This looks way far better than the previous version, thanks! :tada:

app/assets/javascripts/autocomplete.coffee
johndbritton: The Javascript was triggering a search after every single character was typed, which led to exhausting the GitHub Search API [rate limit](https://developer.github.com/v3/search/#rate-limit) of 30req/sec in some cases where a user was indecisive.

cyhsutw: @johndbritton Thanks for the explanation :smile: 

> The Javascript was triggering a search after every single character was typed

I thought this is controlled by the timeout period we set, not because the difference between  `keyup` and `input`. My understanding is that `keyup` would trigger more events than `input` when a user is typing (since `input` is triggered only when the `value` is changed).

> which led to exhausting the GitHub Search API rate limit of 30req/sec in some cases where a user was indecisive

A possible case that would exhaust the API is that a user keeps doing `focus` and `blur` (and this is a likely case).
In this case, `focus input` will trigger a lot of unnecessary requests compared to `change keyup`.

I was trying to follow up the design considerations here, it would be appreciated if you could give me some insights to help me clarify this.

PR #523, opened 2016-03-26
Added Elasticsearch status to ``GET /_ping``

sahilshekhawat:
We use Chewy for Elasticsearch, checking health of cluster tells us
Elasticsearch's status. i.e. ``Chewy.client.cluster.health["status"] == "green"``

Fixes #424

PR #526, opened 2016-03-27
Make `_ping` for `elasticsearch` to actually report the status.

cyhsutw:
Hey guys,

Thanks to @sahilshekhawat for #523 that added the ability to check the status of `elasticsearch`.

I've tried this locally but I found that it does not report the status correctly when my `elasticsearch` cluster is having `yellow` or `red` as its status.

This PR fixes the issue described above.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -66,7 +66,7 @@ module Classroom
         end
 
         ping.check :elasticsearch do
-          Chewy.client.cluster.health["status"] == "green"
+          raise unless Chewy.client.cluster.health['status'] == 'green'
           'ok'
         end
       end
--- a/config/application.rb
+++ b/config/application.rb
@@ -66,7 +66,12 @@ module Classroom
         end
 
         ping.check :elasticsearch do
-          Chewy.client.cluster.health['status'] == 'green' ? 'ok' : false  
+          status = Chewy.client.cluster.health['status'] || 'unavailable'
+          if status == 'green'
+            'ok'
+          else
+            raise "Elasticsearch status is #{status}"
+          end
         end
       end
     end
--- a/config/application.rb
+++ b/config/application.rb
@@ -66,8 +66,7 @@ module Classroom
         end
 
         ping.check :elasticsearch do
-          raise unless Chewy.client.cluster.health['status'] == 'green'
-          'ok'
+          Chewy.client.cluster.health['status'] == 'green' ? 'ok' : false  
         end
       end
     end


tarebyte:
@cyhsutw we don't want to raise errors, what is it reporting?

cyhsutw:
@tarebyte 
Thanks for the suggestion.

After digging in to the source code of [`pinglish`](https://github.com/jbarnette/pinglish/blob/master/lib/pinglish.rb#L117) I found that it's better to return `false` instead of raising an error when the status is not `green`. 

I've updated my code accordingly.

sahilshekhawat:
When a service fails, it is returned as ``failures: [memcached, elasticsearch]`` etc. This is actually an design issue, if we want to return false (which makes it difficult to see failing services comparatively) or return failed services.

cyhsutw:
@sahilshekhawat I thought returning `false` in the `ping.check` block will generate a message just like you mentioned, since `pinglish` will check if there's any `exception` or if the returned value is `false`, either way it's considered as a failure.

I tried this locally, and here's the response of `GET /_ping`:

1. Status of `elasticsearch`  clusters is `green`:

	```json
	{
		"now":"1459050390",
		"status":"ok",
		"db":"ok",
		"memcached":"ok",
		"redis":"ok",
		"elasticsearch":"ok"
	}
	```

2. Status of `elasticsearch` clusters is `red`/`yellow`:

	```json
	{
		"now":"1459050401",
		"status":"failures",
		"db":"ok",
		"memcached":"ok",
		"redis":"ok",
		"failures":["elasticsearch"]
	}
	```

config/application.rb
```diff
@@ -66,8 +66,7 @@ class Application < Rails::Application
         end
 
         ping.check :elasticsearch do
-          Chewy.client.cluster.health["status"] == "green"
-          'ok'
+          Chewy.client.cluster.health['status'] == 'green' ? 'ok' : false
```
johndbritton: How about this:

```ruby
ping.check :elasticsearch do
  status = Chewy.client.cluster.health['status'] || 'unavailable'
  if status == 'green'
    'ok'
  else
    raise "Elasticsearch status is #{status}"
  end
end
```

johndbritton:
Is there a reason you prefer using `false` rather than `raise`?

cyhsutw:
@johndbritton Since the design of `pinglish` treat `false` and `raise` the same ([see here](https://github.com/jbarnette/pinglish/blob/master/lib/pinglish.rb#L117)).

And I thought raising an exception is relatively costly than return `false`.

config/application.rb
cyhsutw: Looks like Pinglish ignores the message of the exception, it just returns the same message for `false` and `raise`.

I agree that this `?:` expression is really confusing, how about we just returned the `status` instead?

johndbritton:
> And I thought raising an exception is relatively costly than return false.

In each of the other checks, we're depending on exceptions. The general rule of thumb I like to use is if a case is exceptional it deserves an exception. Elasticsearch being down is an exceptional case.

tarebyte:
@cyhsutw I agree with @johndbritton let's go ahead and raise the error since it matches the other pinglish endpoints.

Thanks!

cyhsutw:
Just changed the code as @johndbritton suggested.

Just a gentle reminder, the response will remain the same as the one in my [previous comment](https://github.com/education/classroom/pull/526#issuecomment-201981845).

johndbritton:
> Just a gentle reminder, the response will remain the same as the one in my previous comment.

I think that's exactly the response we want. Is there something wrong with that?

cyhsutw:
Nope, it's all good. 

Just thought that you'd like to see exact status of `elasticsearch` because we passed a message in the `raise` call.

johndbritton:
> Just thought that you'd like to see exact status of elasticsearch because we passed a message in the raise call.

Ah, yeah. That's fine. Thanks for pointing it out

PR #528, opened 2016-03-28
Handle Search errors for Autocomplete

tarebyte:
So we wrapped the errors so that we handle them our way, but then since we were still raising them to the user we got stuff like this.

image

Now our errors will look like this.

image

/cc @johndbritton @cyhsutw


lib/github/errors.rb
```diff
@@ -4,50 +4,68 @@ class << self
...
+        error.errors.map do |err|
+          error_message = []
+
+          error_message << "#{err[:resource]}:"
+          error_message << err[:code].tr('_', ' ') if err[:message].nil?
+          error_message << err[:field] if err[:message].nil?
+          error_message << err[:message] unless err[:message].nil?
+
+          error_message.map(&:to_s).join(' ')
+        end.join(' ')
```
tarebyte: We now join all of the errors together instead of just the first one.

johndbritton:
:+1:

PR #529, opened 2016-03-28
Added restriction to new Assignment/GroupAssignment titles

### LINKED ISSUE #507: Assignments should not be named `new`
If a user creates an assignment named `new`, the URL for the assignment would be

```
/classrooms/:organization_id/assignments/new
```

which leads the user to the assignment creating page rather than the assignment page.

Same thing happens to group assignment too.

Potential solutions:
- Adding a `reserved word` restriction on the assignment name field, like the way GitHub handles it when user tries to create a user named `signin` or something.
  
   screen shot 2016-03-19 at 12 10 01 am
  
  or
- Moving the `assignments#new`route to another URL, which does not conflict with the `assignments#show` route.

sahilshekhawat:
Fixes #507

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -22,6 +22,7 @@ class Assignment < ActiveRecord::Base
 
   validates :title, presence: true
   validates :title, length: { maximum: 60 }
+  validates :title, exclusion: {in: AssignmentsController.action_methods.to_a}
 
   validates :slug, uniqueness: { scope: :organization_id }
 
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -22,7 +22,7 @@ class Assignment < ActiveRecord::Base
 
   validates :title, presence: true
   validates :title, length: { maximum: 60 }
-  validates :title, exclusion: {in: AssignmentsController.action_methods.to_a}
+  validates :title, exclusion: { in: AssignmentsController.action_methods.to_a }
 
   validates :slug, uniqueness: { scope: :organization_id }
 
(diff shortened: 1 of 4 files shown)


sahilshekhawat:
Let me know if I need to rebase this.

tarebyte:
@sahilshekhawat thanks for the contribution! I'd really love to see some tests to validate this works.

sahilshekhawat:
Sure, but I am not able to run tests locally since I don't have two github accounts and I need to fill up ``.env`` file. Can you suggest something as to how can I run tests locally?

tarebyte:
@sahilshekhawat making a second GitHub account doesn't cost anything. I would recommend making a username call notsahilshekhawat

sahilshekhawat:
I have added tests for two reserved words, i.e. 'new' and 'update' for Assignment and GroupAssignment.

app/models/assignment.rb
```diff
@@ -22,6 +22,7 @@ class Assignment < ActiveRecord::Base
 
   validates :title, presence: true
   validates :title, length: { maximum: 60 }
+  validates :title, exclusion: { in: AssignmentsController.action_methods.to_a }
```
johndbritton: I'm not a fan of this method of reserving worlds. If we want to make reserved words, there should be a list that we can manage and update.

johndbritton:
I'm not really a fan of creating a reserved words list for this fix, I'd much prefer updating our routes to avoid conflicts. We've got it on the todo list to fix a lot of our routes anyways.

sahilshekhawat:
I would love to fix routes, can you share with me (perhaps open an issue) about the things that needs to be fixed in routes? I can start working on it right away

johndbritton:
@sahilshekhawat Created https://github.com/education/classroom/issues/531 for info on ID based routes. Closing this in favor of updated routing.

PR #532, opened 2016-03-30
Change slug routing to id based routing

### LINKED ISSUE #531: Use id based routing throughout the app
Although the intention of having pretty urls is nice, it's caused way more trouble for us than it's worth. Let's move to a simpler, `id` based routing system.

The default routes for all resources should be based on `id`s.
1. Implement `id` based routes for all resources
2. Make sure slug based routes redirect to their new id based coutnerparts
3. (optional) As a follow-up, bring slug based routes back with user configured slugs rather than auto-generated slugs.

sahilshekhawat:
Default routes are id based while supporting slug based routes as well.
At this point, not redirecting to id based routes from slug based ones,
Need to discuss it with others.

Fixes #531


sahilshekhawat:
Right now, we don't have routes tests. Should I add them here or create a separate PR?

dinever:
@johndbritton I wonder if we should use a 301 redirection to redirect `slug` based routes to `id` based routes, or just showing users the page when they use `slug` based routes.

sahilshekhawat:
Yes, that exactly what i wanted to discuss. We can redirect to just show page with slug based routing.

tarebyte:
@sahilshekhawat @dinever personally I'd prefer a redirect just to let the user know what is going on in the application.

sahilshekhawat:
How can i fix linter error ``Class has too many lines`` for ``OrganizationController``?

tarebyte:
@sahilshekhawat looks like there is some duplication in the controller.

You can remove https://github.com/education/classroom/blob/72ae0d125999d8e09b14dca7de7eb74f316546e7/app/controllers/organizations_controller.rb#L77-L85

Since it is also written here https://github.com/education/classroom/blob/72ae0d125999d8e09b14dca7de7eb74f316546e7/app/controllers/concerns/organization_authorization.rb#L8-L16

sahilshekhawat:
I am not sure why it is redirecting to ``/login`` in tests. Maybe user is not logged in?

sahilshekhawat:
Something weird is happening here, as soon I change description in ``it`` block of a passing test, classroom redirects requests to ``/login`` (user is not logged in) and the test fails. Any clue as to why this might happen?

cyhsutw:
Probably an issue with cassettes. Try to reset them and see if it works.

tarebyte:
@sahilshekhawat just confirmed @cyhsutw suggestion you need to recreate the cassettes

app/controllers/assignments_controller.rb
```diff
@@ -61,7 +61,10 @@ def new_assignment_params
   end
 
   def set_assignment
+    @assignment = @organization.assignments.find_by!(id: params[:id])
+  rescue ActiveRecord::RecordNotFound
     @assignment = @organization.assignments.find_by!(slug: params[:id])
+    redirect_to action: params[:action], id: @assignment.id, organization_id: params[:organization_id]
```
tarebyte: be default `redirect_to` is a `302` Found status.

Let's make this a `301` Permanently moved because this is a permanent change.

You can do this by.

```ruby
redirect_to action: params[:action], id: @assignment.id, organization_id: params[:organization_id], status: 302
```

sahilshekhawat: Done.

sahilshekhawat:
Can someone help me out here. I am having trouble with cassettes, I deleted all cassettes so that they can be created again but then almost all of tests failed. Looks like newly created cassettes are causing tests to fail.

tarebyte:
@sahilshekhawat are the tests failing in the same way or in a different way?

sahilshekhawat:
In the same way, redirecting to /login

sahilshekhawat:
Okay, some of the points that i found are:

1. Not an cassettes issue, although tests were passing due to cassettes.
2. Turned off cassettes by adding:``WebMock.allow_net_connect!`` and ``VCR.turn_off!(:ignore_cassettes => true)`` to ``vcr.rb`` and results are same as with newly created cassettes.

3. Tests with same setting are failing in master too in same way, therefore, nothing specific to this PR.
4. Issue is with login. Somehow user is not able to log in inside tests and therefore, all the requests are being redirected to ``/login`` raising errors.

tarebyte:
@sahilshekhawat can you push up what you have and I can take a look?

sahilshekhawat:
Pushed and its failing due to bad credentials.

tarebyte:
did you push to your remote?

sahilshekhawat:
yeah, to this branch. https://github.com/education/classroom/pull/532/files#diff-7d1f3e8a2258c6af4e3ce66a24435766R88

https://travis-ci.org/education/classroom/builds/119872253#L567

tarebyte:
@sahilshekhawat that looks like something on our end not yours

sahilshekhawat:
Can someone please recreate cassettes in master on their local machine and check if its just me with failing tests?

tarebyte:
@sahilshekhawat I fixed your issues in a PR here: https://github.com/sahilshekhawat/classroom/pull/1

Let me know what you think.

spec/models/group_assignment_spec.rb
```diff
@@ -44,6 +44,22 @@
...
+        create(:group_assignment, organization: organization, title: 'new')
+      end.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Your assignment title is reserved')
+    end
+
+    it 'ensures RecordInvalid error when GroupAssignment title is \'update\'' do
+      expect do
+        create(:group_assignment, organization: organization, title: 'update')
+      end.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Your assignment title is reserved')
+    end
+  end
```
tarebyte: I think this can be made into a single test.

```ruby
describe 'exclusion of reserved words from exclusion from title' do
  let(:organization) { create(:organization) }
  it 'will not save if using an action_method as the title' do
    GroupAssignmentsController.action_methods.to_a.each do |action_method|
      expect do
        create(:group_assignment, organization: organization, title: action_method)
      end.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Your assignment title is reserved')
    end  
  end
end
```

spec/models/assignment_spec.rb
```diff
@@ -48,6 +48,22 @@
...
+        create(:assignment, organization: organization, title: 'new')
+      end.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Your assignment title is reserved')
+    end
+
+    it 'ensures RecordInvalid error when Assignment title is \'update\'' do
+      expect do
+        create(:assignment, organization: organization, title: 'update')
+      end.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Your assignment title is reserved')
+    end
+  end
```
tarebyte: Ditto https://github.com/education/classroom/pull/532/files#r58292255

spec/models/group_assignment_spec.rb
sahilshekhawat: Thats smart :+1: Thanks

sahilshekhawat: I apologize but this is leftover from my other branch. Removing this in favor of id based routing.

sahilshekhawat:
@tarebyte I think its mergeable. Let me know if there is something I missed.

johndbritton:
@sahilshekhawat I had a chat with @tarebyte about this and we agreed that the ideal routing isn't based on ids but based on login, ie `education.github.com/org_login`. The problem we face is that org logins can change.

My recommendation is to map out all the potential routes for one particular organization in the most ideal way possible and then we can spend some time figuring out how to make it work.

As a fallback, we can use id based routing, but the best user experience will be based on login, similar to how GitHub works. The current routing (id + title) isn't as nice as either of these solutions.

sahilshekhawat:
@johndbritton Awesome! :+1:  I was always more inclined to login based routing. I really like how Github handles it. There are many advantages of having the same routing as Github.

>The problem we face is that org logins can change.

If one change Organization name from Github, are we capable of dealing with it? If yes, then we can change routing based on that. If not, we can ask user that organization is not found and enter name/url of the organization and routes gets updates then and there. Let me know if I am missing something. Also, I will raise another PR with login based routing. You can close this one whenever you like. :smiley_cat:

johndbritton:
> If one change Organization name from Github, are we capable of dealing with it?

I think that's the crux of the issue we need to solve. If we had a good solution for this I think we would have gone with routing based on login from the beginning.

The simplest solution I can think of is to add a route that matches on login and use the login to hit the GitHub API and find the `github_id` to then load the appropriate classroom from the DB.

When generating URLs, we'd have to do the same thing in reverse: request the organization from the API using `github_id` and use the login in the response to generate the url.

The nice thing is that we cache API requests and can inspect the ETags header to avoid ratelimiting and keep it performant.

tarebyte:
Closing this due to staleness, but we can revisit this if we want.

PR #536, opened 2016-03-31
Override AssignmentRepo user method

tarebyte:
I was trying to look at a stafftools assignment_repo and realized we didn't cover the case where the user relation might not exist. Sigh.

This allows user to call `user` on an assignment_repo and if the user doesn't exist get the repo_access user

~~Needs tests.~~

/cc @johndbritton


johndbritton:
> we didn't cover the case where the user relation might not exist. Sigh.

Wish we could find a way to "finish" this migration so that we can avoid having two code paths.

tarebyte:
From my discussion with @cyhsutw here: https://github.com/cyhsutw/classroom/commit/7c1f460d37de3beb4ad3f160c368cfb8195b337d#commitcomment-16939964

I realized that we can actually fix another bug.

Since we are now able to call `assignment_repo.user` and return either the `user` or the `user` from the `repo_access` we can remove the `has_many :repo_accesses,    through:   :assignment_repos`.

```ruby
# Before
>> Assignment.first.users
  Assignment Load (0.2ms)  SELECT  "assignments".* FROM "assignments" WHERE "assignments"."deleted_at" IS NULL  ORDER BY "assignments"."id" ASC LIMIT 1
  User Load (0.5ms)  SELECT "users".* FROM "users" INNER JOIN "repo_accesses" ON "users"."id" = "repo_accesses"."user_id" INNER JOIN "assignment_repos" ON "repo_accesses"."id" = "assignment_repos"."repo_access_id" WHERE "assignment_repos"."assignment_id" = $1  [["assignment_id", 4]]
=> []
>>

# After
>> Assignment.first.users
  Assignment Load (0.2ms)  SELECT  "assignments".* FROM "assignments" WHERE "assignments"."deleted_at" IS NULL  ORDER BY "assignments"."id" ASC LIMIT 1
  User Load (0.3ms)  SELECT "users".* FROM "users" INNER JOIN "assignment_repos" ON "users"."id" = "assignment_repos"."user_id" WHERE "assignment_repos"."assignment_id" = $1  [["assignment_id", 4]]
=> [#<User:0x007f96f73c23d8
  id: 2,
  uid: 13497497,
  token: "the-token-was-here",
  created_at: Wed, 30 Mar 2016 14:45:39 UTC +00:00,
  updated_at: Wed, 30 Mar 2016 14:45:39 UTC +00:00,
  site_admin: false>]
```

@cyhsutw thoughts?

cyhsutw:
@tarebyte Looks great! So we don't have to modify the controller :tada:

PR #537, opened 2016-03-31
Add Frozen string literal to Classroom

tarebyte:
So in the spirit of making sure we have a happy path to upgrading this application to Rails 5 and Ruby 3, I turned on `frozen_string_literal: true` for all files in the application.

Why?

It looks like Ruby 3.0 will have frozen string literal on by default, here is the issue https://bugs.ruby-lang.org/issues/11473.

This is going to need some more testing to make sure everything works, but it looks alright just from the specs.

Thoughts?

/cc @johndbritton @mkcode


johndbritton:
> This is going to need some more testing to make sure everything works, but it looks alright just from the specs.

This will definitely need thorough testing, but assuming it works I don't see a reason for us not to do this.

> I turned on frozen_string_literal: true for all files in the application.

Is there a way to enable this without touching every file? Looks like a future version of ruby may have a flag that can be set to enable this.

If not, you might want to add a Rubocop rule that checks that this is enabled for every file.

mkcode:
I'd love to see a memory diff of using this! Going to pull this and see.

mkcode:
Here are the memory profiling results:

branch /  loaded                | string allocated mem | string retained mem | ObjSapce str stats |
:-------------------------------|:---------------------|:--------------------|:-------------------|
frozen-string-literal fresh     |        350160        |       186424        |       228311       |
frozen-string-literal warmed    |        3390069       |       188159        |       230410       |
master fresh                    |        3496091       |       185291        |       222966       |
master warmed                   |        3499968       |       185405        |       418776       |


branch /  loaded             | profile gc | profile memory | analyze memory |
:----------------------------|:---------------------|:--------------------|:-------------------|
frozen-string-literal fresh  | https://gist.github.com/0e9a7ad6f06788d9656df87c184c88ea | https://gist.github.com/dc30f4556f5439ca028a19376a57c2ce | https://gist.github.com/6a1c4c4e6ded8e090c948e9cacf388fb |
frozen-string-literal warmed | https://gist.github.com/46b438b3e42499681872fc18ba50817e | https://gist.github.com/ad2ae5d0629a0f68bb59fa1e80c98d29 | https://gist.github.com/eb071f1c50d2f66028852d7eacee00b9 |
master fresh                 | https://gist.github.com/2f0aec5b26720fd7514e0905b8dd1149 | https://gist.github.com/746324cf3f74926cd6a0989c5aa90f06 | https://gist.github.com/381fd7ecb1154579fac7ad3664239599 |
master warmed                | https://gist.github.com/c47f123e29e00064bdd614d28d69a05f | https://gist.github.com/9ffc6d6795bfedadf02172478d32066d | https://gist.github.com/ae847a8731184a90ac350a9418aa3ac2 |

ObjectSpace String stats:

master fresh:

```
String stats:
------------
3947 : file-digest://
2541 : /
995 :
671 : .coffee
521 : /Users/chris/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/better_errors-2.1.1/lib/better_errors/exception_extension.rb
```

master warmed:

```
String stats:
------------
17888 :
5527 : file-digest://
2527 : /
2370 : app/assets
2324 : /Users/chris/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/better_errors-2.1.1/lib/better_errors/exception_extension.rb
1620 : /Users/chris/src/projects/github/classroom/app/assets
```

frozen fresh:

```
String stats:
------------
3947 : file-digest://
2541 : /
1016 :
671 : .coffee
653 : /Users/chris/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/better_errors-2.1.1/lib/better_errors/exception_extension.rb
```

frozen warmed:

```
String stats:
------------
3947 : file-digest://
2541 : /
1018 :
788 : /Users/chris/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/better_errors-2.1.1/lib/better_errors/exception_extension.rb
671 : .coffee
```

tarebyte:
> Is there a way to enable this without touching every file? Looks like a future version of ruby may have a flag that can be set to enable this.

@johndbritton we can update `./script/*` to have the following option.

`RUBYOPTS=--debug-frozen-string-literal`

However if we set the `TargetVersion` of Ruby in Rubocop it will always complain that we don't have `#frozen_string_literal: true` at the top of all of our files.

--

Thanks for the memory profiling @mkcode this is definitely interesting to look at. I think we'd see an even bigger change if the entire Rails app was set to `frozen-string-literal`.

I think it's still worth pursuing this even with the results of the memory profiling. WDYT?

johndbritton:
@tarebyte I'm :+1: on this as long as you're sure it won't cause any problems for users. Might be worth branch deploying to monitor for a while.

mkcode:
> I think we'd see an even bigger change if the entire Rails app was set to frozen-string-literal.

This is a great point. The ratio of amount of Rails code to our app code is big.

> I think it's still worth pursuing this even with the results of the memory profiling. WDYT?

There was 2 benefits to this, the potential performance improvements and the future proofing of the code. The future proofing is for sure still valuable.

johndbritton:
@tarebyte anything holding us back from merging this?

tarebyte:
@johndbritton I'll give this another look over, I don't think there are any blockers.

tarebyte:
@johndbritton I removed the flag from the scripts just so we can take this incrementally.

PR #541, opened 2016-04-03
[WIP] Handle repository already existed on GitHub

### LINKED ISSUE #319: Handle repository already exists on GitHub

dinever:
Hi guys,

This repo fixes #319

If a repository already existed with the same name as the repository we were trying to create in the organization, we give users the option to join the existing repo with the same name or create a repo with a custom suffix(if user input `123` then we create a repo named `homework3-User-123`).

- If the user chooses to use the existing repo:

Use the existing repo

- If the user chooses to create a new repo:

Create a new repo

<del>I will do some style polish and add the test cases for this later. </del> :white_check_mark: 

Any feedback would be greatly appreciated. :smile: 

Thanks!


app/views/assignment_invitations/show.html.erb
```diff
@@ -12,16 +12,45 @@
...
         repository in the <%= link_to "@#{decorated_organization.login}", decorated_organization.github_url %> organization on GitHub.
       </p>
     </div>
-
     <div class="form-actions">
       <%= submit_tag 'Accept this assignment', class: 'btn btn-green js-navigation' %>
     </div>
+    <% end %>
   <% end %>
 </div>
```
dinever: Maybe I should extract content in the if/else block into separate `.erb` files.

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -34,7 +34,7 @@
                         title: 'ruby-project',
                         starter_code_repo_id: '1062897',
                         organization: organization,
-                        public_repo: false)
+                        public_repo: true)
```
dinever: Sadly I haven't yet received the discount pack I applied from GitHub education for me to create a private repo in an organization, so I changed these into public repo for recording VCRs.

I hope this does not cause any inconvenience and will change them back when I get the discount pack.

spec/controllers/stafftools/assignment_repos_controller_spec.rb
```diff
@@ -4,7 +4,12 @@
   let(:user)    { GitHubFactory.create_owner_classroom_org.users.first }
   let(:student) { GitHubFactory.create_classroom_student               }
 
-  let(:assignment) { create(:assignment, creator: user, organization: user.organizations.first, public_repo: false) }
```
dinever: Strange things happened when I was recording this test case. It's like every time I run this test the `assignment` here is a different one having a different name and I have to record a new VCR each time, so I changed it into a static one. However I don't know if it's fine.

app/controllers/assignment_invitations_controller.rb
```diff
@@ -4,8 +4,7 @@ class AssignmentInvitationsController < ApplicationController
   before_action :check_user_not_previous_acceptee, only: [:show]
 
   def accept_invitation
-    users_assignment_repo = invitation.redeem_for(current_user)
-
+    users_assignment_repo = invitation.redeem_for(current_user, params[:repo_name_suffix])
     if users_assignment_repo.present?
       redirect_to successful_invitation_assignment_invitation_path
     else
```
dinever: I reused this endpoint.

When there is a GitHub repo already exists with the same name:
- If `params[:repo_name_suffix]` is `nil`, we add the user to the repo as a collaborator.
- Else if `params[:repo_name_suffix]` is not `nil`, we create a new repo with `repo_name_suffix` and add the user to the repo.

johndbritton:
@dinever Thanks for the contribution here, I think this is going in the right direction but I think we can make it simpler for the user.

It's likely that the student won't know what's going on and as such may not know if connecting to an existing repository or creating a new one makes sense. 

@tarebyte What do you think makes most sense here?

Reconnecting to an existing repository seems useful, but is there ever a case where that would result in unintended behavior? I think the safest solution is to always create a new repository for the student.

dinever:
@johndbritton 

> I think we can make it simpler for the user.

I agree that this is too complex for users.

Maybe we can just automatically rename the repo by indicating a suffix number(like `assignment-user_login-1`) when there is a name conflict, and increase the suffix number(`assignment-user-login2`) when there is still a name conflict with the modified name.

\c @tarebyte for thoughts

dinever:
There is a situation where different assignments result in the same repository name:

1. Assignment name: `assignment`, user login: `c-programmer`, repo name: `assignment-c-programmer`.
2. Assignment name: `assignment-c`, user login `programmer`, repo name will be the same as above.

In this case, the second user has no idea what's going on and may join the existing repository, which results in unintended behavior.

tarebyte:
@dinever sorry I've been pretty absent on this thread, I should be able to look at this tonight and give you some thoughts here.

dinever:
@tarebyte No worries, take your time. :bow:

johndbritton:
> In this case, the second user has no idea what's going on and may choose to join the existing repository, which results in unintended behavior.

Based on this, I think we should always create a new repository. The nice thing about the way Classroom stores the repositories is that the repo could be renamed later without causing any problems for the app.

dinever:
Closed in favor of #561.

PR #544, opened 2016-04-03
Use Ajax call to load organizations on index page

### LINKED ISSUE #504: Use Ajax call to load organizations on index page.
https://github.com/education/classroom/pull/501 closed a #259, :zap: to @dinever 

Even with eager loading the users on the organizations the page load is rather slow.

It would be great if we could load the organizations using ajax so that the user isn't waiting for entire page to load.

/cc @johndbritton for thoughts.

cyhsutw:
Hello guys,

Just implemented to use AJAX calls to load organizations on index page.

Fixes #504.

Demo:

1. For new users without any classrooms available:

none

2. For returning users with some new classrooms (there's a tiny Octocat spinner on the right of the `Classrooms`):

some

It will be appreciated if you could give me some feedback/suggestion. :smiley: 

Thanks!

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -1,4 +1,3 @@
-# rubocop:disable ClassLength
 class OrganizationsController < ApplicationController
   include OrganizationAuthorization
 
@@ -113,6 +112,10 @@ class OrganizationsController < ApplicationController
       .merge(title: title)
   end
 
+  def organization_service
+    @organization_service ||= OrganizationService.new(current_user)
+  end
+
   def set_organization
     @organization = Organization.find_by!(slug: params[:id])
   end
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -44,6 +44,8 @@
     end
   end
 
+  get '/classrooms-async', to: 'organizations#async_index'
```
tarebyte: This should be scoped underneath line 34

app/controllers/organizations_controller.rb
```diff
@@ -1,17 +1,22 @@
...
-  skip_before_action :set_organization, :authorize_organization_access, only: [:index, :new, :create]
+  skip_before_action :set_organization, :authorize_organization_access, only: [:index, :new, :create, :async_index]
 
   decorates_assigned :organization
 
   def index
-    @organizations = current_user.organizations.includes(:users).page(params[:page])
+  end
+
+  def async_index
```
tarebyte: I'm not a fan of the name at all tbh.

@johndbritton do you have any other suggestions of what we could call this?

config/routes.rb
cyhsutw: Thanks! I'll fix this with my next commit. :smile:

app/controllers/organizations_controller.rb
cyhsutw: Actually I was using `list` but I found it's a little bit confusing (or maybe it's not). 

I would love to hear suggestions from you!

johndbritton:
@tarebyte I'll defer to you on if you think this is the right approach.

app/controllers/organizations_controller.rb
tarebyte: @cyhsutw what about `users_organizations`

app/controllers/organizations_controller.rb
```diff
@@ -1,17 +1,22 @@
+# rubocop:disable ClassLength
```
johndbritton: Maybe it's time to start breaking some of this logic out into `lib` with service objects.

app/controllers/organizations_controller.rb
johndbritton: I haven't used this kind of pattern before, so I'm sorry I'm not much help with the naming here. Are there any examples that we can look at elsewhere?

tarebyte: So this kind of matches this endpoint

https://github.com/education/classroom/blob/master/app/controllers/stafftools/resources_controller.rb#L8-L12

cyhsutw:
Hi @tarebyte @johndbritton 

Thanks for reviewing the code.

I've renamed the routes and method to `user_organization` and the endpoint is now at `/classrooms/user-organizations`.

Also, I extracted some of the code to `service/organization_service.rb`, not sure if I'm doing it right.

I would love to hear feedback/suggestion from you.

nwoodthorpe:
This PR has been stale for over a year now.

PR #545, opened 2016-04-03
[WIP] Kill decorators in Classroom

tarebyte:
This PR removes the draper gem and the decorator pattern in Classroom.

Some things to note here:

- [x] `GitHubResource` objects now use keyword arguments
- [x] We now pass tokens instead of `GitHubClient` objects
- [x] If we cannot find a resource with a users key we fallback on the Application client id and secrert. If that fails we fall back on the `NullGitHubResource`
- [x] GitHubResource attributes are created on the creation of the object.
- [x] Improved the speed of getting a users token for an organization
- [x] Test coverage is up to 90% :tada: 

/cc @johndbritton @mkcode for thoughts


johndbritton:
> If we cannot find a resource with a users key we fallback on the Application client id and secret.

This might not work well due to rate limits. Also, how does the application token behave with regards to access? Is it only able to request public data?

tarebyte:
> This might not work well due to rate limits.

There could be a problem, but there'd have to be over 5000 non accessible attributes in Classroom at any one given time for us to hit the rate limit.

> Is it only able to request public data?

I _think_ this is the case, I can't find the documentation to back up that claim though. If that's the case this would only be useful for public teams repos all users and orgs. Private teams and repos would be left out the loop.

Thoughts?

mkcode:
I am happy to see classroom get rid of it's dependence on draper. Not much of a fan of that gem, plus I usually see it used in place view models, not as model decorators like it was used in classroom.

Keywords args. :+1:

New interface looks a lot like the one that remote_resource provides. I know you were playing with that, I'm curious if you evaluated it for this work and then decided against it. And if so, then why?

Pretty amazing amount of code was deleted here. Especially on the stafftools index. Looks great.

tarebyte:
@mkcode basically I needed a little more flexibility then what the attributes portion gave me.

I didn't want to write custom methods for each attribute so that I could pass in extra params like `headers` an so on.

I also wanted to not have to change the way we were doing caching, it made sense to me to just keep it as is and not have to run an extra configuration step.

mkcode:
Ok. It was goal of mine to make remote-resource work for classroom as well as edu-web. To that end, I would have preferred that feedback was provided to improve remote-resource, rather than doing some redundant work, such as the attributes handling in `GitHubUserResource`.

Perhaps this can still happen in the future. This PR does give me clear picture of what is missing from it. Though I do see your point about changing the caching. There is also a benefit in this PR being more simple than the remote-resource gem. Maybe we can revisit this in the future, after it becomes more mature from use in edu-web. It would be great if at some point we could get classroom and edu-web running on the same set of GitHub API resources.

johndbritton:
> There could be a problem, but there'd have to be over 5000 non accessible attributes in Classroom at any one given time for us to hit the rate limit.

That will happen when we do things like rebuilding the search index.

> Essentially we get a higher rate limit but the permissions of an unauthenticated user.

That's reasonable. I just wanted to be 100% sure we're treated as an unauthenticated user in that case

> I also wanted to not have to change the way we were doing caching

This was a big one for me as well. That's not to say we can't move to `RemoteResource` in the future.

Gemfile
```diff
@@ -9,8 +9,6 @@ gem 'autoprefixer-rails'
 gem 'chewy'
 gem 'coffee-rails', '~> 4.1.0'
 
-gem 'draper'
```
johndbritton: :heart: :fire:

app/controllers/concerns/starter_code.rb
```diff
@@ -7,8 +7,8 @@ def starter_code_repository_id(repo_name)
 
     if repo_name =~ %r{^[a-zA-Z0-9_-]+\/[a-zA-Z0-9_-]+$}
       begin
-        github_repository = GitHubRepository.new(current_user.github_client, nil)
-        github_repository.repository(repo_name).id
+        github_repository = GitHubRepository.new(id: nil, access_token: current_user.access_token)
+        github_repository.repository(full_name: repo_name).id
```
johndbritton: I don't understand why we're using the `GitHubRepository#repository` instance method here. I think this would be more logical as a call to an instance method on an API client that was initialized with the correct token.

The way it reads now is "create a nil GitHub Repository and ask that repository for the id of some other repository" which is awful roundabout.

app/models/github_repository.rb
```diff
@@ -1,39 +1,30 @@
...
     end
   end
 
-  # Public
-  #
-  def full_name
-    GitHub::Errors.with_error_handling { @client.repository(@id).full_name }
+  def disabled?
+    return @disabled if defined?(@disabled)
+    @disabled = full_name == 'Deleted repository'
```
johndbritton: Can we avoid checking for this special string here?

app/models/github_team.rb
```diff
@@ -1,46 +1,41 @@
...
+  def team_repository?(github_repository:)
     GitHub::Errors.with_error_handling do
-      @client.team_repository?(@id, full_name)
+      client.team_repository?(id, github_repository.full_name)
     end
   end
+
+  def disabled?
+    return @disabled if defined?(@disabled)
+    @disabled = (slug == 'ghost')
```
johndbritton: Same here, can we detect disabled without relying on this string?

app/models/github_organization.rb
```diff
@@ -1,99 +1,58 @@
...
+    client.delete_team(github_team.id)
   end
 
-  # Public
-  #
-  def organization
-    GitHub::Errors.with_error_handling { @client.organization(@id) }
+  def disabled?
+    return @disabled if defined?(@disabled)
+    @disabled = (login == 'ghost')
```
johndbritton: again

app/models/github_user.rb
```diff
@@ -1,44 +1,51 @@
...
     []
   end
 
-  # Public
-  #
-  def login(options = {})
-    GitHub::Errors.with_error_handling { @client.user(@id, options).login }
+  def disabled?
+    return @disabled if defined?(@disabled)
+    @disabled = (login == 'ghost')
```
johndbritton: again

spec/lib/github/url_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+describe GitHub::URL do
+  subject { described_class }
+
+  describe '#avatar' do
+    it 'returns the correct with with a default size of 40' do
```
johndbritton: > returns the correct URL

spec/models/github_organization_spec.rb
```diff
@@ -1,63 +1,131 @@
...
+      end
+
+      describe '#create_repository', :vcr do
+        it 'successfully creates a GitHub Repository for the Organization' do
+          expect(WebMock).to have_requested(:post, github_url("/organizations/#{subject.id}/repos"))
+        end
+      end
+
+      describe '#delete_repository' do
+        it 'successfully creates a GitHub Repository for the Organization' do
```
johndbritton: Copypasta error?

spec/models/github_organization_spec.rb
```diff
@@ -1,63 +1,131 @@
...
-    after do
-      @client.delete_repository("#{organization.title}/#{@repo_name}")
+    describe '#add_membership' do
     end
 
-    it 'successfully creates a GitHub Repository for the Organization' do
-      @github_organization.create_repository(@repo_name, private: true)
-      expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/repos"))
+    context 'with GitHub repository' do
+      before(:each) do
```
johndbritton: do you need these before and after blocks in both cases? Seems like you're trying to delete a repository twice.

spec/models/github_team_spec.rb
```diff
@@ -1,37 +1,92 @@
...
+      let(:github_team_membership_url) { "teams/#{@github_team.id}/memberships/#{github_student.login}" }
+
+      describe '#add_team_membership' do
+        it 'adds a user to the given GitHubTeam' do
+          @github_team.add_team_membership(github_user: github_student)
+          expect(WebMock).to have_requested(:put, github_url(github_team_membership_url))
+        end
+      end
+
+      context 'with team membership' do
```
johndbritton: you're missing a `describe '#remove_team_membership'` somewhere/

johndbritton: And you should be testing all of the cases for each method.

spec/models/github_team_spec.rb
```diff
@@ -1,37 +1,92 @@
...
+      describe '#add_team_repository' do
+        it 'adds a repository to the team' do
+          @github_team.add_team_repository(github_repository: @repository)
+          expect(WebMock).to have_requested(:put, github_url(request_url))
+        end
+      end
+    end
+
+    describe '#team' do
+      it 'returns team the GitHub API' do
```
johndbritton: this `it` block is unclear to me.

spec/models/github_user_spec.rb
```diff
@@ -1,53 +1,53 @@
...
+    let(:other_github_user) do
+      other_user = GitHubFactory.create_classroom_student
+      other_user.github_user
     end
-  end
 
-  describe '#organization_memberships', :vcr do
-    it 'returns an array of organizations that the user belongs to' do
-      organization_memberships = github_user.organization_memberships
+    describe '#accept_membership_to' do
```
johndbritton: Missing some tests here?

spec/models/github_user_spec.rb
```diff
@@ -1,53 +1,53 @@
...
+    end
 
-      expect(WebMock).to have_requested(:get, github_url('/user/memberships/orgs?state=active'))
-      expect(organization_memberships).to be_kind_of(Array)
+    describe '#active_admin?' do
```
johndbritton: You should test all the cases here, not just the true case.

spec/models/github_user_spec.rb
```diff
@@ -1,53 +1,53 @@
...
+      it 'returns an Array of scopes' do
+        expect(github_user.client_scopes).to eq(%w(admin:org delete_repo repo user:email))
+      end
+    end
 
-      expect(github_user.user.id).to eql(id)
-      expect(WebMock).to have_requested(:get, github_url("/user/#{id}"))
+    describe '#disabled?' do
+      it 'returns true if the user is not present' do
+        missing_github_user = GitHubUser.new(id: 1_000_000_000, access_token: github_user.access_token)
```
johndbritton: maybe use an id here that we know is a missing user.

spec/models/github_user_spec.rb
```diff
@@ -1,53 +1,53 @@
...
-      id = @client.user.id
+    describe '#client_scopes' do
+      it 'returns an Array of scopes' do
+        expect(github_user.client_scopes).to eq(%w(admin:org delete_repo repo user:email))
+      end
+    end
 
-      expect(github_user.user.id).to eql(id)
-      expect(WebMock).to have_requested(:get, github_url("/user/#{id}"))
+    describe '#disabled?' do
```
johndbritton: Also need to test the opposite case.

spec/models/group_assignment_repo_spec.rb
```diff
@@ -65,9 +65,64 @@
       end
     end
 
-    describe '#creator' do
-      it 'returns the group assignments creator' do
-        expect(@group_assignment_repo.creator).to eql(group_assignment.creator)
+    describe '#disabled?' do
```
johndbritton: test for the false case

spec/models/null_github_organization_spec.rb
```diff
@@ -1,11 +1,13 @@
...
 
 RSpec.describe NullGitHubOrganization do
+  it_behaves_like 'a NullGitHubResource descendant with attributes'
+
   subject { described_class.new }
 
   describe '#html_url' do
     it 'returns #' do
-      expect(subject.html_url).to eql('#')
+      expect(subject.html_url).to eql('https://github.com/ghost')
```
johndbritton: changing this url will mean that links in classroom will go to @ghost. Do we want that?

spec/models/organization_spec.rb
```diff
@@ -10,6 +10,12 @@
     end
   end
 
+  describe '#access_token' do
+    it 'selects a users access_token at random' do
```
johndbritton: This test doesn't actually test what the description says.

`it 'returns an access_token'` would be more accurate.

johndbritton:
@tarebyte just went through this in detail and left you some feedback, looks pretty good.

The biggest thing I'd like to point out is that I think your test cases could be expanded a bit. For each method that you're testing, make sure you test the happy case and then cover the other possible cases as well.

tarebyte:
Closing in favor of breaking this up into smaller chunks

PR #558, opened 2016-04-06
fix subtle exit status bug

mkcode:
@tarebyte - after copying your script/setup contents I ran across a subtle bug that exists in classroom's code as well.

The `cp -n` returns non-zero status when the file already exists, which breaks out of the script because we have the '-e' parameter set.

You can verify this exists with: 

`cp -n .env-example .env; echo $?`

Cheers


tarebyte:
Thanks @mkcode!

PR #561, opened 2016-04-09
Handle repository already existed on GitHub

### LINKED ISSUE #319: Handle repository already exists on GitHub

dinever:
Fixes #319

When creating an assignment, if the repo name `#{org_login}-#{user_login}` already exists, simply append a suffix number `1` and try to create the repo with the modified name `#{org_login}-#{user_login}-1`.

If the modified repo name still exists, increase the suffix number by 1 and try again, e.g. `#{org_login}-#{user_login}-2`, `#{org_login}-#{user_login}-3`, etc.

Thanks!


spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -67,6 +67,45 @@
       end
     end
 
+    context 'github repository already exists' do
```
johndbritton: Can you clarify this a bit:

> GitHub repository with the same name already exists

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -67,6 +67,45 @@
...
+      let(:assignment_repo) { AssignmentRepo.new(assignment: assignment, user: user) }
+
+      before do
+        assignment_repo.create_github_repository
+      end
+
+      it 'creates a new assignment repo with a name suffix' do
```
johndbritton: Split this into two `it` blocks with one `expect` each.

1. 'creates a new assignment repo'
2. 'new repository name has expected suffix'

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -67,6 +67,45 @@
...
+        patch :accept_invitation, id: invitation.key
+        expect(WebMock).to have_requested(:get, github_url("/repos/#{classroom_owner_organization_github_login}"\
+                                                           "/#{assignment_repo.repo_name}-1"))
+        expect(user.assignment_repos.count).to eql(1)
+      end
+
+      context 'modified repository name still exists' do
```
johndbritton: This context is unclear

app/models/assignment_repo.rb
```diff
@@ -17,6 +17,7 @@ class AssignmentRepo < ActiveRecord::Base
 
   before_validation(on: :create) do
     if organization
+      assign_repo_name_suffix while github_repository?
```
johndbritton: This needs a better name, it's unclear what this does at first pass.

app/models/assignment_repo.rb
```diff
@@ -45,9 +46,21 @@ def github_team_id
...
+  end
+
+  # Public
+  #
   def repo_name
     github_user = GitHubUser.new(user.github_client, user.uid)
-    "#{assignment.slug}-#{github_user.login(headers: GitHub::APIHeaders.no_cache_no_store)}"
+    base_name = "#{assignment.slug}-#{github_user.login(headers: GitHub::APIHeaders.no_cache_no_store)}"
+    return base_name unless @repo_name_suffix_number.present?
+    "#{base_name}-#{@repo_name_suffix_number}"
```
johndbritton: What will happen when the suffix takes the name over the character limit for the GitHub API?

johndbritton:
@dinever Left some comments, I think this is the right direction but I'd encourage you to break as much of the logic as you can out of the controller.

You might want to create a class that handles repository naming and write a bunch of tests for that to make sure it always comes up with a unique, unused name that is always valid for GitHub.

dinever:
@johndbritton Thanks! Wokring on that.

lib/github/repo_name.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def repo_name
+      base_name = "#{@assignment_slug}-#{@user_login}"
+      return base_name if @suffix_number.zero?
+      suffix = @suffix_number.to_s
+      "#{base_name[0, 99 - suffix.length]}-#{suffix}"
+    end
+  end
+end
```
dinever: Explanation about how I handle the case when the length of the repo name exceeds the GitHub repository name limitation:

- GitHub repository name length limitation is 100 characters.
- GitHub user login length limitation is 39 characters. 
- Our assignment title length limitation is 60 characters.

The repository name will never exceed 100 characters in the formation of `#{assignment.slug}-#{user.login}`, but may exceeds the limitation when there is a suffix, so we need to make sure that the repository does not exceed when there is a suffix appended.

My solution is to truncate only the base name, keep the suffix and make sure that the total length is not longer than 100 characters. Suppose that the following title has 102 characters:

`super_long_..._assignment_title-super_long_..._user_login-1`

We truncate the last 2 characters of the base name so that the title does not exceed the 100 characters limitation:

`super_long_..._assignment_title-super_long_..._user_log-1`

In this approach we can make sure that the repository name does not exceed and suffix is not removed.

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -67,6 +67,31 @@
...
+
+        expect(WebMock).to have_requested(:post,
+                                          github_url("/organizations/#{organization.github_id}/repos"))
+          .with(body: /^.*#{assignment_repo.repo_name}-1.*$/)
+      end
+
+      after do
+        assignment_repo.destroy_github_repository
+      end
+    end
```
dinever: @johndbritton As the repo name logic is extracted into the class `GitHub::RepoName` and there are test cases for it. I wonder if we still need to keep this context.

dinever:
Hi @johndbritton @tarebyte,

I think this one is ready for review.

Thanks!

johndbritton:
@tarebyte should take a pass at this, once he does I'll go through one more time.

tarebyte:
@dinever I should be able to get to this today

lib/github/repo_name.rb
```diff
@@ -0,0 +1,24 @@
...
+      @user_login = user_login
+      @assignment_slug = assignment_slug
+
+      @suffix_number = 0
+      @suffix_number += 1 while repository?
+    end
+
+    def repository?
+      @client.repository?("#{@organization_login}/#{repo_name}")
+    end
```
tarebyte: I would rather this be move over to `GitHub::Repository` as a class method. So that you could do something like

```ruby
GitHub::Repository.present?
```

lib/github/repo_name.rb
tarebyte: @dinever I'm not really a big fan of this being a `GitHub` lib attribute.

It's not specifically related to a GitHub object or the GitHub API.

I think this might be better as a rails model concern (maybe `Nameable`) and make this functionality a little more generic so that we can handle both Assignments and GroupAssignments.

WDYT?

dinever: @tarebyte 
I agree that we should move this to a better place. Will work on it.

Thanks

dinever:
Hi @tarebyte,

Just got this reworked.

Changes:

1. Moved repo name logic into the model concern `Nameable`.
1. Made `Nameable` generic so that we can use it both in `AssignmentRepo` and `GroupAssignmentRepo`.
1. Moved GitHub repository API call to `GitHub::Repository.present?`.
1. Added test cases for `GroupAssignmentInvitationController` to make sure it handles duplicate repository name correctly when creating a group assignment repo.

Please review them when you are available. Thanks!

spec/controllers/stafftools/assignment_repos_controller_spec.rb
```diff
@@ -5,7 +5,9 @@
   let(:user)    { GitHubFactory.create_owner_classroom_org.users.first }
   let(:student) { GitHubFactory.create_classroom_student               }
 
-  let(:assignment) { create(:assignment, creator: user, organization: user.organizations.first, public_repo: false) }
+  let(:assignment) do
+    Assignment.create(title: 'HTML5', creator: user, organization: user.organizations.first, public_repo: false)
+  end
```
tarebyte: What was the reasoning for changing how we created the assignment?

lib/github/repository.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+module GitHub
+  class Repository
+    class << self
+      def present?(client, full_name)
+        client.repository?(full_name)
+      end
```
tarebyte: This should move to `app/models/github_repository.rb`

spec/controllers/stafftools/assignment_repos_controller_spec.rb
dinever: @tarebyte This is because I need to specify a title for this assignment, otherwise it will have a different title each time and I can not record the VCR for it.

dinever: @tarebyte BTW, I mentioned this in a previous comment. 
https://github.com/education/classroom/pull/541/files#r58305889

> Strange things happened when I was recording this test case. It's like every time I run this test the assignment here is a different one having a different name and I have to record a new VCR each time, so I changed it into a static one. However, I don't know if it's fine.

tarebyte: @dinever but you could also do

```ruby
create(:assignment, creator: user, organization: user.organizations.first, title: 'HTML5', public_repo: false)
```

tarebyte: @dinever that should be fine, not sure why you're having that issue.

dinever: @tarebyte Sure, I'll change it back to the original form but with a title

```
create(:assignment, creator: user, organization: user.organizations.first, title: 'HTML5', public_repo: false)
```

lib/github/repository.rb
dinever: @tarebyte Sorry for this one. I thought we are talking about `GitHub::Repository.present?` so I put it into the `GitHub` module. Just got this fixed.

app/models/github_repository.rb
```diff
@@ -32,6 +32,12 @@ def get_starter_code_from(source)
 
   # Public
   #
+  def self.present?(client, full_repo_name)
+    client.repository?(full_repo_name)
+  end
```
tarebyte: :+1:

tarebyte:
@dinever I have a ideas how to clean this up, what do you think about this diff?

```diff
diff --git a/app/models/assignment_repo.rb b/app/models/assignment_repo.rb
index c95c801..76dcdba 100644
--- a/app/models/assignment_repo.rb
+++ b/app/models/assignment_repo.rb
@@ -48,8 +48,7 @@ class AssignmentRepo < ActiveRecord::Base
   # Public
   #
   def repo_name
-    github_user = GitHubUser.new(user.github_client, user.uid)
-    @repo_name ||= github_repo_name(assignment.slug, github_user.login(headers: GitHub::APIHeaders.no_cache_no_store))
+    @repo_name ||= generate_github_repo_name
   end
 
   # Public
@@ -72,4 +71,15 @@ class AssignmentRepo < ActiveRecord::Base
   def user
     original_user || repo_access.user
   end
+
+  private
+
+  delegate :slug, to: :assignment
+
+  def name
+    return @name if defined?(@name)
+
+    github_user = GitHubUser.new(user.github_client, user.uid)
+    @name = github_user.login(headers: GitHub::APIHeaders.no_cache_no_store)
+  end
 end
diff --git a/app/models/concerns/nameable.rb b/app/models/concerns/nameable.rb
index 52e9c53..8143091 100644
--- a/app/models/concerns/nameable.rb
+++ b/app/models/concerns/nameable.rb
@@ -2,21 +2,38 @@
 module Nameable
   extend ActiveSupport::Concern
 
-  def github_repo_name(assignment_slug, user_login)
-    suffix_number = 0
-    suffix_number += 1 while
-        GitHubRepository.present?(organization.github_client, full_name(assignment_slug, user_login, suffix_number))
-    suffixed_repo_name(assignment_slug, user_login, suffix_number)
+  def generate_github_repo_name
+    @suffix_number = 0
+
+    loop do
+      break unless GitHubRepository.present?(organization.github_client, full_name)
+      @suffix_number += 1
+    end
+
+    suffixed_repo_name
+  end
+
+  private
+
+  def base_name
+    @base_name ||= "#{slug}-#{name}"
   end
 
-  def full_name(assignment_slug, user_login, suffix_number)
-    "#{organization.decorate.login}/#{suffixed_repo_name(assignment_slug, user_login, suffix_number)}"
+  def full_name
+    "#{organization_login}/#{suffixed_repo_name}"
   end
 
-  def suffixed_repo_name(assignment_slug, user_login, suffix_number)
-    base_name = "#{assignment_slug}-#{user_login}"
-    return base_name if suffix_number.zero?
-    suffix = suffix_number.to_s
-    "#{base_name[0, 99 - suffix.length]}-#{suffix}"
+  def suffixed_repo_name
+    return base_name if @suffix_number.zero?
+
+    suffix = "-#{@suffix_number}"
+    base_name.truncate(100 - suffix.length, omission: suffix)
+  end
+
+  def organization_login
+    return @organization_login if defined?(@organization_login)
+
+    github_organization = GitHubOrganization.new(organization.github_client, organization.github_id)
+    @organization_login = github_organization.login(headers: GitHub::APIHeaders.no_cache_no_store)
   end
 end
diff --git a/app/models/github_organization.rb b/app/models/github_organization.rb
index 7efd251..b424c0b 100644
--- a/app/models/github_organization.rb
+++ b/app/models/github_organization.rb
@@ -71,8 +71,8 @@ class GitHubOrganization
 
   # Public
   #
-  def login
-    GitHub::Errors.with_error_handling { @client.organization(@id).login }
+  def login(options = {})
+    GitHub::Errors.with_error_handling { @client.organization(@id).login(options) }
   end
 
   # Public
diff --git a/app/models/group_assignment_repo.rb b/app/models/group_assignment_repo.rb
index 53e586c..7be971c 100644
--- a/app/models/group_assignment_repo.rb
+++ b/app/models/group_assignment_repo.rb
@@ -52,10 +52,7 @@ class GroupAssignmentRepo < ActiveRecord::Base
   # Public
   #
   def repo_name
-    headers     = { headers: GitHub::APIHeaders.no_cache_no_store }
-    github_team = GitHubTeam.new(creator.github_client, github_team_id).team(headers)
-
-    @repo_name ||= github_repo_name(group_assignment.slug, github_team.slug)
+    @repo_name ||= generate_github_repo_name
   end
 
   # Public
@@ -63,4 +60,16 @@ class GroupAssignmentRepo < ActiveRecord::Base
   def starter_code_repo_id
     group_assignment.starter_code_repo_id
   end
+
+  private
+
+  delegate :slug, to: :group_assignment
+
+  def name
+    return @name if defined?(@name)
+
+    headers     = { headers: GitHub::APIHeaders.no_cache_no_store }
+    github_team = GitHubTeam.new(creator.github_client, github_team_id).team(headers)
+    @name = github_team.slug
+  end
 end
```

dinever:
@tarebyte This looks way better. I will try & test it. Thanks!

dinever:
@tarebyte 

Just got this whole thing refactored. Thanks!

app/models/concerns/nameable.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def full_name
+    "#{organization_login}/#{suffixed_repo_name}"
+  end
+
+  def suffixed_repo_name
+    return base_name if @suffix_number.zero?
+
+    suffix = "-#{@suffix_number}"
+    base_name.truncate(100 - suffix.length) + suffix
```
tarebyte: If you skip the omission it truncates your string with a "..." right?

dinever: @tarebyte Yeah, but if we use `suffix` as the omission here. The suffix will only be added when the string is truncated, which means that the suffix will never show up if the string length is less than 100.

dinever: @tarebyte Maybe I can set the omission as an empty string to avoid `...`.

tarebyte: @dinever 👍🏼

app/models/github_repository.rb
```diff
@@ -32,6 +32,12 @@ def get_starter_code_from(source)
 
   # Public
   #
+  def self.present?(client, full_repo_name)
```
tarebyte: Can we update this to just be `full_name`? That way it matches the API response we get.

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -68,6 +68,30 @@
...
+      let(:assignment_repo) { AssignmentRepo.new(assignment: assignment, user: user) }
+
+      before do
+        assignment_repo.create_github_repository
+      end
+
+      it 'creates a new assignment repo' do
+        patch :accept_invitation, id: invitation.key
+        expect(user.assignment_repos.count).to eql(1)
+      end
```
tarebyte: Since you've already created the `assignment_repo` in the `let` hitting the `PATCH` route doesn't do anything. We can kill this test.

spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -68,6 +68,30 @@
...
+
+        expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/repos"))
+          .with(body: /^.*#{assignment_repo.repo_name}-1.*$/)
+      end
+
+      after do
+        assignment_repo.destroy_github_repository
+      end
+    end
+
```
tarebyte: This test was kind of confusing to me just because of how it was setup.

WDYT about something like this?

```ruby
context 'github repository with the name same already exists' do
  
  before do
    assignment_repo = AssignmentRepo.new(assignment: assignment, user: user)
    @original_repository = organization.github_client.repository(assignment_repo.github_repo_id)
    
    # use delete instead of destroy to skip active record callbacks
    assignment_repo.delete

    patch :accept_invitation, id: invitation.key
  end
  
  after do
    organization.github_client.delete_repository(@original_repository.id)
    AssignmentRepo.destroy_all
  end
  ...
```

dinever: Thanks @tarebyte, this looks pretty clear.

However, seems that I still need to put `assignment_repo` in the let as `assignment_repo.repo_name` is referred in the second test case:

```
it 'new repository name has expected suffix' do
  expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/repos"))
    .with(body: /^.*#{assignment_repo.repo_name}-1.*$/)
end
```

tarebyte: @dinever for this case your test would look like this.

```ruby
it 'adds a numerical suffix to the new repository name' do
  expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/repos"))
    .with(body: /^.*#{@original_repository.name}-1.*$/)
end
```

dinever:
@tarebyte All changes mentioned above have been done. Thanks!

spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -97,6 +97,30 @@
...
+        end
+
+        it 'new repository name has expected suffix' do
+          expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/repos"))
+            .with(body: /^.*#{@original_repository.name}-1.*$/)
+        end
+
+        after do
+          organization.github_client.delete_repository(@original_repository.id)
+          GroupAssignmentRepo.destroy_all
```
tarebyte: you're going to want to add `Group.destroy_all` here as well so that the misc team that was created is removed from GitHub.

dinever: Thanks @tarebyte, just have this added.

tarebyte:
Looks good to me, thanks @dinever!

dinever:
Thanks @tarebyte!

johndbritton:
Nice work on this @dinever 💥

PR #566, opened 2016-04-14
Update og-image to GitHub Classroom

### LINKED ISSUE #565: Update og-image to GitHub Classroom
Currently, our `og-image` uses the old name for this app "Classroom for GitHub":

og-image

We need to update it to "GitHub Classroom" to go along with https://github.com/education/classroom/pull/564

cyhsutw:
Fixes #565.

I've changed the title text of the `og-image` to `GitHub Classroom`.

Here's how it looks like now:

og-image


tarebyte:
Thanks @cyhsutw!

I think the color of the text is a bit off, can you use `#9FD1AF`?

tarebyte:
@cyhsutw thanks for the contribution!

I think the text color is off a bit (needs to be `#9FD1AF`) and the also make sure that we are using the latest octicons.

johndbritton:
@tarebyte The wordmark was pure white in the previous version as well I think. This is slightly different than what we had, but I think it works.

We'll likely revisit the share image when we update the homepage.

tarebyte:
@johndbritton :cool: that's fine. :shipit:

cyhsutw:
Update:

I [triggered the crawling process manually](https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fclassroom.github.com) after this PR was merged.
However, it does not update the `og:image` immediately.
After two days, Facebook finally refreshed the cache, and the new image is now live.

screen shot 2016-04-16 at 8 47 30 pm


According to this [StackOverflow question](http://stackoverflow.com/a/20262825/2695329), I think maybe we can change the name of the image a little bit if we need to update the `og:image` so the crawler will replace the cache immediately.

PR #568, opened 2016-04-14
Increased height of modal so the bottom is less cramped

ConnorCartwright:
Tiny update to the reset/remove modal to give the bottom a bit of spacing.


tarebyte:
Thanks @ConnorCartwright!

PR #569, opened 2016-04-15
Fix loading indicator positioning.

cyhsutw:
Hello guys,

Here's another minor UI tweak.

**Problem:**
When clicking a button located at the lower part of a long page, the loading indicator will not be placed at the right place.

5711071b61f44133576366

**About this PR:**
1. Make loading indicator spans the whole viewport.
2. Disable scrolling when loading indicator is shown.

**Result**


johndbritton:
@cyhsutw Thanks for this.

@tarebyte and I have been talking about moving to using the progress bar loading indicator that comes with Turbolinks. I'm not opposed to merging this, but if you're up for it I think doing that would be even better.

Issue #448, opened 2016-01-26
Implement "Student Identifiers" to map between GitHub users and Students


samwhitlock:
A feature request: an optional, text-field identifier associated with each student.

The use case: instructors want to the relation between GitHub users and students in their class. For grading purposes, instructors need an unambiguous way to associate these 2 identities for each enrolled student. Students often have similar / alternate spellings for their names, or don't even use their real name at all in their GitHub profile.

Instructors currently do this by having students fill out a separate form (_e.g._, Google forms / spreadsheet). 

To accommodate all possible use cases, this field could just be an additional text field (_e.g._, to ask for their school ID).

An option that would be quite useful along these lines is if this student ID could be the academic email addresses of students. In this mode, students would enter their academic email address and would be verified by an email to that address. It may also be desirable to limit the domain of the email address (e.g. only `epfl.ch`, `stanford.edu`, ...).

johndbritton:
> To accommodate all possible use cases, this field could just be an additional text field (e.g., to ask for their school ID).

This sounds reasonable to me. We'll consider giving the teacher the ability to make the field required.

> In this mode, students would enter their academic email address and would be verified by an email to that address.

I think it's unlikely that we will go so far as to send an email validation.

> It may also be desirable to limit the domain of the email address (e.g. only epfl.ch, stanford.edu, ...).

I think this is a separate feature, see #377.

johndbritton:
Would love to hear from other teachers about how helpful it would be to have a text-field identifier for the students and if there are other implementations that achieve the same in a superior way.

samwhitlock:
To elaborate more on what I'm using for my course(s):

* I use [this home-brewed github management application](https://github.com/EPFL-IC/sweng-management)
* It primarily relies on students filling out a google form, to fill out a copy of [the template spreadsheet](https://docs.google.com/spreadsheets/d/1lSOhkBQrs7RRY0a-IoyfQRqvfp2kWnB-XvMBX-omfUY/edit#gid=0).
* The main function of the spreadsheet is to link github IDs to the `GASPAR` usernames (the username for the Kerberos clone at my institution), which is how the app names individual assignments. There is some LDAP functionality in this script, but it's not necessary (just to fill populate info if we only have an email _or_ a student ID number).

This is somewhat hacky, but it's what we've been using.

ConnorCartwright:
Hey @samwhitlock @johndbritton 

I'm interested in giving this a crack but had a couple questions before starting.

[...]

PR #570, opened 2016-04-15
Updated creating organization so the 'complete' button also updates the information

ConnorCartwright:
Hey guys,

Whilst working on [#448](https://github.com/education/classroom/issues/448) I noticed that when you create an organization, if you update the organization name in the screen:

And click **_Continue_** rather than _**Save Changes**_ the organization name does not get updated.

I wasn't exactly sure where to put the code to fix this or if there's a better way so I'm open to suggestions. 

Currently I have removed the submit function from the save changes button and added a class to both buttons that I look for in the `update_organization.coffee` file and submit the form on click.

Cheers,
Connor


dinever:
Hi @ConnorCartwright 

> And click Continue rather than Save Changes the organization name does not get updated.

Seems that this is an intended behavior.

Actually, if clicking `Continue` also updates the organization, then the two buttons will have the same behavior and there is no need to keep both of them. 

Maybe we could just remove `Continue` to make things simpler?

ConnorCartwright:
Hey Shawn @dinever 

I think you're right - removing the `Continue` button would make things simpler.

I thought the `Save Changes` button didn't lead to allowing you to invite admins, but seeing as both buttons lead to the same page it would make more sense to remove 'Continue' (unless we plan on having more setup options later on).

johndbritton:
Yeah, I don't know why we have two buttons there.

@tarebyte any insight into this? I think `Continue` makes the most sense as the button to keep since this is part of the creation flow. Maybe this is caused because we want to display `Save changes` on the edit page and `Continue` in the creation flow.

app/assets/javascripts/update-organization.coffee
```diff
@@ -0,0 +1,2 @@
+$(".update_organization").on 'click', ->
+  $('form.edit_organization').submit();
```
tarebyte: Can you add a newline here? I need to add a JS linter to catch this kind of stuff.

app/views/organizations/setup.html.erb
```diff
@@ -17,8 +17,8 @@
       </div>
 
       <div class="form-actions">
-        <%= f.submit 'Save changes', class: 'btn btn-green' %>
-        <%= link_to 'Continue', invite_organization_path(@organization), class: 'btn', role: 'button' %>
+        <%= submit_tag 'Save changes', :type => 'button', class: 'btn btn-green update_organization' %>
```
tarebyte: Can we put this back to using `f.submit` since this button is part of the form?

tarebyte:
> Maybe this is caused because we want to display `Save changes` on the edit page and `Continue` in the creation flow.

I'm pretty sure that was why @ConnorCartwright let's make continue the only button on that page, and have it perform the save action.

ConnorCartwright:
@tarebyte I've removed the .coffeescript file and made the only button `Continue` :thumbsup:

app/assets/javascripts/update-organization.coffee
dinever: @tarebyte I'm adding the CoffeeScript linter right now.

tarebyte:
Looks good, thanks @ConnorCartwright!

PR #573, opened 2016-04-16
Add star and watch buttons

johndbritton:
screen shot 2016-04-16 at 5 34 24 pm

cc/ @tarebyte


tarebyte:
⛵️

tarebyte:
@johndbritton yeah know on second thought I'm not really digging the way that looks on the footer. It breaks up the text in a way that bugs me.

@jasonlong do you have an opinion on this?

johndbritton:
I was also thinking about dropping the 🎓 with ❤️ by GitHub text, now that the app is called "GitHub Classroom"

tarebyte:
@johndbritton do we want to update the footer to make it more official looking?

johndbritton:
>  to make it more official looking?

Sure, but don't know how we'd do that.

jasonlong:
My initial reaction is that it feels kind of strange to have these buttons in the footer – or on the site at all. What's the rationale for showing them? If it's to reiterate that the tool is open source, could we just make that text more clear? Having "open source" linked like it is now doesn't necessarily say "the code for this site is on GitHub" (not that we want to say exactly that).

johndbritton:
> What's the rationale for showing them?

The idea was to drive more people to the repo and get them to be more involved. I'm fine going without it.

jasonlong:
Sure, makes sense. Could we make the footer text more actionable then? Something like:

----

GitHub Classroom is an [open source project on GitHub.com](). Contributions are welcome and encouraged.

johndbritton:
Will update the footer in another pull request.

PR #575, opened 2016-04-18
Add Coffeelint

dinever:
Added a linter for Coffeescript.

Thanks

PR #577, opened 2016-04-20
Vendoring mobile device detection javascript

dinever:
Currently we're having a very long third-party js snippet used for mobile device detection in [app/assets/javascripts/clipboard-js.coffee](https://github.com/education/classroom/blob/master/app/assets/javascripts/clipboard-js.coffee#L3), of which the usage is [here](https://github.com/education/classroom/blob/master/app/assets/javascripts/clipboard-js.coffee#L26).

This snippet keeps us from introducing coffeelint(#575) into the project as it can not pass the linter requirement.  So let's replace it with [jquery-browser-plugin](https://github.com/gabceb/jquery-browser-plugin) and put it in `vendor`.

screen shot 2016-04-20 at 1 23 32 pm

Thanks! 🐸


dinever:
Closed in favor of #578

PR #578, opened 2016-04-21
Kill JS mobile detection

tarebyte:
This is an alternative to #577.

I'd like to kill the JS we have in place here, we use the check once and I don't think it's worth vendoring all of that just for one check.

Thoughts?

/cc @johndbritton @dinever


dinever:
👍  Great. Now we can get rid of the JS.

johndbritton:
Works for me.

PR #584, opened 2016-04-29
Remove staff access to organization pages

johndbritton:
This removes the ability for users with `#staff? == true` to access organization pages that they're not an admin of.

Previously, this was required for debugging. It's unnecessary now since we have an impersonation feature for staff.

I also noticed that this code is duplicated in `OrganizationAuthorization`.

I'm good with just removing this for consistency, but if you want to remove the duplication I'm for it @tarebyte.

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -76,7 +76,7 @@ class OrganizationsController < ApplicationController
   private
 
   def authorize_organization_access
-    return if @organization.users.include?(current_user) || current_user.staff?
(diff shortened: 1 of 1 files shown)


tarebyte:
@johndbritton, I think we can kill this entire method in the controller and lean on OrganizationAuthorization.

johndbritton:
@tarebyte Looks like that module was already being included so I just deleted the method.

tarebyte:
:shipit:

tarebyte:
Confirmed in production

PR #589, opened 2016-05-11
Disable rack timeout in devopment

mkcode:
Rack timeout can get rather annoying while debugging in development mode. This PR disables rack-timeout in development mode.


tarebyte:
Nice, thanks @mkcode!

PR #590, opened 2016-05-11
attack -> timeout

tarebyte:
/cc @mkcode


tarebyte:
Closed in favor of #591

PR #591, opened 2016-05-11
get the gem right this time

mkcode:
Use the correct name for the rack-timeout fixes in #590.


tarebyte:
👏

mkcode:
thanks for merging my version...I think

Issue #420, opened 2016-01-05
Invalid repository name, use the format owner/name


meling:
The text entry field marked *Add your starter code from GitHub (optional)* is not clear what is expected, and the example is not helpful. I've tried all the different path and URL combinations that I can think of to no avail. That is, I get the following error message:

    Invalid repository name, use the format owner/name

Suggestions?

tarebyte:
Hi @meling an example of the format `owner/name` would be `rails/rails` or `github/gitignore` it's the owner (organization or user) and the name of the repository.

Is there we can change the wording to make this more clear?

johndbritton:
@meling What would make the most sense to you for that field?

guntbert:
@johndbritton This had me stumped too for a while. I'd suggest
>use account/repository (e.g. education/classroom)

instead of the current
>e.g. education/classroom

(btw I only found out just now, that pressing <kbd>-></kbd> opens useful suggestions.)

cyhsutw:
@guntbert thanks for the feedback!

The text you suggested looks clearer, @johndbritton @tarebyte @dinever what do you think?

johndbritton:
:+1: @cyhsutw do you want to send a pull request?

dinever:
Looks good @cyhsutw. However, I'd suggest that we use *username* instead of *account* in that text:

> use username/repository (e.g. education/classroom)

guntbert:
@dinever not quite my intention, as the organization's name isn't a *username* per se.

johndbritton:
To keep consistent with the terminology on GitHub let's use

> login/repository

guntbert:
@johndbritton remember we are talking about people (like myself) who have not got a feeling for the overall concept and therefore want to follow suggested procedures very carefully. Now I would read
> login/repository

but my login is *guntbert*, whereas the organization that owns that repo is *psysKursRE*. So telling me to use my login will not clarify things for me.
<sub> I admit though that this **is** better than the current version</sub>

johndbritton:
@guntbert Your organization also has a login, "psysKursRE"

Maybe we're coming at this from the wrong angle. Rather than trying to make help text, how can we make the repository select box more clear?

guntbert:
@johndbritton Ah now I see more light :smile_cat: 

As is often the case we are using the same words for different things: 
- to me a *login* is something that somebody uses (probably along with a password..) to gain access to some ressource.
[...]

PR #592, opened 2016-05-13
Improve autocomplete suggestions for empty query

cyhsutw:
Use [list your repos](https://developer.github.com/v3/repos/#list-your-repositories) API for empty autocomplete query.

- remove github_id param from GitHub::Search
- update specs accordingly
- update cassettes accordingly


tarebyte:
@cyhsutw so these two endpoints should be delivering the same content, I'd prefer to keep the user ID, because adding a id to the URL makes it unique for when we cache the search results.

Are we getting different results with your change?

tarebyte:
@cyhsutw let me think on this one.

/cc @johndbritton

johndbritton:
@cyhsutw Can you provide some visuals of what this changes?

cyhsutw:
Here's the comparison:

| Before | After |
|:--------:|:-------:|
| before | after |

The difference here is that the new one shows the private repos of the user while the old one does not.

The reason is mentioned in @tarebyte's [comment](https://github.com/education/classroom/pull/592#issuecomment-219020163):

> https://developer.github.com/v3/repos/#list-user-repositories. It only lists public repositories.

cyhsutw:
Actually I think these changes are making the code a little bit more hard to read (with the `nil` param).

johndbritton:
@cyhsutw Did you see my comment here: https://github.com/education/classroom/issues/420#issuecomment-218819356

Maybe a better solution would be to improve what the user sees in the dropdown with some help text rather than just listing repositories. "Start typing a repository name" or similar

cyhsutw:
@johndbritton Thanks for the information. 

I also agree that your approach will be a much better solution. 

Close this in favor of #598.

Issue #597, opened 2016-05-20
Issue creating a team 


nilmestremas:
Hello,

Once I've chosen my GitHub team's name from a subject of my university degree, I get an error that says something like "You don't have the permit to acces this repository". What should I do?

Thanks,
Nil

johndbritton:
@nilmestremas This is a known issue. It looks like the organization owner has tested out the platform and accepted an assignment. Since we require fewer permissions from students his token was downgraded.

To solve the issue, please tell your teacher to log out and log back in to Classroom and to not accept an assignment invitation with an organization owner account.

johndbritton:
cc/ https://github.com/education/classroom/issues/445

Issue #588, opened 2016-05-10
Students cannot join assignments


pilla:
It is my second assignment for an Operating Systems class and students that were able to enter the first assignment cannot enter the second one. 

I first thought it might be that I was reusing the same set of groups of the former assignment and thus created a new assignment without reusing groups. Students still get the "forbidden" message when they try to get their shinny new repository. 

They can get as far as entering the group name and pressing the button to create the repo.

The assignment can be found at [this link](https://classroom.github.com/group-assignment-invitations/b03e973fff656e6e81e8a66f991a0fb5).

tarebyte:
@pilla you need to sign back in to https://classroom.github.com

Your token scope needs to be `admin:org, delete_repo, repo, user:email` but it currently is `admin:org, user:email`

Did you try to accept an assignment?

pilla:
I logged in again and was able to accept. But I am an admin, the problem is that students cannot accept it. Oddly enough, I was able to accept it using another account. I guess it may be an issue with my students' credentials. Will check.

pilla:
The same student that reported me the problem tried again and it worked. Closing as I cannot reproduce it, but looks like some weird transient fault.

johndbritton:
I think this is another case of https://github.com/education/classroom/issues/445

Issue #586, opened 2016-05-08
Documentation: tell teachers/classroom owners to create dummy-accounts


guntbert:
There have been several issues ( #461 ,...) where teachers tried to see/test an assignment and thereby triggered that bug #445.

I suggest to include something in the guide to tell <strike>them</strike> us
- create some dummy-accounts to play the roles of students
- maybe even suggest a naming-scheme
- best use different browsers when testing assignments (closing the browser may still not be sufficiet when switching roles because of caching)

johndbritton:
Thanks for the suggestion @guntbert. I'm hoping that we'll have a fix for that issue within the next couple of weeks. We really don't want to suggest creating multiple / dummy accounts. I'd rather just get this to work correctly.

guntbert:
@johndbritton My approach was primarily born out of the desire to

1. see what exactly my students will see (that need diminishes with growing experience on teaching with github, of course)
2. be able to demo the first steps of creating an account, joining the organization, accpeting assignments,...

I asked for permission before :smile:   

Wouldn't it be useful if github created two such accounts? Only one is needed for demoing but for seeing what students see of each other's work, of repos (what permissions are useful,...) I need a  2nd one.
<sub> I have no idea how to create passwords for those dummy-accounts, and about probably a bunch of further problems with this approach.</sub>

johndbritton:
> see what exactly my students will see 

Once this permission issue is resolved, you will be able to do this with your own account.

guntbert:
:+1: 
Please don't forget to write some guide on the appropriate procedure.

BTW: I haven't said my **very big thanks** to the whole team for providing us with such a great tool.
So here comes
I :bow: and :clap: , you are doing tremendously, thank you!!

Issue #581, opened 2016-04-25
Authentication issues


Omarasifshaikh:
I am seeing repeated errors in the site saying: "You are forbidden from performing this action on GitHub".

The students are also not being assigned to their repositories. They receive the error "You are forbidden from performing this action on github.com"

It appears that the repository is being created, the template is being cloned. But the individual student is not being assigned as a contributor.

Omarasifshaikh:
This is what a student is seeing:
<img width="582" alt="screen shot 2016-04-24 at 7 40 29 pm" src="https://cloud.githubusercontent.com/assets/6437181/14772548/8411dabc-0a54-11e6-8d29-fa8feb81d89b.png">

dinever:
Hi @Omarasifshaikh,

> I am seeing repeated errors in the site saying: "You are forbidden from performing this action on GitHub".

This is a known problem in GitHub Classroom and we are working on a solution.

This problem occurs when teachers try to accept assignments using their own GitHub account. When accepting an invitation, we reduce the permission scope we have for them and they will not be able to create assignments anymore using the reduced permission scope.

To resolve this issue, please log out from GitHub Classroom and log back in.

> It appears that the repository is being created, the template is being cloned. But the individual student is not being assigned as a contributor.

To temporarily fix this, you can manually add the student to the assignment repo as a contributor. 

---

Sorry for causing this trouble for you. We are currently working on this problem. However, to prevent it from happening again, please notice not to try out accepting the assignment invitation using teachers' GitHub account. 

Related: #445

Omarasifshaikh:
@dinever 
Thanks for the quick response. I have 55 students int the class, and they don't all accept the assignment at the same time. Thus would not be very practical to manually add them myself here.

Do you recommend me using teachers_pet for now?

https://github.com/education/teachers_pet

cyhsutw:
Hi @Omarasifshaikh

Is there any other teachers in your organizations?

tarebyte:
@Omarasifshaikh I looked at your teachers and all of them have the correct token scopes.

I'll investigate this today, sorry for the trouble.

Omarasifshaikh:
@tarebyte 
Thanks for verifying that. Classroom is a great app and I have enjoyed using it this semester. Please let me know If I can contribute in any way to fix the issue.

Thanks!!

johndbritton:
@Omarasifshaikh It looks like you have multiple admin users on your organization and that you have been using one of them for testing.

Please try the following steps and report back:
[...]

PR #604, opened 2016-05-26
Finally the fix the terrible token bug

### LINKED ISSUE #445: Don't reduce OAuth scopes
When visiting a page that requires fewer scopes, we currently ask the user to reduce permissions. This can lead to problems for organization owners, so let's stop reducing scopes.

What's happening is that if any organization administrator accepts an assignment we downgrade the authentication token for that user. Since we create the repositories as the organization administrator, a downgraded token will lead to permissions errors. As a workaround, all admins should log out of Classroom and log in again. Do not accept an assignment using an organization administrator account.

tarebyte:
This closes #445 where a user can reduce their token scopes if they sign in as teacher, and then sign in as a student.

Closes #604 #461 #46 #464 #497 #557 #560 #567 #581 #586 #588 #597 #604 

/cc @johndbritton @mkcode 

Apologies for the cassettes diff


app/controllers/application_controller.rb
```diff
@@ -28,7 +28,7 @@ def current_scopes
   end
 
   def required_scopes
-    %w(user:email repo delete_repo admin:org)
+    Classroom::Scopes::TEACHER
```
johndbritton: I like this.

app/models/user.rb
```diff
@@ -37,10 +39,38 @@ def github_client
...
+  def ensure_no_token_scope_loss
+    return true if token_was.blank?
+    return true unless token_changed?
+
+    old_scopes = GitHub::Token.scopes(token_was)
+    new_scopes = GitHub::Token.scopes(token)
+
+    # This is naive, if the token scopes ever change
+    # come back an revist this.
+    return true if old_scopes.size < new_scopes.size
```
johndbritton: don't love this, but it would only matter if we change the scopes we request for the different roles

tarebyte: @johndbritton we can do a smarter comparison here if you think that's better.

spec/lib/classroom/scopes_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe Classroom::Scopes do
```
johndbritton: Could we add a test in here that will fail if the ordering the different roles by `scopes.size` changes and invalidates the logic we're using to make sure we don't downgrade. That way that dependency is codified.

app/models/user.rb
johndbritton: Yeah, I think we should.

I think the logic should be as follows:

* If the new token has the same scopes, keep the new token.
* If the new token has different scopes, compare scopes
  * If scopes are only added, keep the new token
  * If scopes are only removed, keep the old token
  * If scopes are added and removed ???

The only odd case is if scopes are added and removed, not quite sure to do there. I think the only solution would be to store multiple tokens. In our case I think we should raise an error. It should never happen, but if we try and do it in the future we want to know that it's unimplemented.

tarebyte: I think there's a good amount of logic that would go into this, since this works, and we aren't changing out scopes, I'm going to punt on this for now and come back to it.

spec/lib/classroom/scopes_spec.rb
tarebyte: Done in 840fd59

johndbritton: Yeah, this is good enough to jog our memory if things need to be changed in the future.

PR #605, opened 2016-05-26
Add option to make students admins of their repos

rmacklin:
For some assignments, I would like my students to have admin permissions on their repositories. This PR adds a checkbox that allows you to enable this for any assignment:

screen shot 2016-05-26 at 4 35 13 pm


rmacklin:
@tarebyte I'd be happy to remove bc0a833fba068a1e6f5ff206ace3984dfaf8a4fb if you can generate the cassettes for an organization with private repositories.

rmacklin:
cc @johndbritton

johndbritton:
This looks reasonable to me, the only concern I have is "too many checkboxes" syndrome. What do you think about putting this and team limit settings into a group of advanced settings that are hidden by default on assignment pages.

We want to make the default experience really smooth and not have so many choices.

@tarebyte what do you think? If we do want this setting, I'd be ok with creating the advanced settings section in a follow up PR.

johndbritton:
@rmacklin A note on DB migrations:

Since this app is out in production, deploying this branch as is would cause errors until the migration is completed.

Usually we split the migration out into a separate PR, deploy, run, merge, and then deploy the feature that depends on the migration in a separate PR.

I'll leave it to @tarebyte to say if it's necessary for us to spilt the migration out, as it will only take a very short time to run. We should probably document the need to split migrations in `CONTRIBUTING.md` if we decide it's necessary.

tarebyte:
@rmacklin thanks for the PR! 

> Usually we split the migration out into a separate PR, deploy, run, merge, and then deploy the feature that depends on the migration in a separate PR.

> I'll leave it to @tarebyte to say if it's necessary for us to spilt the migration out, as it will only take a very short time to run. We should probably document the need to split migrations in CONTRIBUTING.md if we decide it's necessary.

We will definitely want to split the PR and run the migration first. I'll do that for you when this is ready to go and add the documentation for it as well.

> @tarebyte I'd be happy to remove [bc0a833](https://github.com/education/classroom/commit/bc0a833fba068a1e6f5ff206ace3984dfaf8a4fb) if you can generate the cassettes for an organization with private repositories.

I wouldn't worry about it, I've really stopped caring if it's private or public.

> If we do want this setting, I'd be ok with creating the advanced settings section in a follow up PR.

I like this setting, and definitely think we should follow up with a better user experience here. I'd also like to add these preferences into stafftools. That will be for another PR though.

--

@rmacklin I'll give this a read through, just to be thorough, looks great though!

rmacklin:
Re: splitting out the migration - makes sense, I can do that. Here is a new PR with only the migration: https://github.com/education/classroom/pull/608

rmacklin:
Re: "too many checkboxes" syndrome and improving the UX in a new PR

Sounds good to me. I appreciate you guys wanting to help get this deployed ASAP (so I can start using it), and I'm happy to work on the subsequent PR to improve the experience.

> I wouldn't worry about it, I've really stopped caring if it's private or public.

Ok, sounds good. (I do agree, though, it does clutter the feed quite a bit if you're running the tests a few times in a row)

app/models/concerns/github_repoable.rb
```diff
@@ -8,14 +8,18 @@ def add_team_to_github_repository
...
     github_user = GitHubUser.new(user.github_client, user.uid)
     repository  = GitHubRepository.new(organization.github_client, github_repo_id)
 
-    delete_github_repository_on_failure { repository.add_collaborator(github_user.login) }
+    options = {}
+    options[:permission] = 'admin' if give_admin_permission?
+    delete_github_repository_on_failure { repository.add_collaborator(github_user.login, options) }
   end
 
   # Public
```
tarebyte: I would rather `give_admin_permission?` be in this module, and in one method.

Maybe something similar to this:

```ruby
def give_admin_permission?
  student_assignment = self.has_attribute?(:assignment) ? assignment : group_assignment
  student_assignment.students_are_repo_admins?
end
```

app/models/concerns/github_repoable.rb
```diff
@@ -8,14 +8,18 @@ def add_team_to_github_repository
     github_repository = GitHubRepository.new(organization.github_client, github_repo_id)
     github_team       = GitHubTeam.new(organization.github_client, github_team_id)
 
-    github_team.add_team_repository(github_repository.full_name)
+    options = {}
+    options[:permission] = 'admin' if give_admin_permission?
```
tarebyte: WDYT about making this into a reusable method?

```ruby
def assignment_permissions
  {}.tap do |options|
    options[:permission] = 'admin' if give_admin_permission?
  end
end

app/models/concerns/github_repoable.rb
rmacklin: So originally I had started to write it where `GitHubRepoable` referenced `assignment` or `group_assignment`, but I noticed that no other methods in the file reference either of those. That made me think that it'd be better to keep it that way, keeping `GitHubRepoable` more generic and reusable. This means that if we ever added something other than an `Assignment` or `GroupAssignment` that needed to create a repository, we'd just have to implement `give_admin_permission?` on that thing without needing to change `GitHubRepoable` at all. Similarly, we could conceivably add a "make all my assignments default to making students be repo admins unless otherwise specified" setting to organizations. Would it make sense for `GitHubRepoable` to include that logic too, or should it only ask/care about "should I set the admin permission or not"? Of course, we may never make these kinds of changes, but it still seemed nice to decrease the coupling that way.

Anyway, just wanted to explain my line of thinking there. If you don't think those concerns matter, I can move `give_admin_permission?` into `GitHubRepoable`.

app/models/concerns/github_repoable.rb
rmacklin: 👍 I like it. Do you prefer `assignment_permissions` over `repository_permissions`?

tarebyte: Let's go with `repository_permissions`

rmacklin: Done in 85693763b228e1a40128a86d4d109a54d1c4c20b

app/models/concerns/github_repoable.rb
tarebyte: The logic for `GitHubRepoable` should only care about the creation of the Repository. Since the methods are basically identical, I think for right now it's fine to put it in `GitHubRepoable` if we add these other options we'll come back to it.

rmacklin: Done in 4b8a388df4647e442dbf687e9de5fa5da2f7b352

rmacklin:
Thanks!

PR #610, opened 2016-05-28
The aftermath

tarebyte:
This PR does a few things:

- [x] Uses the Rails cache for sessions, so that we can just clear the cache we ever need to do this again.
- [x] Enforces a 2 hour limit on being logged in
- [x] Creates a Rake task that finds all users that own an organization and verifies their scopes


tarebyte:
Need to split this up.

PR #614, opened 2016-05-31
Student Identifier

### LINKED ISSUE #448: Implement "Student Identifiers" to map between GitHub users and Students

dinever:
Hi guys,

I'm working on the student identifier feature. This PR should close #448 at the end of the day.

For this feature I created two tables, one is for _student identifier type_, another one is for _student identifier_.

Each _student identifier_ relies on a _student identifier type_, which contains the name and description for the student identifier.

Teachers will be able to create _student identifier type_ in the classroom, and indicate an id type for an assignment.

Creating a **student identifier type**:

demo-mobile

Choosing a **student identifier type** for an assignment:

demo-mobile

When students visit the assignment invitation page, they will be redirected to another page and will be asked to fill in their identifiers if they don't have the required type of identifier recorded:

demo-mobile

Students will not be asked to fill in the identifier again when accepting other assignment invitations using the same identifier type.

**Roadmap:**
- [x] Student identifier type selection in assignment options
- [x] Index for Student identifier types
- [x] Adding new student identifier types
- [x] Deleting old student identifier types
- [x] Recording student identifier when accepting an assignment
- [x] Recording student identifier when accepting a group assignment
- [x] Displaying student identifier with student github login on the assignment page
- [x] Style polishing
- [x] Test cases

@johndbritton @tarebyte Is this too complicated? Would ❤️  to hear your voice. 😸


db/migrate/20160524234655_create_student_identifiers.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.string :description
+      t.integer :content_type
+    end
+
+    create_table :student_identifiers do |t|
+      t.belongs_to :organization, index: true
+      t.belongs_to :user,         index: true
+      t.belongs_to :student_identifier_type
+      t.string :value
+    end
```
tarebyte: Should these column values ever be `NULL` if that's the case we should enforce this on the database as well as in the Rails model.

tarebyte:
@mastahyeti I'd like your take on this from a security standpoint.

We're adding student identifiers to Classroom, the identifier may be sensitive information that I'm not super comfortable with storing in plain text here. But it needs to be reversible so that the teacher can see the identifier.

What would be the best approach here?

db/migrate/20160524234655_create_student_identifiers.rb
dinever: @tarebyte They should not be `NULL`. I'll add NOT NULL constraint, thanks for the feedback!

mastahyeti:
I don't think this app should be storing anything sensitive. Rather than allowing arbitrary student-identifier-types to be defined by the instructor is might be better to provide a few common identifier-types like student-ID. It would be pretty bad if an instructor started asking for social security number, for example.

tarebyte:
@mastahyeti even if they are student-ids I think that's still relatively sensitive no?

mastahyeti:
I don't think student-ids would need to be encrypted. @darcyck might have some thoughts on this discussion though.

DarcyCK:
@mastahyeti @tarebyte @dinever  - looping in hannah from legal as the privacy SME - I have some very limited experience with student data, and would need to research, so hoping hannah already has some solid perspective on the above to expedite getting you and answer.

@nsqe - hannah - asking you to take a look at this one from a US student data privacy law/protection perspective as the privacy subject matter expert. what are your thoughts on guidance (see above)? Would we need to consider the restrictions under FERPA or another regulatory ruling?

dinever:
@tarebyte 

To make sure an instructor does not ask for some sensitive information, maybe we should just provide him with several hard-coded types of student identifiers (e.g. `student ID`, `academic email address` and `full name`) instead of giving them the choice to define their own.

tarebyte:
> To make sure an instructor does not ask for some sensitive information, maybe we should just provide him with several hard-coded types of student identifiers (e.g. `student ID`, `academic email address` and `full name`) instead of giving them the choice to define their own.

@dinever I think that's fine, we also need to specify in the UI that teaches must not ask for private or sensitive information as the ID, ie SSN passport number etc.

app/controllers/assignment_invitations_controller.rb
```diff
@@ -5,6 +5,8 @@ class AssignmentInvitationsController < ApplicationController
   before_action :check_user_not_previous_acceptee, only: [:show]
 
   def accept_invitation
+    create_student_identifier
```
tarebyte: This should be apart of `invitation.redeem_for`

mastahyeti:
It would be good to also tell students that they should never be entering sensitive information into the app.

app/assets/javascripts/group-assignment-invitations.coffee
```diff
@@ -10,6 +10,11 @@ $('.group_assignment_invitations.show').ready ->
       $submitButton.prop('disabled', true)
   )
 
+  $('form').submit( ->
+    hiddenField = $(this).find('#student_identifier_hidden')
+    hiddenField.val($('#student_identifier').val())
+  )
```
dinever: Explanation on this:

In the group assignment invitation page, we have multiple forms, and students will choose either one of them. See the image below:

screen shot 2016-06-06 at 9 24 09 am

We will need to ask students to fill in their identifier when accepting the assignment. Since we have multiple forms on the same page, they have to share the same input field.

So I created a hidden input field for each form, and one stand-alone visible input field to let the student input the identifier. When the form is submitted, copy the value from the visible identifier input field to the hidden identifier input field, then the identifier value will be included in the form data.

I believe we can come up with a better solution for this. It would be great if you can give me some feedback on this. Thanks! 😸 

/cc @tarebyte @jasonlong

dinever:
Just got this updated, added some GIF images to show how it works. 

@tarebyte It'd be great if you can give me some feedback on this when you're available. Thanks!

jasonlong:
It would probably be better all around (UX-wise and implementation-wise) if we made this a 2 step thing. 

Step 1: enter your ID (which can then be checked before moving on)
Step 2: select your group

Thoughts?

tarebyte:
@jasonlong that sounds like a good idea to me

dinever:
@jasonlong Sounds good. Thanks!

dinever:
All changes mentioned above have been done. Thanks!

Now I need to do some style polishing and figure out a 🆒 way to display student identifiers on the assignment page.

/cc @tarebyte

app/models/organization.rb
```diff
@@ -11,6 +11,7 @@ class Organization < ActiveRecord::Base
   has_many :groupings,         dependent: :destroy
   has_many :group_assignments, dependent: :destroy
   has_many :repo_accesses,     dependent: :destroy
+  has_many :student_identifier_types, dependent: :destroy
```
tarebyte: Would you mind lining all of these up?

```ruby
has_many :repo_acesses,             dependent: :destroy
has_many :student_identifier_types, dependent: :destroy
```

app/models/student_identifier_type.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+class StudentIdentifierType < ActiveRecord::Base
+  default_scope { where(deleted_at: nil) }
+
+  belongs_to :organization, -> { unscope(where: :deleted_at) }
+
+  enum content_type: [:text, :number, :email]
```
tarebyte: Maybe we need to make these more explicit?

```ruby
enum content_type: [:name, :id, :email]
```

dinever: Hi @tarebyte,

Actually `content_type` here is made to specify the input form type on the assignment invitation page:

screen shot 2016-06-08 at 10 43 12 pm

For example, if the teacher set `content_type` to `number`, the identifier form appeared on the assignment invitation page would have type `number`, so that students can not input anything else other than decimal numbers:

screen shot 2016-06-08 at 10 38 11 pm

If we change this to `[:name, :id, :email]`, then it may be a little bit confusing with the `name` field. WDYT? 😄

jasonlong: This is minor, but that "Please don't ask..." notice isn't in a great spot below the save button. That seems like an important thing people should see, so maybe have that at the top of the form?

dinever: Hi @jasonlong, thanks for the feedback!

What do we think about using a `flash` to alert the teacher?

screen shot 2016-06-10 at 4 40 32 pm

Also, to implement the identifier feature we will need to display student identifiers on the group assignment page. So I'm proposing to modify the group assignment page so that we have some space left to show other information about the student including student identifiers:

**Before:**

screen shot 2016-06-10 at 4 44 00 pm

**After:**

screen shot 2016-06-10 at 4 39 45 pm

In the modified version we can see that student identifiers were placed on the right of student login.

Haven't updated the code but it would be great if you can give me some feedback on this. Thanks!

nsqe: That looks great to me!

app/models/github_user.rb
```diff
@@ -15,6 +15,10 @@ def github_avatar_url(size = 40)
     "#{avatar_url}&size=#{size}"
   end
 
+  def name(options = {})
+    GitHub::Errors.with_error_handling { @client.user(@id, options).name }
+  end
```
tarebyte: This is redundant method. I made a change in https://github.com/education/classroom/pull/622 which allows us to create method based on the `attributes` method down below.

app/models/student_identifier_type.rb
tarebyte: @dinever I like the flash message, and the grouping. Nice work!

dinever:
**Update:**

1. Modified the group assignment team style so that we have space to show additional information:

    Before:

    screen shot 2016-06-12 at 10 39 46 pm

    After:

    screen shot 2016-06-12 at 10 38 08 pm

2. Modified the assignment page style so that we can display the student identifiers:

    Before:
   
    screen shot 2016-06-12 at 10 57 46 pm

    After:

    screen shot 2016-06-12 at 10 55 36 pm

/cc @tarebyte @jasonlong

dinever:
@tarebyte 

Just added the tests, I believe this one is ready for review. Thanks!

spec/controllers/student_identifier_types_controller_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe StudentIdentifierTypesController, type: :controller do
+  include ActiveJob::TestHelper
+
+  let(:organization)  { GitHubFactory.create_owner_classroom_org }
+  let(:user)          { organization.users.first                 }
+  let(:student)       { GitHubFactory.create_classroom_student   }
+  let(:student_identifier_type) do
```
tarebyte: Since this is pretty generic, what do you think about moving this into a factory?

app/assets/javascripts/identifier-form.coffee
```diff
@@ -0,0 +1,22 @@
...
+  $form = $('form')
+
+  $('form').on('change keyup', ->
+    $submitButton = $('.js-form-submit')
+
+    if $form_values_present()
+      $submitButton.prop('disabled', false)
+    else
+      $submitButton.prop('disabled', true)
+  )
```
tarebyte: I think we can cut this down quite a bit, what do you think about this?

```coffee
$('.assignment_invitations.identifier, .group_assignment_invitations.identifier').ready ->
  $('form').on('change keyup', -> 
    $('.js-form-submit').prop('disabled', $('student_identifier).length)
  )
)
```

app/controllers/assignments_controller.rb
```diff
@@ -62,6 +72,13 @@ def set_assignment
     @assignment = @organization.assignments.includes(:assignment_invitation).find_by!(slug: params[:id])
   end
 
+  def student_identifier_type_param
+    unless params.key?(:student_identifier_type) && params.key?(:student_identifier_type)
```
tarebyte: This can just be

```ruby
return unless params.key?(:student_identifier_type) && params.key?(:student_identifier_type)
StudentIdentifierType.find_by(id: student_identifier_params[:id], organization: @organization)
```

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -26,8 +40,8 @@ def accept_assignment
   end
 
   def accept_invitation
-    selected_group       = Group.find_by(id: group_params[:id])
     selected_group_title = group_params[:title]
+    selected_group = Group.find_by(id: group_params[:id])
```
tarebyte: Any reason why this change was needed? Seems aesthetic.

app/decorators/assignment_repo_decorator.rb
```diff
@@ -0,0 +1,44 @@
...
+    student.login
+  end
+
+  def student_name
+    student.name
+  end
+
+  def student_identifier
+    assignment_repo.user.identifier(assignment_repo.assignment.student_identifier_type)
+  end
```
tarebyte: I finished removing decorators from Classroom, mind moving this to the model?

app/models/user.rb
```diff
@@ -54,6 +54,10 @@ def staff?
     site_admin
   end
 
+  def identifier(identifier_type)
```
tarebyte: `identifier_type` seems redundant

```ruby
def identifier(type)
  StudentIdentifier.find_by(user: self, student_identifier_type: type)
end
```

tarebyte:
screen shot 2016-06-14 at 23 26 01

It would be great if instead of an empty select this linked you to creating your first student identifier

tarebyte:
screen shot 2016-06-14 at 23 27 41

Would a teacher ever want to edit these?

tarebyte:
screen shot 2016-06-14 at 23 29 01

I feel like this text could be styled better, thoughts?

tarebyte:
screen shot 2016-06-14 at 23 35 46

This should list the ones already created, and give the teacher the ability to create a new one.

app/views/assignment_invitations/identifier.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      <span class="assignment-icon assignment-icon-individual">
+        <%= octicon('person') %>
+      </span>
+      Accept the <strong><%= assignment.title %></strong> assignment
+    </h2>
+  </div>
+</div>
+
+<div class="site-content-body">
+  <%= form_tag({ controller: :assignment_invitations, action: :submit_identifier}, method: :patch) do %>
```
tarebyte: Mine helping my OCD out and fixing this line?

``` diff
- <%= form_tag({ controller: :assignment_invitations, action: :submit_identifier}, method: :patch) do %>
+ <%= form_tag({ controller: :assignment_invitations, action: :submit_identifier }, method: :patch) do %>
```

app/views/assignment_invitations/identifier.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      <%= render partial: 'shared/identifier_form', locals: {student_identifier_type: assignment.student_identifier_type} %>
```
tarebyte: Same here:

```diff
- <%= render partial: 'shared/identifier_form', locals: {student_identifier_type: assignment.student_identifier_type} %>
+ <%= render partial: 'shared/identifier_form', locals: { student_identifier_type: assignment.student_identifier_type } %>
```

app/views/group_assignment_repos/_group_assignment_repo.html.erb
```diff
@@ -3,31 +3,39 @@
...
+          </span>
+          <p class="student-description" itemprop="name">
+            <%= student.github_user.name %>
+          </p>
+        </div>
+      </div>
+  <% end %>
+
+</div>
+
```
tarebyte: ✂️ the extra whitespace please 😄

app/views/student_identifier_types/index.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <div class="blankslate large-format spacious clean-background">
+        <h3>This organization doesn’t have any student identifier types yet.</h3>
+        <%= link_to new_organization_student_identifier_type_path(@organization), class: 'btn btn-green btn-big', role: 'button' do %>
+            Create your first student identifier type
+        <% end %>
+      </div>
+    <% end %>
+  </div>
+</div>
+
```
tarebyte: ✂️ the extra whitespace please 😄

app/controllers/assignments_controller.rb
```diff
@@ -49,6 +52,13 @@ def destroy
 
   private
 
+  def student_identifier_types
+    @student_identifier_types ||= @organization.student_identifier_types.map do |student_identifier|
```
tarebyte: WDYT about using `select` here? So that we only load what we need.

```ruby
@student_identifier_types ||= @organization.student_identifier_types.select(:name, :id).map do |student_identifier|
  [student_identifier.name, student_identifier.id]
end
```

tarebyte:
@dinever I want to make another pass at this tomorrow morning, there's some coupling that I think we can make a lot cleaner.

Let me know what you think about my comments above.

Thanks!

app/assets/javascripts/identifier-form.coffee
```diff
@@ -0,0 +1,5 @@
+('.assignment_invitations.identifier, .group_assignment_invitations.identifier').ready ->
+  $('form').on('change keyup', ->
+    console.log($('student_identifier').length)
```
tarebyte: Probably should ✂️ this line

dinever: Done in f5a051c, didn't notice that. :fearful:

spec/controllers/student_identifier_types_controller_spec.rb
dinever: 👍  Done in 5e8644c876ba687c70ad12068d2fa168d6c02a16.

app/controllers/assignments_controller.rb
dinever: Done in fa12824404b328bfc6c7916d173a014341be59ea

dinever:
> It would be great if instead of an empty select this linked you to creating your first student identifier

@tarebyte 

I've added a button in 4731ceec055319f738a59ab4370a84b1ce7e2e43 to lead the teacher to the new identifier page if there is no existing one:

demo

However, as we can see from the GIF, when the teacher clicked `Save changes`, he was lead to the identifiers page, which is not very convenient. The intended behavior should be that the teacher being led back to the assignment editing page so that he can select an identifier for the assignment and save the changes. Is there a way that we can lead the teacher back to the previous page wherever he's from after saving the identifier?

dinever:
Also, the `button` in the identifier type creating page should be labeled `Create Identifier`, not `Save changes`.

I'm adding an editing page for each identifier type, then fix this ⬆️  problem.

tarebyte:
> Is there a way that we can lead the teacher back to the previous page wherever he's from after saving the identifier?

@dinever we can save where they were at as part of the session and redirect them back to it.

something like

```ruby
session[:pre_identifier_location] = our_url
```

dinever:
> screen shot 2016-06-14 at 23 35 46

> This should list the ones already created, and give the teacher the ability to create a new one.

@tarebyte This is what I was thinking about too. However, the settings page may get a little bit long if we do so. What do we think about using a sidenav to split the settings page into multiple parts? ⬇️ 

screen shot 2016-06-15 at 10 44 55 pm

/cc @jasonlong

tarebyte:
@dinever I like that a lot. I hate to be a pain, would you mind making this settings change it's own PR. This diff is getting large enough as it is.

tarebyte:
The side nav in the initial PR can just be the change you have listed, and have two links. The selected Classroom profile and the other being Invite others.

dinever:
> screen shot 2016-06-14 at 23 27 41

> Would a teacher ever want to edit these?

Hi @tarebyte,

I've added some new endpoints in b8b2488cb2e49dbaf8b23bfbb5a778b7c27ddedb for teachers to edit the identifier types.

**Demo:**

demo


**Showcase:**

screen shot 2016-06-17 at 9 01 59 am

screen shot 2016-06-17 at 9 02 30 am

screen shot 2016-06-17 at 9 02 19 am

tarebyte:
@dinever this is getting really large really fast.

Can we start breaking this into smaller PRs to break this down for easier reviewing.

I think the first one should be just the identifier models and the migrations, and we can build up from there.

:cool:?

johndbritton:
@dinever @tarebyte I'm pretty sure this can be closed as it was split up, can you confirm and close if so?

Issue #615, opened 2016-05-31
Creating teams - interface


drbeco:
I don't quite understand what should I write here:

![image](https://cloud.githubusercontent.com/assets/5805111/15687387/c9d3f0e0-274a-11e6-94b6-497bbd28cb9c.png)

I've already tried to create teams beforehand, and that proved a waste of time.

Now I'm starting a new assignment. What should I do?

Thank you;

tarebyte:
@drbeco so the end goal here is that you are creating a name for your group of teams.

For example `Teams for Project 1`.

What would make this more clear?

drbeco:
Hello @tarebyte ,

Thank you for explaining. Now that I did it the first time, it is a bit more clear.

But I'm taking naively your last question as a true question, not just rhetoric; and as such I'm taking the liberty to answer to the best of my knowledge, ok?

I believe it would be much more clear at once if the form had standardized both questions using the same vocabulary (set/group, choose/reuse, select/create). Let me explain in a low level why I think that is better.

I am very experienced in teaching programming to foreign (non-english native) students, and one thing the non-native brain thinks is that all english words can be used as a _noun_ to label a characteristic. That means you can read things like "eu criei um conjunto 'set' e outro 'group'" and that makes sense. It will translate roughly as "I've created a set 'set' and other 'group'", but this translation doesn't grasp what it really means. A more explicit translation would be "I've created a set nominated _set_ and another set nominated _group_", and because it is in another language you don't mix the meanings. This is one hand; on the other hand, reading technical texts may be confusing if the author, a native english speaker, start to use synonymous for style. If you change one word to another, in a romance, it sounds better, avoid repetition; but in a technical text, when you do, you are just implying that what you are saying is not related, but another characteristic. So, after explaining a bit of the reasoning, let me give some solutions that will fit this brain-theory of mine and see if we can get things a bit more clear.

For example:

Option 1:

![github3-groups](https://cloud.githubusercontent.com/assets/5805111/15694644/ed14cc5c-2775-11e6-8c73-4d0a212a1168.png)


Option 2:

![image](https://cloud.githubusercontent.com/assets/5805111/15693505/05e1710e-276c-11e6-9ca0-ad0bd0171b2a.png)


Not only the words repeat itself, but also the gray area gives an example. 

The current example is based on dates. My first group/set was created just by chance with a strange name that mislead me to think that I need to know a range of dates to setup the team.

`Teams formed on may 24th, 2016:`

[...]

PR #617, opened 2016-06-03
Changing labels from `group of teams` to `set of teams`

### LINKED ISSUE #601: Workflow - What if I delete this team?
Hello guys,

It is my first time using git classrom, and I'm not certain how the workflow goes.

My steps until now was:

Before EVERYTHING:

1. I've created a "startup repository" to give as assignment start point code (exwumpus)
2. I've created 6 teams and invited the students in each, accordingly to my own division of the class.

Now, with all set, I went to create a new ASSIGNMENT:
1. Set assignment title (pretty straight)
2. Private (easy)
3. Start code (done)
4. Maximum members per team: 7 (*)
5. Existing group of teams by date. (**)

Then I WAITED... :)
After a day or two, the first member created the assignment. For my surprise, I didn't know, but classroom created another team with only one member.

Lets talk with pictures.

Teams created by me are the ones in the image bellow that I can "leave". And the only one I can "Join" was created by classroom:

image

Well, I don't want it. Can I delete it?

If I go to classrooms, I can see the new repo there:

image

So I went into the repo configuration and added the correct team:

image

I also went to **MY** the team configuration to check:

image

And the **CLASSROOM** team doesn't have any members:

image

----

Bottom line: it is duplicating the teams I already set and put members in.

Can I delete the **classroom** team, or this will break/bug my assignment?

Also, what am I doing wrong with the workflow? Why I have this effort of creating the teams and setting them not recognized by the application?

Thank you!

(*) 4. Maximum members per team: 7

I told them to create groups of 5, but let a range of +-2 to be sure there is room for everyone. So I would like to suggest to ask not only the maximum, but also the "minimum", or a range. 

(**) 5. Existing group of teams by date.

Here, just by date may be trick. Maybe if we could list in a comma-separated list all the teams it would be more precise. Maybe this is the problem I'm having here.

Thanks all for any input clarifying the workflow (and suggesting new ways of doing it, also welcomed!)

dinever:
Change the text label to avoid misunderstanding

Closes #601 #615


tarebyte:
@dinever thanks for updating these.

From a comment in https://github.com/education/classroom/issues/615#issuecomment-222860889 it might also be helpful to give a better name for the set of teams.

What are your thoughts on this?

drbeco:
Do you mean the gray area?

dinever:
Hi @drbeco 

This is a preview of the change:

screen shot 2016-06-04 at 8 49 33 pm

As you can see, the first label is modified from `group of teams` to `set of teams`.

I wonder if you have any suggestions or ideas on a better name for `set of teams`.

Thanks for the feedback in advance! 🙇

drbeco:
Yes, please. The word reusable is misleading.

You may say it in both or in none.

Just an option:

* Reuse an existing set of teams
* Or create a reusable set of teams

Or

* Choose an existing set of teams
* Or create a new set of teams

Pair: Reuse/reusable or existing/new. Not existing/reusable, that doesn't fit properly in poor vocabulary. It may sound a bit pedantic, I know, but trust me it will make it clear.

tarebyte:
> Choose an existing set of teams
> Or create a new set of teams

@dinever let's go with this option

dinever:
@tarebyte 

Thanks! This is done.

PR #618, opened 2016-06-03
Prevent Javascript event callbacks from firing multiple times

### LINKED ISSUE #607: Javascript events fired multiple times
Just noticed that some Javascript events fired multiple times when the page first loads.
This might cause some problem like sending multiple requests to the server.

As discussion with @tarebyte, this might be a problem with TurboLinks.

One possible solution is setting `data-turbolinks-eval` to `false` as stated in [TurboLinks Classic - Evaluating script tags](https://github.com/turbolinks/turbolinks-classic#evaluating-script-tags) to prevent re-evaluation.

cyhsutw:
Summary of changes:
- set `data-turbolinks-eval` to `false` in `javascript_include_tag 'application'` of `application.html.erb`
- remove `page:load` binding in `navigation.coffee`

Closes #607

### DIFF
--- a/app/assets/javascripts/navigation.coffee
+++ b/app/assets/javascripts/navigation.coffee
@@ -13,4 +13,3 @@ ready = ->
     $('.loading-indicator').show())
 
 $(document).ready(ready)
-$(document).on('page:load', ready)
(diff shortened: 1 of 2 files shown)


tarebyte:
Great work, thanks @cyhsutw!

PR #571, opened 2016-04-16
Keep track of user activity

cyhsutw:
Hello guys,

This PR is related to #356, it tracks the timestamp of a user's last access to the app.

Staffs can view the timestamp using `stafftools`, as shown below:

screen shot 2016-04-16 at 7 55 19 pm

PR #619, opened 2016-06-04
[WIP] Keep track of user activity

cyhsutw:
Hello guys,

This PR is related to #356, it tracks the timestamp of a user's last access to the app.

Staffs can view the timestamp using stafftools, as shown below:

last_active_at

*The difference between #571 and this pull request is that whether the `updated_at` of `User` will be updated or not.*


app/views/stafftools/users/_user.html.erb
```diff
@@ -5,5 +5,8 @@
     <strong class="search-result-title"><%= user.login %></strong>
   <% end %>
 
-  <span class="text-muted css-truncate css-truncate-target">Joined <%= local_time_ago(user.created_at) %></span><br>
+  <span class="text-muted css-truncate css-truncate-target">
+    Joined <%= local_time_ago(user.created_at) %> /
+    Last active: <%= user.last_active_at.present? ? local_time_ago(user.last_active_at) : 'Unknown' %>
```
tarebyte: Why don't we set the `last_active_at` to `created_at` when the user is first created? That way we never have the case where the `last_active_at` is unknown.

db/migrate/20160416105605_add_last_active_at_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastActiveAtToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :last_active_at, :timestamp
```
tarebyte: I don't think this should ever be null? At worst case they were last active when the signed in for the first time.

tarebyte:
@cyhsutw thanks for doing this.

What do you think about writing a data transition script so that we can set these timestamps for everyone?

The process I see here is:
- Make a PR with just the migration, with `last_active_at` allowed to be null
- Run a data transition script that uses the `created_at` or `updated_at` timestamp whatever you prefer.
- Make a PR with another migration that enforces `last_active_at` to be present
- Merge this PR with the appropriate changes.

Does that sound alright?

db/migrate/20160416105605_add_last_active_at_to_users.rb
cyhsutw: You're right, this should not be null. Thanks for pointing this out.

cyhsutw:
@tarebyte thanks for the feedback.

I'm totally 🆒  with the process you suggested.

Just a little question about:

> Run a data transition script that uses the created_at or updated_at timestamp whatever you prefer.

If we do this in a migration like:

```ruby
class SetUserLastActiveAtValue < ActiveRecord::Migration
  def change
    User.update_all('last_active_at = updated_at')
  end
end
```

The `elasticsearch` indexes will not be updated, thus we might have `null` values in `Stafftools`.
Is there any way to rebuild the indexes, or we should create a rake to scan all of the users and update index for each user instead?

app/views/stafftools/users/_user.html.erb
cyhsutw: Yes, this should not be null in our main Postgres database.

I was wondering if there is a case that the indexes on Elasticsearch is corrupted/outdated, which might cause this value to be null.

cyhsutw:
@tarebyte I added some migration files, if there's no problem with those,  I'll go ahead and open separate PRs for each of them.

Also, I am working on adding few more test for this change, so I'm marking this as WIP for now.

cyhsutw:
I added a test to the `User` model. 

Should we add another test to check if the `StafftoolsIndex::User` is updated?

cc/ @tarebyte

tarebyte:
> Should we add another test to check if the `StafftoolsIndex::User` is updated?

I don't think that's necessary

app/controllers/application_controller.rb
```diff
@@ -36,7 +36,11 @@ def adequate_scopes?
   end
 
   def authenticate_user!
-    auth_redirect unless logged_in? && adequate_scopes?
+    if logged_in? && adequate_scopes?
+      become_active
+    else
+      auth_redirect
+    end
```
tarebyte: @cyhsutw What do you think about something like this?

```ruby
return become_active if logged_in? && adequate_scopes?
auth_redirect
```

app/models/user.rb
```diff
@@ -52,6 +54,13 @@ def staff?
     site_admin
   end
 
+  def become_active
+    # we don't want to touch updated_at
+    update_columns(last_active_at: Time.zone.now)
+    # but we still want indexes to work
+    self.class.update_index('stafftools#user') { self }
+  end
```
tarebyte: I think these comments are useful, but would be better placed at the top.

```ruby
# This updates the `last_active_at` column without
# updating the model, but keeps the index updated.
def become_active
  update_columns(last_active_at: Time.zone.now)
  self.class.update_index('stafftools#user') { self }
end
```

app/controllers/application_controller.rb
cyhsutw: 🆒  That's much more elegant!

spec/models/user_spec.rb
```diff
@@ -12,6 +12,12 @@
     end
   end
 
+  describe '#become_active' do
+    it 'updates the last_active_at attribute' do
+      expect { user.become_active }.to change { user.last_active_at }
```
tarebyte: We should also check here that the `updated_at` column remains the same before and after.

```ruby
it 'does not change the updated_at column' do
  updated_at = user.updated_at
  user.become_active
  expect(updated_at).to eql(user.updated_at)
end
```

tarebyte:
I left a few minor comments, @cyhsutw. After that this is good to go!

spec/models/user_spec.rb
cyhsutw: Done in #b3ed41b.

cyhsutw:
@tarebyte Thanks for reviewing the changes, just updated the code based on your comments.

tarebyte:
Confirmed in production!

Issue #572, opened 2016-04-16
Student projects roadmap summer 2016


johndbritton:
We're expecting to have a few students focused on improving Classroom this summer with help from @tarebyte and me. Let's use this issue to plan goals for the summer and track progress against them. For a first draft, I've split tasks into three main categories: setting up, monitoring progress, and winding down. By splitting our goals across the different parts of the usage cycle my hope is that we can work on the various projects in parallel.

**Setting up (@dinever)**
- [ ] Student identifiers #614 
- [ ] Team management
- [ ] Pre-defined teams
- [ ] Improved invitation experience

**Montoring progress (@cyhsutw)**
- [ ] Repository creation
- [ ] Due dates
- [ ] Explicit submission
- [ ] Visualize classroom activity

**Winding down (@nicktikhonov)**
- [ ] Download all student submissions
- [ ] Access "submitted" commits
- [ ] Archive submissions
- [ ] Track feedback to students
- [ ] Transfer code to student ownership

I'm open to other ways of organizing this work, so please comment here if you have ideas. Also, if there's something you don't see listed here that you think we need to address, please share. I can't promise that we'll be able to implement every feature requested but my hope is to get through the most needed features in time for the fall semester.

cyhsutw:
Hi @johndbritton @tarebyte 

Thank you so much for considering my proposal and giving me the opportunity to work on GitHub Classroom this summer.

I have some ideas/questions about the items listed above, is there any preferred way for the discussion?

tarebyte:
I would raise another issue for a discussion thread and reference this issue.

PR #626, opened 2016-06-08
[WIP] Explicit Assignment Submission

cyhsutw:
Hello guys,

I'm working on the explicit submission feature and it's a one of the subtasks in #572.

The procedure of submission works as follows:

1. Classroom sets a organization webhook and receiving `push` events.
2. Upon receiving a `push` event of the **default branch**, Classroom will create a `release` for the `HEAD` commit.

Students can see their submissions in the release page of their repos:

releases

### TODO
- [ ] Delete all previous releases come with the starter code repo
- [ ] Ensure webhooks are correctly set: schedule background jobs to trigger `ping` events for hooks that have not been pinged
- [ ] Put this feature under Flipper
- [ ] More descriptive messages for the `release`s created by Classroom
- [ ] Add tests for this feature

### Questions / Discussions

- Since we're using `release` feature to create submissions, should we delete all the releases comes with the starter code repo? Yes.

- What do you think about exposing more information to the students so they can check if they have submitted their work properly?

    demo


tarebyte:
> * Since we're using `release` feature to create submissions, should we delete all the releases comes with the starter code repo?

Yes we should, or we need to make a way to identify the releases we create.

cyhsutw:
> Yes we should, or we need to make a way to identify the releases we create.

I'm 👍  with this.

Although currently we have a special prefix for submission releases: `classroom/submission/*` that can help us identify the submissions programmatically, deleting previous releases would make things more clear to the students (no submission at the point when the repo is created).

lib/classroom/scopes.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 module Classroom
   module Scopes
-    TEACHER                  = %w(user:email repo delete_repo admin:org).freeze
-    GROUP_ASSIGNMENT_STUDENT = %w(admin:org user:email).freeze
-    ASSIGNMENT_STUDENT       = %w(user:email).freeze
+    TEACHER                  = %w(user:email repo delete_repo admin:org admin:org_hook).freeze
```
cyhsutw: @tarebyte How can we make this Flippable? I think we don't have information about `current_user` here.

tarebyte: @cyhsutw I would just make a new constant, and then when we ask for the scope in the controller check for the `current_user` then.

app/controllers/organizations_controller.rb
```diff
@@ -19,6 +19,9 @@ def new
 
   def create
     @organization = Organization.new(new_organization_params)
+    if Classroom.flipper[:explicit_assignment_submission].enabled? current_user
+      @organization.setup_webhook(webhook_events_url)
```
cyhsutw: @tarebyte I was doing this in the `before_validation(on: :create)` callback of the `Organization` model. However, as we must have the `webhook_events_url` in order to setup the webhook, including `Rails.application.routes.url_helpers` in `Organization` and setting the `host` of the application in `environment/*.rb` are necessary.

Is it fine to place this piece of code here or we have better options?

tarebyte: @cyhsutw We should probably make it a service object, I'd really rather not put the logic for this in a controller.

cyhsutw: Thanks for the suggestion, I'll try to do that.

cyhsutw:
Close for a better PR.

PR #631, opened 2016-06-11
Remove Organization Decorator

tarebyte:
This is continuing the work of #622, for organizations.

This PR gives the `Organization` model a `GitHubOrganization` instead of delegating it to a decorator.


tarebyte:
Confirmed in production

PR #633, opened 2016-06-12
[Migration] Ensure that last_active_at can't be null

tarebyte:
Part of #619 

/cc @cyhsutw


cyhsutw:
Looks good to me! 👍

tarebyte:
Migration in production successful

PR #643, opened 2016-06-13
Kill GroupAssignmentDecorator and remove draper

tarebyte:
This kills the last decorator, and removes the draper gem... Never again.


rmacklin:
Just wanted to say nice job! You really ran through that work

PR #651, opened 2016-06-19
Fixed a bug in group assignment invitations controller

dinever:
We used the helper method `group` in [app/views/group_assignment_invitations/accept.html.erb#L17](https://github.com/education/classroom/blob/0e039e4cd3f3a9eef40ee699982ce95bff140f8e/app/views/group_assignment_invitations/accept.html.erb#L17), but don't have `helper_method :group` declaration at [app/controllers/group_assignment_invitations_controller.rb#L75](https://github.com/education/classroom/blob/0e039e4cd3f3a9eef40ee699982ce95bff140f8e/app/controllers/group_assignment_invitations_controller.rb#L75), which causes the view unable to be rendered.

I've fixed the bug and added some tests in fb05d0b47c68ee228453e43cd932c330a2007b15 and have fixed it in 64d28e719cad4c211e6be579657fdea89f05a73c. The failed build showing the error message can be found at [travis-ci.org/education/classroom/builds/138661782#L1303](https://travis-ci.org/education/classroom/builds/138661782#L1303).

Thanks!


tarebyte:
Nice catch, thanks @dinever!

Issue #292, opened 2015-10-01
Team management for administrators


johndbritton:
When accepting a `GroupAssignmentInvitation` users have the ability to create or join a team, but in some cases they make a mistake as in #291.

One potential solution to this problem is to provide an interface for teachers to manage teams. This way they can reorganize the users if there's a mistake and we can keep the organization administrators in control of team changes.

cboettig:
I believe this issue is related to #730, and thus might be (at least in part) addressed by pending PR #980.  Linking them here since neither is mentioned above, though I'd appreciate clarification on this issue.


I agree that the inability to manage or change teams is a significant problem. Summarizing the above, a user cannot un-join a team; deleting the team using the GitHub 'Teams' interface just leaves the user a member of the team `(Deleted Team)`, creating `ghosts` in assignments.  This prevents changing teams between assignments or undoing accidentally joined teams.

One work-around I have found is to create a new "Set of Teams" when creating the assignment.  This allows a user who had previously joined a team in some other "Set of Teams" to join a new new team.   Unfortunately, there is no interface to list or manage these "Sets", and they appear to have no correspondence to the "Teams" menu in the GitHub organization (teams appear in that menu but with no mention or association to the "Set of Teams".). 

 Ideally these sets would show up as "meta-teams" (teams of teams) in the usual Teams menu of the organization, and deleting them there would simply erase the effect of creating a team.  


A related but independent issue is that the current interface allows any student to join any other student's team.  Maybe this is fine and nicely streamlines the process by sidestepping the need for the instructor to laboriously create all the student teams in advance.  However, it also seems brittle, in that invariably someone will join the wrong team or want to change teams (an issue which is exacerbated by the inability to 'undo' this by 'leaving' or even deleting the team).  The first student to create the 'team' doesn't get any control over which other student joins: perhaps a better workflow would be for the student who creates the team to get some ability to invite their teammates directly, rather than letting anyone opt in to their team?  

Thanks for all the work on GitHub Classroom, this is a truly awesome and compelling workflow!

leouieda:
> One work-around I have found is to create a new "Set of Teams" when creating the assignment. This allows a user who had previously joined a team in some other "Set of Teams" to join a new new team. 

After having a lot of trouble with students dropping out of the class or having problems with their teammates, I defaulted to creating new teams for every assignment. Most problems of that kind went away after that. It's not very efficient but for large classes it seems easier than micromanaging teams.

[...]

PR #652, opened 2016-06-19
List groupings and groups in a classroom

dinever:
This is part of the team management interface #292. To make sure a PR doesn't get too large to view, I'm breaking the team management feature into smaller PRs.

I created two endpoints: `organizations#show_groupings` and `groupings#show`.

demo

Showcase:

screen shot 2016-06-19 at 12 34 31 pm

screen shot 2016-06-19 at 12 34 40 pm

screen shot 2016-06-19 at 12 34 48 pm

Changes are hidden behind the Flipper `team_management`. Would ❤️  to hear your feedback!


app/controllers/groupings_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  before_action :set_grouping
+
+  def show
+    not_found unless Classroom.flipper[:team_management].enabled? current_user
+  end
+
+  private
+
+  def set_grouping
+    @grouping = Grouping.find_by!(id: params[:id])
```
dinever: Should use slug here. I'm going to create a migration and make `grouping` sluggable.

app/views/groupings/_group.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <p class="team-description" itemprop="description"><%= pluralize(group.repo_accesses.count, 'student') %></p>
+    <div class="team-members">
+      <% group.repo_accesses.each do |repo_access| %>
+        <% student = repo_access.user %>
+        <%= link_to student.github_user.html_url, class: 'tooltipped tooltipped-s', 'aria-label' => student.github_user.login do %>
+            <%= image_tag student.github_user.github_avatar_url(30), class: 'avatar avatar-small ', height: 30, width: 30, alt: "@#{student.github_user.login}" %>
+        <% end %>
+      <% end %>
+      <% if group.repo_accesses.count > 7 %>
+        <%= link_to '#' do %>
```
tarebyte: I think you want `group.github_team.html_url` here

dinever: @tarebyte Actually I wanted this link to lead the teachers to a `group#show` route, which hasn't been implemented yet.

However, since we will implement a drag-and-drop feature on this page. It will need some UI adjustment so that we can show not only 7 but all students in each group.

app/views/groupings/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+                <% end %>
+            <% end %>
+          </div>
+          <div class="one-half column">
+            <% @grouping.groups.each do |group| %>
+                <% if cycle('odd', 'even') == 'even' %>
+                    <%= render partial: "group", locals: { group: group } %>
+                <% end %>
+            <% end %>
+          </div>
```
dinever: Hi @jasonlong, we're implementing a team management interface here. Teachers will be able to move students from a team to another by simply drag and drop the student avatars. 

It would be great if you can give me some feedback on this UI. Thanks!

/cc @johndbritton @tarebyte

jasonlong: Visually, I think the Masonry-style layout is nice. A couple thoughts though:

1. There doesn't seem to be any sort of affordance for users to know that these are drag and droppable. If nothing else, there should probably be a one-liner at the top of the page saying so.

2. If you're dragging an avatar to a drop area that's above or below the current viewport, that operation will probably be impractical. Will there be a fallback method?

3. The site uses a gray heading and white content area for the various boxes and these boxes flip that to have a gray content area.

4. Minor design nit: the line with the student count should be tightened up to reduce the whitespace above it.

dinever:
Update:

- [x] Add a `flash` to remind users that the student avatars are drag-and-droppable.
- [x] Change the style to gray heading and white content area.
- [x] Reduce the line height above the student count.

Before:

screen shot 2016-07-14 at 10 53 19 pm

After:

screen shot 2016-07-14 at 10 41 08 pm

/cc @tarebyte @jasonlong

jasonlong:
My 2 cents:

* Could we fix the team cards so they align with the left and right page margins?
* I'm not sure the message at the top needs to be in flash container. Could we try just having it show as regular text. Maybe the **drag and drop** phrase could be bold to make it stand out?
* The student count still feels like it makes the heading unnecessarily tall to me. Could that text (if it's really needed) be in a lighter gray next to the team name?

dinever:
@jasonlong Thanks a lot for the feedback! 😄 

I made some changes to the style.

Before:



After:

screen shot 2016-07-15 at 12 52 23 am

BTW maybe we can also add a screen shot 2016-07-15 at 1 00 10 am before the message(like dotcom showing ProTip messages at the bottom of each PR). What do you think? 😄 

screen shot 2016-07-15 at 1 01 43 am

/cc @tarebyte

jasonlong:
Awww yeah, that's feeling a lot better to me. The **ProTip** thing could be a nice touch. I think if we did that though, that blurb should go at the bottom of the page because it's like a "oh, did you know this?" vs. "you need to know this". If dragging and dropping will be a core part of the use of this page, I'd probably keep it at the top minus the ProTip label.

Oh, one last super-duper minor nit: we don't normally use italics, so I think just the lighter color would be sufficient for the student count.

dinever:
@jasonlong Thanks a lot for the feedback! 😄  I dropped the `ProTip` thing since drag-n-drop will be a core feature of this page.

Update:

- [x] Removed the italic font style.

screen shot 2016-07-15 at 11 17 41 am

@tarebyte I wonder if you can take another 👀  at this. After this one being merged we can move on to another PR. Thanks!

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,10 @@ def peek_enabled?
 
   private
 
+  def ensure_team_management_flipper_is_enabled?
```
tarebyte: Can we cut the `?` out of this method?

spec/controllers/groupings_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  describe 'GET #show', :vcr do
+    it 'returns success status' do
+      get :show, organization_id: organization.slug, id: grouping.slug
+
+      expect(response.status).to eq(200)
+      expect(assigns(:grouping)).to_not be_nil
+    end
+  end
```
tarebyte: Just for sanity sake let's add a test here to show if the user doesn't have the flag.

app/controllers/application_controller.rb
dinever: Nice catch! 😅  Done in 716c7063484b58ce7199a7a988895626288f1a32.

spec/controllers/groupings_controller_spec.rb
dinever: Done in 98fa77e8976c66fca1315048f61c0b3d0df8b150. Thanks! 👍

tarebyte:
@dinever I'm seeing double when I create a new Group.

screen shot 2016-07-16 at 14 15 02

dinever:
@tarebyte Thanks for pointing that out. 👍  I've fixed that in 3dc2503e52e002b1e8b237de608026c61652e492.

tarebyte:
👍 

screen shot 2016-07-16 at 17 39 28

PR #653, opened 2016-06-20
Student management migrations

dinever:
This is some migrations we may be using in the team management feature.

I added `slug` column to the tables `group` and `grouping`.


db/migrate/20160620143002_add_slug_to_groupings_and_groups.rb
```diff
@@ -0,0 +1,37 @@
...
+    add_index     :groupings, :slug
+
+    add_column :groups, :slug, :string
+    Group.all.each do |group|
+      group.slugify
+      group.save!
+    end
+    change_column :groups, :slug, :string, null: false
+    add_index     :groups, :slug
+  end
```
dinever: There is some data migrations to assign `title` to `slug`.

As the existing grouping data may result in duplicate slugs (e.g. a grouping named `Teams 1` and another grouping named `Teams-1` will have the same slug `teams-1`), I need to check the uniqueness of `slug` before migration.

dinever:
Hi @tarebyte, 

what do we think about using slug-based routing for groups & groupings?

db/migrate/20160620143002_add_slug_to_groupings_and_groups.rb
tarebyte: @dinever can we break this up into 3 migration files?

- Create the columns
- Transition the data
- Ensure the values can't be null

tarebyte:
> what do we think about using slug-based routing for groups & groupings?

Since they will be unique, I think that will be fine.

db/migrate/20160620143002_add_slug_to_groupings_and_groups.rb
dinever: @tarebyte Sure!

dinever:
@tarebyte 

> can we break this up into 3 migration files?

> Create the columns
> Transition the data
> Ensure the values can't be null

This was done in 343d6bcf9208888e4a3a74e76a7040035da53f81 😄

tarebyte:
Successful on production

tarebyte:
@dinever one thing to keep in mind if we do this again.

N + 1's if we need to make a call like `grouping.organization` or `group.grouping` we should save ourselves the SQL call and query like `Group.includes(:grouping).all`.

PR #658, opened 2016-06-22
Octicons migration

dinever:
Migration from font-based icons to SVG-based icons.

Changes:

- Introduced `octicons_helper`.
- Removed our helpers implementation `octicon` and `mega_octicon`.
- Removed `mega-octicon` class, since there is no way to scale SVG icons using normal width/height attributes in CSS as far as I know.
- Omit parentheses for octicons helper method call.

What do we think about this? Looking forward to your feedback! 😄 

/cc @tarebyte @johndbritton


Gemfile
```diff
@@ -47,6 +47,7 @@ gem 'sprockets'
 gem 'sass-rails', '~> 5.0'
 gem 'sidekiq', '~> 4.1.0'
 gem 'sinatra'
+gem 'octicons_helper', github: 'primer/octicons_helper'
```
tarebyte: Can we pin this down to something a little more concrete?

```ruby
gem 'octicons_helper', github: 'primer/octicons_helper', tag: 'v1.0.0'
```

app/views/assignments/_assignment_form_options.html.erb
```diff
@@ -6,13 +6,13 @@
 <dl class="form">
   <div class="form-checkbox">
     <label><%= f.radio_button :public_repo, true %> Public</label>
-    <%= mega_octicon('repo') %>
+    <%= octicon 'repo', :height => 32 %>
```
tarebyte: Can we use the ruby 1.9 hash syntax?

```ruby
<%= octicon 'repo', height: 32 %>
```

app/views/group_assignments/_group_assignment_form_options.html.erb
```diff
@@ -6,13 +6,13 @@
 <dl class="form">
   <div class="form-checkbox">
     <label><%= f.radio_button :public_repo, true %> Public</label>
-    <%= mega_octicon('repo') %>
+    <%= octicon 'repo', :height => 32 %>
```
tarebyte: Same comment about 1.9 hash here.

app/views/group_assignments/_group_assignment_form_options.html.erb
```diff
@@ -6,13 +6,13 @@
...
     <span class="note">Submit assignments using public repositories. All submissions will be visible to the world.</span>
   </div>
 
   <div class="form-checkbox">
     <label><%= f.radio_button :public_repo, false %> Private</label>
-    <%= mega_octicon('lock') %>
+    <%= octicon 'lock', :height => 32 %>
```
tarebyte: Same comment about 1.9 hash here.

tarebyte:
Left a few comments, thanks for doing this @dinever!

dinever:
Hi @tarebyte,

Thanks for the feedback.

> Can we use the ruby 1.9 hash syntax?

> <%= octicon 'repo', height: 32 %>

This was done in b98f7ddda6953da40fcf83585c45fada10650558. 🎉 

> Can we pin this down to something a little more concrete?

> gem 'octicons_helper', github: 'primer/octicons_helper', tag: 'v1.0.0'

When I changed this and run `bundle install`, Bundler raised an error:

```
Resolving dependencies..........
Bundler could not find compatible versions for gem "rails":
  In Gemfile:
    rails (= 4.2.6)

    octicons_helper was resolved to 1.0.0, which depends on
      rails (~> 3.2)
```

If we use the lastest version rather than the tagged version, it worked fine(done in b6aca1bd1f22417c1ec22b9265edc561159263b3):

```
gem 'octicons_helper', github: 'primer/octicons_helper', ref: 'fa28cb06c67ce0a1974a9e3238483d79157503c2'
```

Also, seems that I bumped several gems when doing `bundle install`, is it fine? Thanks! 😄

tarebyte:
@dinever I went ahead and undid the bundle update, I'd rather do that in it's own PR.

tarebyte:
Everything else looks fine though

tarebyte:
screen shot 2016-06-22 at 14 10 10

@dinever homepage looks pretty funky

dinever:
@tarebyte,

Oops! 😨  Missed the homepage, I'll go ahead and fix it.

dinever:
@tarebyte 

The homepage should be good now. Thanks!

tarebyte:
Nice work @dinever!

jonrohan:
Just randomly found this pull 🎉 Nice work, any feedback on `octicons_helper`? I'm planning to swap that in on dotcom soon.

> Removed mega-octicon class, since there is no way to scale SVG icons using normal width/height attributes in CSS as far as I know.

It's best to size these on the svg element. Works better cross browser.

dinever:
Hi @jonrohan,

Thanks for creating `octicons_helper`!

> any feedback on 'octicons_helper'?

The gem itself is pretty good. However, I ran into some issue when installing it.

When installing the latest release(v1.0.0) from RubyGems, Bundler raised this error message:

```
Resolving dependencies..........
Bundler could not find compatible versions for gem "rails":
  In Gemfile:
    rails (= 4.2.6)

    octicons_helper was resolved to 1.0.0, which depends on
      rails (~> 3.2)
```

The gem [depends on rails (~> 3.2)](https://github.com/primer/octicons_helper/blob/4a9efdc6188917f3b7c6d45e7de8b980242de431/octicons_helper.gemspec#L17), but we are using rails 4+. 

Then I tried to install it directly from GitHub([the latest commit](https://github.com/primer/octicons_helper/tree/fa28cb06c67ce0a1974a9e3238483d79157503c2)), the problem was solved. So it would be great if we can create a release for the newest version to fix this issue.

I'll also create an issue for `octicons_helper` about this.

Thanks again!

PR #659, opened 2016-06-24
Fix octicon syntax for repo autocomplete

tarebyte:
Closes https://github-education.airbrake.io/projects/116727/groups/1714557334048436580
/cc @dinever


tarebyte:
Confirmed fix in production.

screen shot 2016-06-24 at 16 17 43

Also we should support emoji, just sayin'.

PR #663, opened 2016-07-03
Upgrade GitHub Classroom to Rails 5

tarebyte:
Per the [Rails upgrade guide](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0) this updates the application from Rails 4.2.6 to Rails 5.0.

Would ❤️ some 👀  on this since this is such a big change.

Please feel free to pull this branch down and kick the tires to make sure I haven't missed anything.

/cc @johndbritton @mkcode @dinever @cyhsutw @NickTikhonov


app/controllers/application_controller.rb
```diff
@@ -72,7 +72,7 @@ def flash_and_redirect_back_with_message(exception)
       end
     end
 
-    redirect_to :back
+    redirect_back(fallback_location: (request.referer || root_path))
```
tarebyte: Would love to gets some 👀 on this. This seems like the most reasonable action here.

dinever: Hi @tarebyte, 

I found that `redirect_back` uses `request.referer` as the default redirect location. The argument `fallback_location` will only be used when the HTTP `Referer` is missing, which means that `request.referer` is `nil`. So I guess we can drop the `request.referer` here as it should always be `nil` when `fallback_location` is used. WDYT?

Please correct me if i'm wrong, thanks! 😄 

Reference:

- http://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back

tarebyte: Works for me! Yeah I overlooked the fact that it was optional.

tarebyte: Done in 567cb1d

config/environments/development.rb
```diff
@@ -27,15 +42,13 @@
...
-  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
-  # yet still be able to expire them through the digest params.
-  config.assets.digest = true
-
-  # Adds additional error checking when serving assets at runtime.
-  # Checks for improperly declared sprockets dependencies.
-  # Raises helpful error messages.
-  config.assets.raise_runtime_errors = true
+  # Suppress logger output for asset requests.
+  config.assets.quiet = true
```
mkcode: So nice!

config/puma.rb
```diff
@@ -1,15 +1,47 @@
-workers Integer(ENV['WEB_CONCURRENCY'] || 2)
-threads_count = Integer(ENV['MAX_THREADS'] || 5)
+# Puma can serve each request in a thread from an internal thread pool.
+# The `threads` method setting takes two numbers a minimum and maximum.
+# Any libraries that use thread pools should be configured to match
+# the maximum value specified for Puma. Default is set to 5 threads for minimum
+# and maximum, this matches the default thread size of Active Record.
+#
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
```
mkcode: FYI - The second arg is also the default value for `ENV.fetch`. No need for blocks in this file: 

http://ruby-doc.org/core-2.1.4/ENV.html#method-c-fetch

mkcode:
Looks great to me 🎉

config/puma.rb
tarebyte: So this is what was generated from the Rails update, but looking at https://github.com/JuanitoFatas/fast-ruby#hashfetch-with-argument-vs-hashfetch--block-code

It looks like having the second argument when it's a constant is faster. TIL

mkcode: Good to know. Thanks for the deep dive. 👍 

IMHO - the second argument way is better for style, not speed, but bonus! I would assume the generated update does it that way for n00bs that may want to evaluate some code; the block works for both values and evals.

johndbritton:
Looked through all the changes, everything seems reasonable to me. Might be worth branch deploying for a while before merging.

Can you open an issue with a checklist of followup items to address? I think you mentioned wanting to get us to default settings after this is merged.

tarebyte:
Got this error trying to login: https://github-education.airbrake.io/projects/116727/groups/1724971694461776679

```
uninitialized constant Classroom::Scopes Did you mean? Sprockets
```

tarebyte:
I can sign in locally just fine, will need to investigate.

tarebyte:
This is way to big for one PR, going to break it up

PR #670, opened 2016-07-18
Add blank state for team management pane

cyhsutw:
Hi guys,

This pull request adds blank state for the team management pane.

Before:
before

After:
after

I'm not sure if the text shown on the blank state is clear enough, would ❤️  to 👂  feedback/suggestions on this!

/cc @dinever @jasonlong


dinever:
Nice catch, thanks @cyhsutw!

PR #671, opened 2016-07-21
Make grouping editable

### LINKED ISSUE #499: An interface to groups
Currently we can not rename the group of teams after we have created them, so it may get a little bit messy if we have a lot of groups and always use the default group name:

screen shot 2016-03-16 at 10 50 12 am

What about we add an interface to groups so that we can rename them and have an overview of teams inside the group?

Like this:

screen shot 2016-03-16 at 10 53 32 am

This can be part of the [GSoC project - Enhanced user management](https://github.com/github/mentorships/issues/90)

What do you guys think? :smile:

dinever:
Simply makes the groupings editable.

Closes #499

**Showcase:** 

demo

Looking forward to your feedback! 😄


tarebyte:
👍

LennardF1989:
I see this is merged and deployed to production, yet I don't see it on https://classroom.github.com - am I missing something?

tarebyte:
@LennardF1989 it's behind a feature flag because it's not quite done yet. Sorry for the confusion!

PR #672, opened 2016-07-21
Add endpoints to show groups and remove student from groups

dinever:
Add a new interface for showing groups and a button for removing students from a group.

Although drag-n-drop will be the core feature for team management, we still need a fallback method for users who can not use the drag-n-drop feature because of the limitation of their browsers or devices.

Showcase:

demo

TODO:
- [ ] Style polishing.
- [ ] Figure out how to reuse the `remove_membership` endpoint for both AJAX call (drag-n-drop) and non-AJAX call (this fallback method).
- [ ] Teachers may need a way for them to navigate through each groups. At least they should have a button to get back to the grouping page.
- [ ] The `remove student` button is too large, I was thinking about a dropdown menu like dotcom:
  
  screen shot 2016-07-21 at 3 02 37 pm
  
  However we may need to introduce a vendor library for dropdown if we take this route.


app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def set_grouping
+    @grouping = Grouping.find_by!(slug: params[:grouping_id])
+  end
+
+  def set_group
+    @group = Group.find_by!(slug: params[:id])
```
tarebyte: To save ourselves an N+1 can we do something like this?

```ruby
  @group = Group.includes(:repo_accesses).find_by!(slug: params[:id])
```

aitrusgit:
This would be a great feature. I'm currently using github to handle group assignments in a large course and believe me, errors while groups are assigned ARE bound to happen (students joining the wrong team, stuff like that).

I need the capability of managing users and teams.

dinever:
Hi @aitrusgit, thanks for the feedback! We're currently working on this feature and it will be delivered soon.

nwoodthorpe:
I would love to see this merged, but it seem's to have gone stale. Closing, if there's progress on this, feel free to re-open.

PR #676, opened 2016-08-01
Create a new repository when revisiting assignment invitation link if it was deleted

### LINKED ISSUE #471: Undefined behavior on re-accepting an invitation link of a repo that has been deleted
If I delete a repo and try to accept the invitation by opening the invitation link again it results in an  undefined behavior 

```
You may receive an invitation to join @OOM-Blasters via email invitation on your behalf. No further action is necessary.

Your assignment has been created here: # 

```
The problem is 

1) The messsage shows `#` instead of repo name
2) It's not showing that the repo has been deleted

I have fixed [https://github.com/education/classroom/pull/470] this by displaying a message that the repo no longer exists. The fix is for individual repo. I can make a pull request for group assignment once this get merged.

cyhsutw:
Fixes #471.

- Check if the associated GitHub repo exists when a student visits the assignment invitation page of an assignment she already accepted.
- Creates a new Classroom AssignmentRepo/GroupAssignmentRepo if the associated GitHub repo is deleted.

TODO:
- [x] Add functionality to recreate repo for **individual assignment**
- [x] Add functionality to recreate repo for **group assignment**

/cc @hackerkid


spec/models/github_repository_spec.rb
```diff
@@ -18,7 +18,7 @@
     @client.delete_repository(@github_repository.id)
   end
 
-  describe '#present?', :vcr do
+  describe '::present?', :vcr do
```
cyhsutw: @tarebyte should we change the symbol from `#` to `::` for class methods?

tarebyte: @cyhsutw I'm ok with this change.

app/models/github_repository.rb
```diff
@@ -13,6 +13,10 @@ def get_starter_code_from(source)
     end
   end
 
+  def present?
```
tarebyte: @cyhsutw I'd like to add an optional `options` hash to these methods.

```ruby
def present?(**options)
  self.class.present?(@client, @id)
end

def self.present?(client, repo_identifier, **options)
  client.repository?(full_name, options)
end
```

This way when we ask if the repo is present we can do it without caching.

```ruby
# Example
assgnment_repo.github_repository.present?(GitHub::APIHeaders.no_cache_no_store)
```

cyhsutw:
@tarebyte @johndbritton This PR is ready for review. 
Would you mind taking some time looking at it?

app/models/github_repository.rb
cyhsutw: 🆒

cyhsutw: Will do it later today

tarebyte:
@cyhsutw this looks good to me! I have just the one comment above, and this should be good to ship.

Thanks for doing this!

spec/models/github_repository_spec.rb
```diff
@@ -18,31 +18,97 @@
...
+        it 'returns true if the repo is present' do
+          expect(GitHubRepository.present?(@client, 'rails/rails')).to be_truthy
+        end
+
+        it 'returns false if the repo is not present' do
+          expect(GitHubRepository.present?(@client, 'foobar/jim')).to be_falsey
+        end
+      end
+
+      context 'with options' do
```
cyhsutw: @tarebyte thanks for reviewing! 

I just modified `::present?` and `#present?` based on your feedback.
Also added some tests for the custom options, would you mind having a 👀  at them?

tarebyte: Looks good to me, thanks for making that change @cyhsutw!

tarebyte:
Looks good to me, thanks @cyhsutw!

tarebyte:
@cyhsutw I updated the method to use `no-cache no-store` just a heads up.

cyhsutw:
Thanks @tarebyte!

PR #678, opened 2016-08-03
Enable users to choose student identifier type when creating assignments

dinever:
This PR makes it possible for teachers to indicate a student identifier type when creating || editing an assignment.

Demo:

demo2

Would love to 👂  your thoughts! 😄


tarebyte:
@dinever maybe another line of text explaining what the student identifier is for.

@johndbritton is better that the wording than I am.

dinever:
@tarebyte Thanks for the feedback.

I'm not good at wording either, but what do we think about this?

> Students will be required to enter their identifiers before accepting the assignment.

Demo⬇️ :

screen shot 2016-08-08 at 7 57 08 am

/cc @johndbritton

PR #680, opened 2016-08-03
Add endpoints `add_membership` and `remove_membership`.

dinever:
Added two endpoints `add_membership` and `remove_membership` to `GroupsController`.

In the drag and drop feature we will be using these two endpoints.

Would love to 👂  your thoughts! 😄


app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  private
+
+  def set_grouping
+    @grouping = Grouping.find_by!(slug: params[:grouping_id])
+  end
+
+  def set_group
+    @group = Group.includes(:repo_accesses).find_by!(slug: params[:id])
+  end
```
tarebyte: Instead of `before_action` let's turn these into helper methods.

```ruby
def grouping
  @grouping ||= Grouping.find_by!(slug: params[:grouping_id])
end
helper_method :grouping

def group
  @group ||= Group.includes(:repo_accesses).find_by!(slug: params[:id])
end
helper_method :group
```

tarebyte:
One small nit, other than that this looks great!

dinever:
Hi @tarebyte,

> Instead of `before_action` let's turn these into helper methods.

> ```ruby
def grouping
  @grouping ||= Grouping.find_by!(slug: params[:grouping_id])
end
helper_method :grouping

> def group
  @group ||= Group.includes(:repo_accesses).find_by!(slug: params[:id])
end
helper_method :group
```

This had been done in f93f5b4ccf447fdde449c6ecd41db290bc3c295c. Thanks!

PR #683, opened 2016-08-08
Fix wrong Flipper check

dinever:
Changes:

1. In `app/views/organizations/_side_menu.html.erb`, changed `student_identifier_enabled?` to `team_management_enabled?` as it's actually flipping the team management menu item.
2. Extracted student identifier flipper check into `ApplicationController` to make it normalized with the team management flipper check.

Thanks!


app/views/organizations/_side_menu.html.erb
```diff
@@ -9,7 +9,7 @@
                 class: "menu-item #{'selected' if current_page?(organization_student_identifier_types_path(@organization))}" %>
   <% end %>
 
-  <% if student_identifier_enabled? %>
+  <% if team_management_enabled? %>
```
dinever: This is the fix, we are actually flipping the team management feature here. Sorry for the mistake.

PR #668, opened 2016-07-11
Database Migration and Model Update for Organization Webhook 

cyhsutw:
Hello guys,

This pull request is a part of #667. 

Summary of changes:

- Create two webhook related fields for `Organization`
- Add two methods to `create` and `delete` webhook on GitHub
- Add related test cases

Would love to hear your feedback on this 😄

Issue #667, opened 2016-07-11
Assignment Submission Flow


cyhsutw:
Hi everyone,

We're going to add a feature that students can choose a specific version of their works to submit for grading (a part of #572).

_This is a potential implementation, would love to hear your feedback/suggestions on this!_ 😄 
### Usage flows
- For **students**, the submission is really simple: create a _Release_ on GitHub.com for the version she/he would like the teacher to grade.
- For **teachers**, Classroom will aggregate the latest submissions so teachers can grade.
### Roadmap

Backend:
- [x] Create migrations and update models to support organization webhooks and due dates for assignments
- [ ] Add functionalities to set webhook for organization and create a controller to handle `ping` event
- [ ] Add functionalities to create commit statuses for `push` events and `release` events

Frontend:
- [ ] Add functionality to let teachers specify due date of assignments
- [ ] Add functionality to let organization admins to set webhook if there isn't one
- [ ] Expose submission statuses (on time/overdue) to students

Edit: 
As @jayvdb mentioned in [#720 (comment)](https://github.com/education/classroom/pull/720#issuecomment-249790724), we should also keep a list of submitted SHAs (along with the timestamp of the submission) so the grading process for teachers can be more flexible and reasonable.

nwoodthorpe:
I'd love to see this happen, but it's fallen inactive. If we decide we want to implement this, we'll re-open.

PR #694, opened 2016-08-15
Webhook Creation

cyhsutw:
Part of #667, follow up of #668.

This pull request:
- creates orgnization webhook when creating a new classroom (organization)
- handles `ping` event of the webhook
### TODO:
- [x] Add tests for `WebhookEventController`
- [x] Add background job to send `ping` event if a hook has not receive one yet


app/controllers/concerns/organization_webhook.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+module OrganizationWebhook
+  extend ActiveSupport::Concern
+
+  def create_organization_webhook
```
cyhsutw: Rubocop will complain if we put this method inside `OrganizationController`. 

I'm not sure if it's correct to extract this as a `concern`.

app/models/organization.rb
```diff
@@ -46,8 +46,9 @@ def slugify
     self.slug = "#{github_id} #{title}".parameterize
   end
 
-  def create_organization_webhook(webhook_url)
-    webhook = github_organization.create_organization_webhook(config: { url: webhook_url })
+  def create_organization_webhook(webhook_url, client = nil)
```
cyhsutw: I added an optional parameter `client` here just in case we randomly picked other admins who don't have enough token scopes to manipulate organization webhook.

cyhsutw:
@johndbritton @tarebyte I think this is ready for code review now.

Also, I left some questions in the diff comments, it would be greatly appreciated if you could take a 👀  at them.

Thanks!

cyhsutw:
Close this to re-trigger CI.

app/controllers/concerns/organization_webhook.rb
tarebyte: I'm fine with it being here for now.

config/routes.rb
```diff
@@ -52,6 +52,7 @@
       resources :assignments
       resources :group_assignments, path: 'group-assignments'
       resources :student_identifier_types, path: 'identifiers', except: [:show]
+      resources :webhook_events, path: 'events', only: [:create]
```
tarebyte: Can we update the route to only accept the GitHub user agent.

Example:
```ruby
constraints user_agent: %r{\AGitHub-Hookshot\/}, format: 'json' do
  resources :webhook_events, path: 'events', only: [:create]
end
```

app/controllers/webhook_events_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def event_handler
+    @event_handler ||= "handle_#{event_name}".to_sym
+  end
+
+  def verify_payload_signature
+    return unless Rails.application.secrets.webhook_secret.present?
+
+    algorithm, signature = request.headers['X-Hub-Signature'].split('=')
+    payload_validated = actual_payload_signature(algorithm) == signature
```
tarebyte: @cyhsutw I'd like to make a few small changes here. This is pretty much what you have, except we do a [`Rack::Utils.secure_compare`](http://www.rubydoc.info/github/rack/rack/Rack/Utils.secure_compare) instead of `==`.

I'd also like to send back statuses which gives GitHub a heads up that something is wrong 😁 

```ruby
  def receive_and_verify_payload
    payload_body = request.body.read
    render json: { message: 'No payload received' }, status: 400 if payload_body.nil?

    expected_hmac = OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha1'), Rails.application.secrets.webhook_secret, payload_body)

    expected_signature = "sha1=#{expected_hmac}"
    received_signature = request.env['HTTP_X_HUB_SIGNATURE']

    unless received_signature && Rack::Utils.secure_compare(received_signature, expected_signature)
      render json: { message: 'Invalid payload signature' }, status: :forbidden
    end
  end
```

app/controllers/webhook_events_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def actual_payload_signature(algorithm)
+    OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new(algorithm),
+                            Rails.application.secrets.webhook_secret,
+                            request.body.read)
+  end
+
+  def verify_organization_presence
+    payload_github_id = params.dig(:organization, :id).to_s
```
tarebyte: Ooo til [`dig`](http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig)

spec/controllers/webhook_events_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+              type: 'Organization',
+              id: 999
+            },
+            sender: {
+              id: user.uid
+            },
+            organization: {
+              id: organization.github_id
+            }
+          }
```
tarebyte: Can we store a JSON file of a hook provided, so that we aren't building it in the test?

app/jobs/ping_organization_webhook_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def should_perform_job?(organization, user)
+    (organization.present? && user.present?) &&
+      (organization.webhook_id.present? && !organization.is_webhook_active?)
+  end
+
+  def perform(organization_id, user_id)
+    organization = Organization.find_by(id: organization_id)
+    user = User.find_by(id: user_id)
+
```
tarebyte: I think we can condense this a little bit and take advantage of SideKiq's ActiveJob error handling https://github.com/mperham/sidekiq/wiki/Active-Job#customizing-error-handling

What do you think about this?

```ruby
# frozen_string_literal: true
class PingOrganizationWebhookJob < ApplicationJob
  queue_as :github_webhook

  rescue_from(GitHub::Error) do
    retry_job wait: 1.minute, queue: :github_webhook_failures
  end
  
  def perform(organization_id, user_id)
    organization = Organization.includes(:user).find(organization_id)
    user = organization.users.find(user_id)
  rescue ActiveRecord::RecordNotFound => error

    return if error.present?
    return if organization.webhook_id.present? && organization.is_webhook_active?

    organization.ping_organization_webhook(user.github_client)
  end
end

File: app/models/github_organization.rb [hunk 8]
```diff
@@ -107,6 +107,10 @@ def create_organization_webhook(config: {}, options: {})
     end
   end
 
+  def ping_organization_webhook(webhook_id)
+    @client.ping_org_hook(@id, webhook_id)
```
tarebyte: Can we wrap this in to catch any Octokit errors?

```ruby
def ping_organization_webhook(webhook_id)
    @client.ping_org_hook(@id, webhook_id)
  end
end
```

app/models/github_organization.rb
```diff
@@ -107,6 +107,10 @@ def create_organization_webhook(config: {}, options: {})
...
+  end
+
   def remove_organization_webhook(webhook_id)
     return unless webhook_id.present?
     @client.remove_org_hook(@id, webhook_id)
```
tarebyte: Just realized we aren't doing it here and I missed it, mind wrapping this up as well?

spec/controllers/webhook_events_controller_spec.rb
cyhsutw: Good idea! However, we'll need to create the `organization` and `sender` manually since we need them to be the same as the ones in the stored JSON payload.

tarebyte:
@cyhsutw I left some feedback sorry for the wait

spec/controllers/webhook_events_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  include ActiveJob::TestHelper
+
+  let(:organization)  { GitHubFactory.create_owner_classroom_org }
+  let(:user)          { organization.users.first                 }
+
+  describe 'POST #create' do
+    describe 'ping event' do
+      context 'valid payload signature' do
+        before(:each) do
+          params = JSON.parse(File.open("#{Rails.root}/spec/support/fixtures/webhook_payloads/ping.json").read)
```
cyhsutw: @tarebyte I placed the JSON payload at this path, is this okay?

spec/controllers/webhook_events_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          params.symbolize_keys!
+          params[:sender][:id] = user.uid
+          params[:organization][:id] = organization.github_id
```
cyhsutw: I modify the params manually since we need to verify the existences of the `sender` and the `organization`.

cyhsutw:
@tarebyte Thanks a lot for reviewing!

I've made changes based on your feedback. Would you mind taking a 👀  at them?

Thanks!

cyhsutw:
I'll mark it as <kbd>in progress</kbd> for now since this PR should be a part of *due date* feature, which will be built on *explicit assignment submission* feature.

Please see #720 for the *explicit assignment submission* feature.

cyhsutw:
Closing this in favor of [#756 (comment)](https://github.com/education/classroom/pull/756#issuecomment-249466026)

Issue #698, opened 2016-08-16
Can't remove classroom and start over


eduvik:
I tried creating a classroom as a test, but it defaulted to using an organisation that I don't want to have access to classroom. I tried to de-authorise this org but was unable to do so. So I de-authorised classroom entirely from my account, but now if I go to classroom.github.com I get a server error:
We're sorry, but something went wrong.
If you are the application owner check the logs for more information.

tarebyte:
@eduvik would you mind clearing your cookies for classroom and try logging in again.

Thanks!

johndbritton:
@eduvik Once you've cleared your cookies, you should be prompted to re-authorize. If you don't want to give Classroom access to all of your orgs, make sure that you have enabled third party access restrictions for your organization.

You can read more in this thread: https://education.github.community/t/github-oauth-permissions/1047

johndbritton:
Closing this issue as it's not actionable for us, but feel free to reopen if your problem isn't resolved.

PR #699, opened 2016-08-16
Fallback to root_path if we rescue an error

tarebyte:
This should fix https://github-education.airbrake.io/projects/116727/groups/1753280335681912528 and close https://github.com/education/classroom/issues/698

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -83,7 +83,7 @@ class ApplicationController < ActionController::Base
     begin
       redirect_to :back
     rescue ActionController::RedirectBackError
-      redirect_to :root_path
+      redirect_to root_path
     end
   end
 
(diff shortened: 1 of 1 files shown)


cyhsutw:
As mentioned in #698:

> So I de-authorised classroom entirely from my account, but now if I go to classroom.github.com I get a server error: We're sorry, but something went wrong. If you are the application owner check the logs for more information.

What do you think about checking the token validity if we rescue an error?

```ruby
def flash_and_redirect_back_with_message(exception)
  flash[:error] = exception.message
  
  unless flash[:error].present?
    case exception
    when NotAuthorized
      flash[:error] = 'You are not authorized to perform this action'
    when GitHub::Error, GitHub::Forbidden, GitHub::NotFound
      flash[:error] = 'Uh oh, an error has occurred.'
    end
  end

  begin
    if current_user.authorized_access_token?
      redirect_to :back
    else
      reset_session
      flash[:error] = 'Your session has expired. Please sign in again.'
      redirect_to root_path
    end
  rescue ActionController::RedirectBackError
    redirect_to root_path
  end
end
```

PR #703, opened 2016-08-17
Fix missing x of flash message

cyhsutw:
Just some small nits missing at #658.

Note: 
I updated primer-css in order to place the SVG #`x` icon at the vertical center of the `flash`. 
This should not introduce much appearance updates according to the diffs of the `.scss` files. 

/cc @dinever 

## Preview:

| Before | After |
|:------:|:-----:|
| before | after |


tarebyte:
👍

Issue #595, opened 2016-05-14
Allow Multiple Classrooms Per Organization


tabinks:
Now that GitHub has unlimited repositories for educational organizations (I'm making this assumption since my account was upgraded, although I never saw anything official from GitHub), it would be useful to have multiple classrooms per organization.  You are currently only allowed to have a single classroom per organization.

For example, if I had an organization named "University-CS101", I could have classrooms for "Fall-2015,Winter-2016,Spring-2016, etc.".  Previously, there could only be a single classroom for the organization and I would have to create organizations name "University-CS101-Fall-2015", etc.  This would allow educators to organize by the course.

johndbritton:
> it would be useful to have multiple classrooms per organization.

I think this is something we want to support eventually. For now, you can create another organization and apply for the education program at https://education.github.com/discount to get free private repositories for your course.

We decided to limit it to one classroom per organization at first because it's easier to move to multiple classrooms per organization than it is to move the other way.

Would love to have your input on how this would work, what do you imagine the flow looking like? Can you use more than one organization per account? How do you group repositories from multiple courses in one organization?

drbeco:
I support this idea. For example, one organization should be enough to put together diverse classrooms from one teacher (I teach 4 courses, -- each semester --).

Also, if we can have multiple owners in one organization, we can set other professors to be part of it. Then we would actually have the hierarchy:

Organization: University Foobar
Team1: Professors
Team2...N: different groups of students from the same class, or different classes, or a subset for each task and so on.
Classrooms: each professor can create several within the organization
Assignments: each classrooms can have multiples assignments, each one with a different team, that can or cannot intersect with others.

I'm new to these business (github education and classroom) but I like the way it is going, and I hope we will have more freedom to organize our stuff.

Thank you all!

johndbritton:
I've been thinking about this. What if we changed the mapping in Classroom to just be 1:1 organization with organization and within an organization in Classroom you could group assignments. Each group of assignments would be one course but all the repositories would live in one organization on GitHub.

One thing that complicates multiple classrooms in one organization is teams. You can't have two teams with the same name in two classrooms if they're both in the same GitHub organization.

A way around that would be to have a top level organization that has teams and let teachers group assignments. In Classroom you would have one organization for all of your courses, but you could look at each course independently even though some things are shared amongst the courses (like teams).

johndbritton:
@tarebyte what do you think about this?

drbeco:
[...]

PR #704, opened 2016-08-17
Rename Classroom to GitHubClassroom

tarebyte:
This is the first step to allowing us to have multiple Classrooms per organization. See feature request #595 

My thought here was that we rename the `Organizations` model/table to `Classroom` since it's makes more sense inside the codebase.

In order to do that though the app can't be called `Classroom`.

/cc @johndbritton @mkcode @dinever @cyhsutw


johndbritton:
This looks reasonable to me.

PR #681, opened 2016-08-03
Drag-and-drop front end mockup

dinever:
This one is a proof of concept of the drag-and-drop team management feature.

Demo:

demo2

Would love to 👂  your thoughts! Thanks!

PR #707, opened 2016-08-17
Drag-and-drop team management

dinever:
This one makes #680 and #681 to work together.

Showcase:

team_management

Would ❤️  to 👂  your thoughts. Thanks!


app/views/groupings/show.html.erb
```diff
@@ -1,5 +1,10 @@
 <%= render 'organizations/organization_banner' %>
 
+<%= javascript_tag do %>
+    window.organization_id = "<%= @organization.slug %>";
+    window.grouping_id = "<%= @grouping.slug %>";
+<% end %>
```
tarebyte: @dinever not sure how I feel about this.

Can we parse the url to get these values for us instead?

dinever: @tarebyte Sure, I'll try that. Thanks for the feedback! 👍

nwoodthorpe:
Same comment as your other PR. I love this change, but it seem's to have gone stale. Closing!

Issue #613, opened 2016-05-30
Dashboard for Teachers


cyhsutw:
Hi everyone,

We're going to build a dashboard of teachers and we're looking for feedback, suggestions and ideas about what items will be shown in the dashboard to help teachers better understand the activities of an assignment.

Here are some basic items that will be included:
- time of the last push
- number of commits per student
- diff statistics (lines added / removed) per student
- build (CI) status 

Please feel free to leave comments if you have some ideas, thanks! 😄

drbeco:
I believe a summary in a easy list of students, showing who did more commits, more lines of code, how many issues per student, pushs, wikipages, etc., that would allow me to evaluate a student in a glance.

Now I need to click on graphics, stats, student profiles, find them all by hand, it is really hard.

Thanks for asking the feedback. I hope mine is helpful to you.

cyhsutw:
Hi @drbeco 

Thank you so much for the ideas, great to hear your feedback from the point of view of a teacher.

> how many issues per student, pushs, wikipages, etc.

These metrics are quite interesting, I would love to hear how you utilize metrics like number of issues per student when you grade their work.

What I can think of now is that in a group assignment, teachers might like to know how active are their students in participating discussions with others group members.

Are their more insights that would help you grade the works of the students?

Looking forward to you feedback 😄

bfritscher:
Hi,
For my specific needs (web project course) I hacked together a small dashboard using github's api 
In addition to the discussed metrics (pushes,...) I also tracked a list of branches and tags.
Having a history of commits with different colors for different contributor helped to assess who was doing what and judging the quality of the commit texts.

I also implemented small assignment specific checks to test some conditions (maybe for version 2 ):
(https://github.com/bfritscher/github-classroom-dashboard/blob/master/app/scripts/controllers/main.js).
- gh-pages branch present
- specific tag present
- readme.md present
- gh-pages branch has a specific file in it that proves that it is a deployed website

Having direct links to the specific tags (version to be corrected), files, commits helped in navigating quickly.

cyhsutw:
Hi @bfritscher,

Thank you so much for your feedback and example dashboard project you built*!*

> Having a history of commits with different colors for different contributor helped to assess who was doing what and judging the quality of the commit texts.

I like this idea, we'll try to keep to charts/figures as readable as we can.

> Having direct links to the specific tags (version to be corrected), files, commits helped in navigating quickly.

[...]

PR #708, opened 2016-08-18
Dashboard for Teachers (time of the last push & build status)

cyhsutw:
Part of #613.

This pull request adds functionality for teachers to check the time of the last push and the build status of the latest commit.

**Preview**:

preview

Edit: This feature is now placed under Flipper `teacher_dashboard` flag.

Would love to hear you feedback! Thanks 😄


cyhsutw:
Currently, this function will cost a lot of time for the pages to load, I'll change the implementations using AJAX.

nwoodthorpe:
Hey @cyhsutw how close is this to being finished? I'd love to see this get merged, it looks great from what I can see.

nwoodthorpe:
Closing for now, feel free to re-open if there's progress.

cyhsutw:
@nwoodthorpe Thanks for the heads-up. I've merged `master` into this and made necessary modifications. Please review it at your convenience. 😄

spec/factories.rb
```diff
@@ -52,7 +52,9 @@
     end
 
     after(:build) do |organization, evaluator|
-      create_list(:user, evaluator.users_count, organizations: [organization])
+      if evaluator.users.count < evaluator.users_count
+        create_list(:user, evaluator.users_count, organizations: [organization])
```
cyhsutw: I added this check because this is adding random extra users to the organization. 

This will break the functionality of [`Organization#github_client`](https://github.com/education/classroom/blob/master/app/models/organization.rb#L37:L40) since the `access_token` of the randomly created user is not valid, which will make the API calls return 401.

nwoodthorpe: Sorry, could you explain this a bit more? What's creating random users in the organization?

Having to edit this random unrelated line of code that's been fine for 2 months worries me a little bit

nwoodthorpe:
Thanks for getting this sorted! It's a pretty hefty diff, I'll try to find some time today to step through it.

spec/factories.rb
cyhsutw: I'll try my best to explain it.

So this line of code will generate a list of users with length equals to `evaluator.users_count`, and the users are generated using [this factory](https://github.com/education/classroom/blob/master/spec/factories.rb#L74-L87). 

This is totally okay if we don't access any of the users generated by this line of code (and *yes*, there isn't a line in the codebase that does this). However, if we try to use any of the access tokens that are associated with these users to call GitHub APIs, it simply fails as these are just some random hex string.

This line of code will create a new user even if we already specified users to be associated with the organization (e.g. [GitHubFactory#classroom_org](https://github.com/education/classroom/blob/master/spec/support/github_factory.rb#L16-L26)).

This is the issue I faced when I run test examples:

```ruby
# => organization.users.size  <--- this is 2!
```

In this case, you can only get a working [organization.github_client](https://github.com/education/classroom/blob/master/app/models/organization.rb#L37:L40) on a 50% chance.

IMHO the reason we have this line of code is to ensure that there's at least one user in an organization, we don't need to add another user if we already got one. That's the reason I made this change.

Please let me know if there's anything I can help. Thanks! 😄

app/controllers/concerns/github_repo_status.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module GitHubRepoStatus
```
nwoodthorpe: +1 for using a concern here. Assignment and group-assignment are super similar, we should be using these more often.

lib/github/event.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def initialize(access_token, options = { auto_paginate: false })
+      options[:access_token] = access_token
+
+      @event_client = Octokit::Client.new(options)
+    end
+
+    def latest_push_event(repo_id, options = { per_page: 100, page: 1 })
+      events = event_client.repository_events(repo_id, options).select do |event|
+        event.type == 'PushEvent' ||
```
nwoodthorpe: These can be on the same line

lib/github/event.rb
```diff
@@ -0,0 +1,22 @@
...
+          (event.type == 'CreateEvent' && event.payload.ref.present?)
+      end
+
+      events.length.positive? ? events.first : nil
```
nwoodthorpe: Any reason for this vs `#empty`?

spec/factories.rb
nwoodthorpe: Right, because a token is chosen from a random user. In that case, this makes sense to me.

nwoodthorpe:
This looks really good to me! Thanks for finishing this up

nwoodthorpe:
Also I'd like to squash before we merge. You don't have to worry about it yet though.

lib/github/event.rb
cyhsutw: Nice catch! Should have used `#empty` for this. 

I'm going to simplify this further using `#first` only as the result will be the same according to the [docs](https://ruby-doc.org/core-2.4.0/Array.html#method-i-first).

lib/github/event.rb
cyhsutw: 👌

cyhsutw:
@nwoodthorpe Thanks for reviewing! I've updated the code to resolve the issues you spotted, please have a 👀 .

> Also I'd like to squash before we merge.

+1 on this. Would you like me to do that or this can be done using the *squash on merge* feature on GitHub?

nwoodthorpe:
Code LGTM

Yeah, squash on merge should be fine. @johndbritton for review

PR #717, opened 2016-08-21
Migrate Organization to Classroom

tarebyte:
Part two to complete feature request in #595.

This will **only** be the schema and name change. This is the big piece because we'll need to put the app into maintenance mode in order to migrate the table / update the columns.

The functionality of the app will be in another pull request.


tarebyte:
Too many conflicts, will try this again.

PR #718, opened 2016-08-22
Enable users to submit their identifiers before accepting an assignment

dinever:
This one enables students to submit their identifiers when opening an assignment invitation page.

student_identifier

**P.S.** I didn't hide these endpoints under Flipper as they are used by students. If we hide these endpoints under Flipper we will need to enable Flipper on every student opening the assignment invitation page. Also, as long as an assignment has no student identifier type selected the identifier submission page will not be triggered.

Would love to 👂  your thoughts. Thanks!


dinever:
Hi @tarebyte, I wonder if we can take a 👀 at this at your convenience.

/cc @dhowe

tarebyte:
This looks good to me @dinever, I'll deploy this later today.

dinever:
Thanks! @tarebyte

PR #727, opened 2016-08-30
Dev updates

mkcode:
@tarebyte - Here are some updates that I thought classroom could use as I was updating the scripts portion of the education app.

 * reuse config/sidekiq.yml across app
 * do not show old rails logs in foreman startup
 * update puma Gemfile and config to be consistent with rails 5

For the rails 5 puma related commits, I had created a fresh rails 5 app, copied the defaults, and merged in the app's settings, so it will be a no-op when updating to rails 5.


tarebyte:
Looks good to me, thanks @mkcode!

Issue #723, opened 2016-08-29
Cannot create an assignment with Unicode characters in title


SkyZH:
After click on create assignment, I cannot even access my classroom page https://classroom.github.com/classrooms/20321486-zodiac-robotics/

tarebyte:
@SkyZH we just merged #728 which will fix your problem. I've also back filled a possible slug for your assignment.

PR #728, opened 2016-08-31
Support slug field for assignment creating and updating

dinever:
Closes #410 #723

Preview:

slugs

Added another input field for assignment slug, so that users can decide the slug they want to use.

Would love to hear your thoughts!


app/models/concerns/sluggable.rb
```diff
@@ -9,6 +9,7 @@ module Sluggable
   end
 
   def slugify
+    return if slug.present?
     self.slug = title.parameterize
   end
```
dinever: @tarebyte I added this line so that we can still create the assignment even if the `slug` is not present in the form values. What do we think of this?

dinever:
Hi @tarebyte, since the slugs are editable through the form, I wonder if we still need to check the uniqueness of assignment titles.

tarebyte:
> since the slugs are editable through the form, I wonder if we still need to check the uniqueness of assignment titles.

I think we still should, but that's just my personal opinion.

app/assets/javascripts/assignments-slug.coffee
```diff
@@ -0,0 +1,357 @@
+charMap =
+  'À': 'a',
```
tarebyte: Yikes this is long, did you pull this from somewhere?

app/models/concerns/sluggable.rb
tarebyte: I think the slug should be present in the form, and we should require it for Assignment creation.

app/views/group_assignments/_group_assignment_form_options.html.erb
```diff
@@ -4,6 +4,12 @@
 </dl>
 
 <dl class="form">
+  <dt><%= f.label :slug %></dt>
+  <dd><%= f.text_field :slug, class: 'textfield' %></dd>
+  <span class="note">Slug is the user friendly and URL valid name of an assignment. It will also be used as the name prefix of assignment repositories.</span>
```
tarebyte: What about instead of `Slug` we just say `Assignment repository prefix` with the same explanation?

dinever:
Hi @tarebyte,

All changes mentioned above have been done, thanks!

Changes:

1. Made slug field required for assignment and group assignment creating: a4802c0b910f9c55ed7b446af7b3c8f6eab59f05
2. Change slug label to *Assignment repository prefix*: 51fe2aac2c44ea5c9d9d8da265a3ed2032474f8c
3. Made assignment and group assignment title unique among organization: b9a87e13eae203005920ac67cb5e4779cc8d83e9

tarebyte:
@dinever this is looking good to 🚢  I made a small change in 4125699 so that we can reuse the text for both forms.

tarebyte:
@dinever I also took the time to backfill all Assignments and GroupAssignments that didn't have a slug. It wasn't that many but hopefully this will make users experiences better.

Thanks again for this PR!

dinever:
Good job, thanks @tarebyte!

Issue #662, opened 2016-07-03
Internationalization roadmap


fabriciojr:
As discussed in #308, this issue provides a roadmap to the project internationalization at all. Before add any new language inside classroom, the strings will be fully replaced by I18n keys. Each of the following changes needs to generate one pull request, this atomicity will make easier the code review, testing, tracking and collaboration on it. 
### Milestone 1: String to key replacing at views

In this milestone, we'll replace the strings on all classroom views, including the staff tools area.
- [x] Replace strings to keys at **shared** views (#732)
- [x] Replace strings to keys at **pages** view (#737)
- [ ] Replace strings to keys at **organizations** views
- [ ] Replace strings to keys at **autocomplete** views
- [ ] Replace strings to keys at **assignments** views
- [ ] Replace strings to keys at **assignments_invitations** views
- [ ] Replace strings to keys at **assignments_repos** views
- [ ] Replace strings to keys at **groups** view
- [ ] Replace strings to keys at **group_ assignments** views
- [ ] Replace strings to keys at **group_ assignments_invitations** views
- [ ] Replace strings to keys at **group_ assignments_repos** views
- [ ] Replace strings to keys at **stafftools/repo_accesses** views
- [ ] Replace strings to keys at **stafftools/organizations** views
- [ ] Replace strings to keys at **stafftools/autocomplete** views
- [ ] Replace strings to keys at **stafftools/assignments** views
- [ ] Replace strings to keys at **stafftools/assignments_invitations** views
- [ ] Replace strings to keys at **stafftools/assignments_repos** views
- [ ] Replace strings to keys at **stafftools/groupings** view
- [ ] Replace strings to keys at **stafftools/resources** view
- [ ] Replace strings to keys at **stafftools/groups** view
- [ ] Replace strings to keys at **stafftools/group_ assignments** views
- [ ] Replace strings to keys at **stafftools/group_ assignments_invitations** views
- [ ] Replace strings to keys at **stafftools/group_ assignments_repos** views
### Milestone 2:  String to key replacing at controllers

In this milestone, we'll replace the strings that are located on controllers flash messages.
- [ ] Replace strings to keys at **StarterCode** concern
- [ ] Replace strings to keys at **ApplicationController**
- [ ] Replace strings to keys at **AssignmentInvitationsController**
- [ ] Replace strings to keys at **AssignmentsController**
- [ ] Replace strings to keys at **GroupAssignmentInvitationsController**
[...]

PR #732, opened 2016-09-08
Add i18n to shared views

CGA1123:
PR to complete the first objective in #662


app/views/shared/_header.html.erb
```diff
@@ -5,45 +5,45 @@
     <nav class="site-nav">
       <ul>
         <li>
-          <%= link_to "GitHub Education", 'https://education.github.com', target: '_blank' %>
+          <%= link_to t('views.shared.github_education'), 'https://education.github.com', target: '_blank' %>
```
tarebyte: @CGA1123 GitHub Education is standard no matter what language. Let's revert this back so that we don't have to duplicate this for every language we add.

CGA1123: @tarebyte Okay, will do.

tarebyte:
@CGA1123 it looks like your changes broke the invitation buttons. Mind checking it out?

screen shot 2016-09-08 at 11 10 45

CGA1123:
There is still a problem with a `undefined method `deep_symbolize_keys' for "Sign in":String`

Due to `aria-label => t('translation=string')` I believe.

CGA1123:
@tarebyte Nevermind, the `NoMethodError` was being thrown because `I18n.t` was not returning anything due to the structure of `config/locales/views/shared/en.yml` not being correct. 

Should be fixed now.

tarebyte:
Awesome thanks @CGA1123!

PR #733, opened 2016-09-08
I18n setup + shared/footer I18n

jiteshjha:
Hi all!

Following #662, I have added the initial I18n setup, introduced translate helpers to shared/footer partial as a start, and wrote a small test.

Let me know your thoughts!
Thanks!


tarebyte:
Hi @jiteshjha thanks for the pull request!

At this point because we have so much text, I'd like to forgo tests for checking to make sure the text is right. Would you mind ✂️  the test and capybara?

Other than that this looks good to me, thanks!

Issue #693, opened 2016-08-12
Add Link to Apply for Discount on Assignment Settings Page


sanicki:
On the Assignment Settings page under the **Private** option there should be a link to https://education.github.com/discount_requests/new

![screenshot_20160812-001600-01](https://cloud.githubusercontent.com/assets/6764844/17624706/a48f6adc-605a-11e6-986d-48968ef09232.jpeg)

> Submit assignments using private repositories. Submissions will only be visible to the submitter and organization owners. If you have not already, you may apply for unlimited free private repositories for this classroom organization [here](https://education.github.com/discount_requests/new).

PR #736, opened 2016-09-09
Add link to apply for discount on assignment settings page

rabelmarte:
Hey everyone!

I recently stumbled upon GitHub Classroom and I think that Spark (a student outreach organization that focuses on inspiring and educating K-12 students on the wonders of technology) might just jump on board.

At the same time, I would like to dig in a little bit more deeper and help out. Therefore, I started by tackling a simple request #693.

Here's a preview of what it looks like on the new group assignment page.
screen shot 2016-09-08 at 9 40 00 pm

I look forward to working with all of the contributors!


tarebyte:
Hi @rabelmarte thanks for the pull request!

This seems like something that should be more prominent on the page, and maybe something we should only show when the organization does not have any private repositories.

What do you think?

rabelmarte:
You’re right 😺! I shall look into the project a little bit more to get the flow of it and then I’ll give it another try!

rabelmarte:
As for the prominency, is this closer to what we had in mind @tarebyte ? 

screen shot 2016-09-10 at 10 21 03 pm

tarebyte:
@rabelmarte something like that I think would work.

Shekharrajak:
@rabelmarte , If you are not active, Can I work on it ?

PR #741, opened 2016-09-11
Fix the problem that can not save assignments with empty slug field

### LINKED ISSUE #740: Saving an assignment with an empty title and/or slug
If a user saves an assignment (individual or group) with an empty title and/or an empty slug on the settings page, the site crashes.

dinever:
Closes #740

When the user sets an assignment with an empty slug value. The assignment will be failed to save but the object `@assignment` still carries the empty slug, which causes the page not able to be rendered. So we reload the assignment before rendering the view when user sets an empty assignment slug to avoid the problem.


tarebyte:
@dinever mind adding a test so future @tarebyte doesn't break this?

dinever:
@tarebyte Sure 😄  Working on that!

dinever:
Hi @tarebyte, I believe this one is ready for a 👀

tarebyte:
:metal:

tarebyte:
fix_slug

PR #745, opened 2016-09-12
I18n for groups and groupings views

jiteshjha:
Hello!
As per #662

Thanks!


config/locales/views/groupings/en.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  views:
+    groupings:
+      manage: "Manage"
+      set_team_settings: "Set of teams settings"
+      drag_and_drop_student_avatars: "You can <b>drag and drop</b> the student avatars to move them from one team to another."
```
nwoodthorpe: Hmmm, it feels wrong to pull style information into the i18n locales. I'm not sure there's an alternative here besides splitting the string into 3. @johndbritton do you have experience with this?

config/locales/views/groupings/en.yml
```diff
@@ -0,0 +1,8 @@
...
+      group_has_no_teams: "\"%{group_title}\" does not have any teams."
```
nwoodthorpe: The title won't change between locales, so can the string just be "does not have any teams."? If we ever decide to change that variable name, or some sort of logic around it, ideally we don't have to go combing through locale files to find usages.

config/locales/views/groupings/en.yml
tarebyte: @nwoodthorpe yeah I'm not a fan of bringing in styles either. Not sure what the best way approach this is?

Do we need to emphasize **drag and drop**

nwoodthorpe: Yeah, we can probably drop the bolding.

PR #752, opened 2016-09-24
Reset blocking UI elements when navigating back

### LINKED ISSUE #750: UI freeze returning to assignment list
Steps to reproduce:

1. Login
2. Click on the orgs classroom
3.  Click on an existing assignment
4.  Press browser back button

Result:

Busy icon indefinately spinning

Expected results:

Return to a fubctional assignment list page

cyhsutw:
Closes #750.

Hide the `loading-indicator` when detecting a back event.

The reason that causes the UI to freeze is that the page is not fully reloaded due to TurboLinks.

Other `loading-indicator`s are not affected since all of them are related to form submissions which is not `GET` request thus are [not cached by TurboLinks](https://github.com/turbolinks/turbolinks-classic#initialization).


cyhsutw:
Maybe we should switch to the newer version of TurboLinks (v5) and use the built-in progress bar instead.

cyhsutw:
Close this in favor of #755

PR #756, opened 2016-09-25
Receive webhooks

tarebyte:
Largely off based on the work of @cyhsutw in https://github.com/education/classroom/pull/694.

This allows Classroom to start receiving [GitHub webhook events](https://developer.github.com/webhooks/#events).

Each event will have it's own background job, so for example  the [`ping`](https://developer.github.com/webhooks/#ping-event) event has the job `PingEventJob`. Which as of right now does nothing.

I'm extracting this work for a few reasons:

- The original PR is rather large, and I'd like to break it down for easier reviewing
- I wanted to make the route more flexible for all kind of events
- I wanted to move responsibility to the background job instead of at the controller level.

/cc @cyhsutw

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,12 @@ Rails.application.routes.draw do
 
   get '/autocomplete/github_repos', to: 'autocomplete#github_repos'
 
+  scope 'github', as: 'github' do
+    constraints user_agent: %r{\AGitHub-Hookshot/\w+\z}, format: 'json' do
+      post :hooks, to: 'hooks#receive'
+    end
+  end
+
   resources :assignment_invitations, path: 'assignment-invitations', only: [:show] do
     member do
       get   :identifier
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -14,6 +14,12 @@
 
   get '/autocomplete/github_repos', to: 'autocomplete#github_repos'
 
+  scope 'github', as: 'github' do
```
tarebyte: This is in case we ever decide to take in other hooks

cyhsutw:
@tarebyte Thanks!

I would love to split #694 into small chunks if it's still a preferable way to create webhooks for organizations.

tarebyte:
> I would love to split #694 into small chunks if it's still a preferable way to create webhooks for organizations.

@cyhsutw yeah for sure.

I've been playing around with a `creator` pattern to try and move some of the responsibility of creation and setup out of ActiveRecord callbacks.

What do you think about this idea: https://github.com/education/classroom/compare/creator-pattern

cyhsutw:
Thanks for the input! I like the idea about extracting the code into creators :cool:

PR #757, opened 2016-09-26
Update user in background job

tarebyte:
Instead of updating our current_user in the foreground on every page load, let's update them in the background so that we affect the performance of the page.

/cc @cyhsutw


cyhsutw:
Thanks @tarebyte, this looks great!

PR #762, opened 2016-09-27
Fix alias method chain deprecation warnings

tarebyte:
/cc @mkcode


mkcode:
Looks great :eyes: 

Still this one with the `peek` gem itself too: https://github.com/peek/peek/pull/95

mkcode:
Only shows up during the peek#results request though. I'm just ignoring it for now until that is merged.

PR #774, opened 2016-10-21
Docker Development Workflow

hermanzdosilovic:
Hi,

I have created special development setup based on Docker and Docker Compose. It will allow new developers to dive into Classroom development more quickly and less painfully.

Idea is that you don't need to have Ruby, Rails, PostgreSQL, etc. on your local machine. All you need is Docker and Docker Compose which define all dependecies and connect them together.

With this _Docker Development Workflow_ initial project setup lasts about 10 minutes and even less.

Please take a look at updated `README.md` where I breifly described procedure. :smile: 

Thanks,
Herman

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM ruby:2.3.1
+
+RUN apt-get update && apt-get install -y \
+  libpq-dev \
+  nodejs \
+  sudo \
+  vim
+
+ENV TERM xterm
+
+ARG DEV_USER
+ARG DEV_USER_ID
+RUN useradd -u $DEV_USER_ID -m -r $DEV_USER && \
+  echo "$DEV_USER ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers
+USER $DEV_USER
+
+COPY Gemfile /usr/src/app/
+COPY Gemfile.lock /usr/src/app/
+WORKDIR /usr/src/app
+RUN gem install bundler && \
+  bundle
+
+CMD sleep infinity
--- a/README.md
+++ b/README.md
@@ -132,6 +132,26 @@ script/server
 
 And that's it! You should have a working instance of GitHub Classroom located [here](http://localhost:5000)
 
+
+### Docker Development Workflow
+
+Docker Development Workflow allows you not to worry about project dependecies. You don't need to have
+Ruby, Rails, Node, PostgreSQL, etc. to work on this project. All you need is Docker and Docker Compose.
+
+First, install Docker and Docker Compose on your macihine. Follow instructions [here](https://docs.docker.com/).
+
+Next, you need to build your development environment. You will do this only once. To do that
+run `./dev build`. This will build new image called `classroom:dev`.
+
+After that you are ready to open your **development shell**. Development shell is bash
+shell inside Docker container. In this shell you have Ruby installed and you can run all the commands
+specified in this guide. Every time you need new development shell you will run `./dev shell`.
+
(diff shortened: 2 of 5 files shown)


tarebyte:
@hermanzdosilovic thank you so much for doing this! My apologies for not getting back to you sooner.

Let me take a look at this, and try it out on my machine.

nwoodthorpe:
This is pretty sweet, this app is a PITA to configure sometimes. @tarebyte did you ever get around to trying it out? If not, I can give it a whirl when I get a chance.

tarebyte:
@nwoodthrope give a spin and let me know what you think!

nwoodthorpe:
Looks like it works fine, took about 10 minutes to download and process all the images but everything worked fine. No issues with any of classroom's dependencies. I'm not familiar with docker-compose and wasn't able to figure out how to access the web container from my host. I might spend more time on this another day, if we can get that working, this is 💯

hermanzdosilovic:
@nwoodthorpe you can access the web container from your host by executing `./dev shell` in your host console. More generally when working with Docker Compose you can access to any container executing `docker-compose exec [SERVICE NAME] /bin/bash`. In this project there are 5 services: `postgres`, `redis`, `memcached`, `elasticsearch` and `app` as defined in `docker-compose.yml` file. So you can access web container from you host by executing `docker-compose exec app bash`. I wrote this `dev` script so that you don't need to worry about what Docker Compose command you need to use to access web container.

Let me know if I can help you more with this. :smile:

nwoodthorpe:
@hermanzdosilovic Sorry, I should have been more clear. I'm able to `./dev shell` into the app container and start the web server. But as a developer, I'm interested in hitting the web-server from a browser on my host and interacting with the app. Is that possible currently?

Cheers

tarebyte:
@nwoodthorpe I think the only way to do that would to spin up the necessary bits (postgres, elasticsearch, redis) as a docker container but actually run the rails server on your local machine.

@hermanzdosilovic would you be interested in exploring this, or does it sound like a really bad idea 😬

hermanzdosilovic:
@tarebyte @nwoodthorpe I just tried one more time whole process of doing this and I was able to run web application and access it on http://localhost:5000.

It is worth noting that you should run `script/server` script **inside** Docker container.

Here is whole procedure once again :smile:

On your host run `./dev build`, and after that run `./dev shell`. You are now inside Docker container. Inside Docker container run `./script/setup`. Now on your host in your favorite editor open `.env` and edit it with your needs. I did this using Vim, but I did it inside docker container. Editing your project inside Docker container and on your host is the same thing because project files from your host are mounted to container. Now save `.env` file and **inside** Docker container run `./script/server`. After that open http://localhost:5000.

hermanzdosilovic:
So the whole idea is that you run everything inside Docker container, but you edit your code on your host! This is powerful! :smile: All other dependency services like Redis, Postgres, etc. are also run with Docker but in different containers. All containers are run in [`host`](https://docs.docker.com/engine/userguide/networking/) network mode, which adds containers to host's network stack and you can use them as if they are installed through package manages on your computer. For example, you can open pgAdmin on your host and access Classroom database which runs inside Docker container.

Hope this makes sense. :smiley:

nwoodthorpe:
Strange, I've followed your steps but I can't access services running inside containers from my host. Doing `docker network inspect host` shows all the containers connected to the host network, and they're all interfacing with each other, but I can't access them. It may be an issue with my network config, I'll try to debug it when I have a bit more time.

nwoodthorpe:
This doesn't seem to work on our end. Closing in favor of #926

PR #777, opened 2016-11-01
Added webcomponents.js to support document.registerElement

### LINKED ISSUE #776: Group assignment creation is impossible with Firefox 49
Hi,

**Update**: this doesn't happen in Chrome.

When I try to create a group assignment for my organization (@ensc-glog), the **Create Assignment** button stays disabled event after filling all the fields. I am able to create an individual assignment for the same orga.

In both cases, the search field for adding starter code from a repo doesn't work.
bug classroom

cyhsutw:
Fixes #776.

*The base branch of this pull request is `stafftools` which is not yet merged into `master`.*

We will need [`document.registerElement`](https://developer.mozilla.org/en-US/docs/Web/API/Document/registerElement) in order to use `github/include-fragment-element`, which was added recently on the `stafftools` branch.

Looks like the latest Chrome supports the function out-of-the-box, while other browsers like Safari/Firefox do not. 

Using a [polyfill](https://github.com/webcomponents/webcomponentsjs) as suggested in [github/include-fragment-element](https://github.com/github/include-fragment-element) will fix this.

According to [WebComponents](https://github.com/webcomponents/webcomponentsjs#browser-support), the supported browsers for the `CreateElement` polyfill are aligned with browsers stated in [GitHub site policies](https://help.github.com/articles/supported-browsers/).

These changes are tested on macOS 10.11.6 with: 

- Chrome 54.0.2840.71
- Firefox 49.0.2
- Safari  10.0.1

/cc @tarebyte


tarebyte:
🤘 

@cyhsutw I redeployed master to fix the issue, but I'll still take the PR. Thanks!

PR #778, opened 2016-11-03
Move from Bonsai to Elastic Cloud

tarebyte:
I've been paged more times than I'd like to think about for Elasticsearch not being reliable.

This moves us from Bonsai to Elastic Cloud


tarebyte:
Stale, this broke things.

PR #788, opened 2016-11-22
Update visibilities of existing repos when updating the visibility of an assignment

### LINKED ISSUE #782: Changing repository visibility only affects new repositories
I was informed that my students could see each others' code for an individual assignment, so I changed the visibility in the assignment to private. Unfortunately, this only affects NEW repositories created under the assignment, not existing ones! This behavior is dangerous and misleading. Changing the visibility setting should affect all repositories under the assignment.

cyhsutw:
Closes #782 

Added a background job to update the visibility of existing assignment repositories when updating the visibility of an assignment/a group assignment.

IMHO, in most of the cases, this operation only affects a few repositories (in the early stage), thus would not consume a lot of API calls.

Not sure if it's a preferable to fix the issue, would love to hear your feedback on this. Thanks!


tarebyte:
Thanks for the pull request @cyhsutw!

Not sure how I feel about the running it only on update save.

This feels like a data integrity thing, so maybe we want to run this like once a day in batches?

I'll put some more brain cycles on this later today.

cyhsutw:
@tarebyte Thanks for the feedback!

> This feels like a data integrity thing, so maybe we want to run this like once a day in batches?
Totally agree with that, would adding another job that performs regularly be sufficient?

app/controllers/group_assignments_controller.rb
```diff
@@ -35,7 +35,8 @@ def edit
 
   def update
     if @group_assignment.update_attributes(update_group_assignment_params)
-      flash[:success] = "Assignment \"#{@group_assignment.title}\" updated"
+      AssignmentVisibilityJob.perform_later(@group_assignment)
```
tarebyte: What do you think about checking to see if the privacy setting was changed before kicking off the background job?

Otherwise if someone changes the title this will iterate through all of the repositories which is an expensive process.

cyhsutw: Agree. Will do 👍

app/controllers/assignments_controller.rb
```diff
@@ -29,7 +29,8 @@ def edit; end
 
   def update
     if @assignment.update_attributes(update_assignment_params)
-      flash[:success] = "Assignment \"#{@assignment.title}\" updated"
+      AssignmentVisibilityJob.perform_later(@assignment) if @assignment.previous_changes['public_repo']
```
cyhsutw: I'm using `previous_changes` here to check if `public_repo` attribute is changed.

cyhsutw:
@tarebyte I've updated the `#update` methods as your previous comment. 

Would you please have a look at them at your convenience?

app/controllers/assignments_controller.rb
cyhsutw: I'm using `previous_changes` here, but I'm not sure if this is a preferable way to detect changes.

tarebyte:
@cyhsutw I posted a WIP to solve #831 (https://github.com/education/classroom/compare/tarebyte/admin-rights-background) what do we think about applying this pattern to your PR?

cyhsutw:
@tarebyte That looks great to me! I'll finish up the remaining parts based on your design.

cyhsutw:
@tarebyte Sorry for the delay. I've made the changes you mentioned. 

Would you please have a 👀  at them at your convenience?

Thank you!

Gemfile
```diff
@@ -1,9 +1,8 @@
 source 'https://rubygems.org'
 
 ruby '2.4.0'
-gem 'rails', '~> 5.0', '>= 5.0.1'
+gem 'rails', '~> 5.0', '>= 5.0.2'
```
tarebyte: Apologies for the piggy backing, this annoyed me while I was running tests 😄

tarebyte:
Confirmed on production 👍

PR #790, opened 2016-11-22
Refactor GitHub resources

tarebyte:
We have a problem where every time we grab a `GitHub` attribute we go and git the cache instead of hitting the cache once and setting it all.

This fixes that and cleans up the logic to make it more readable.


tarebyte:
Welp it's a start at least.

PR #792, opened 2016-11-24
Allows repos to load async

tarebyte:
There are pages that take _really_ slow to load because we wait for all of the resource to render which is N+ 😱 network requests.

This uses the https://github.com/github/include-fragment-element and the https://github.com/webcomponents/webcomponentsjs projects so that we can load out resources afterwards.

async

It's not the prettiest, _but_ it will help with load times. It will also allow us to use Kaminari pagination without any extra javascript.


tarebyte:
This is what the row will look like when the request fails.

screen shot 2016-11-24 at 20 43 48

PR #799, opened 2016-12-02
Create Rake task to migrate Organization's default repository permissions to 'none'

tarebyte:
Now that the new Org membership API's have shipped https://developer.github.com/changes/2016-11-28-preview-org-membership/ we can use the API to update every existing org will have the default repository permission of 'none'.

This will ensure that there aren't anymore assignment leaks in the classroom orgs.

## To run
```shell
heroku run rake organization_default_repository_permission:migrate
```

/cc @johndbritton


tarebyte:
Task failed on an error I didn't account for 😢 

I got about 75% of the way there. Will run it again tomorrow.

tarebyte:
So instead of running the `authenticated_token?` call and then `admin?` we're just going to run `admin?`

tarebyte:
Done!

PR #801, opened 2016-12-04
Add tutorial videos to Classroom

tarebyte:
Per request of @sanicki this adds the Classroom tutorial videos to the app.

screen shot 2016-12-04 at 15 23 07

/cc @jasonlong for 🎨 critique

### DIFF
--- a/config/videos.yml
+++ b/config/videos.yml
@@ -1,0 +1,12 @@
+"ChA_zph7aao":
+  title: "Getting started"
+  provider: "youtube"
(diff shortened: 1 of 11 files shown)


config/videos.yml
```diff
@@ -0,0 +1,12 @@
+"ChA_zph7aao":
```
tarebyte: I thought about pulling from the Youtube API. However since these don't have all that often, and that would an extra dependency I opted to hard code the data.

tarebyte:
screen shot 2016-12-05 at 19 52 22

PR #805, opened 2016-12-10
Add creator for AssignmentRepo

tarebyte:
## The current situation
Right now we have a pretty severe coupling of our models to the GitHub API. This makes testing a pain, and doesn't separate concerns very well. I'd like to change that.

## What this does
This PR (which I'm looking for feedback on) creates an `AssignmentRepo::Creator`. The creator hold all of the code to setup an AssignmentRepo with a GitHub repository, add the user as a collab, push starter code etc.

This means that we can now make `AssignmentRepo`s in tests without having to setup a bunch for the API.

## What this looks like in the code
```ruby
# Creates normal model no side effects
assignment_repo = AssignmentRepo.create!(assignment: assignment, user: invitee)

# Creates a GitHub repository etc.
AssignmentRepo::Creator.perform(assignment: assignment, user: invitee)
```

## The problems
There aren't _major_ problems as far as I can tell.

- We still have one ActiveRecord callback (tries to delete the GitHub repository) that I'll deal with in another PR.
- There is some code duplication right now, but we can clean that up once we've migrated GroupAssignmentRepos to use this pattern.

## TODO
- [x] Tests, lot's of em.

<br>
❤️

---
/cc @johndbritton @mkcode @dinever @cyhsutw


tarebyte:
Confirmed it works on production!

PR #807, opened 2016-12-13
Delete Assignment/GroupAssignment repos when a repo is deleted on GitHub

tarebyte:
When a GitHub repository is deleted we should destroy the record of it on Classroom.

This PR allows will accept the [`repository`](https://developer.github.com/v3/activity/events/types/#repositoryevent) and if the action is `deleted` then we will try to do a lookup and see if we can find a record to destroy.


tarebyte:
assignment-delete

PR #809, opened 2016-12-14
Allow public assignments for orgs that don't have a plan.

### LINKED ISSUE #808: Students cannot use a public assignment repository
We have a public repository for a public individual assignment:
https://classroom.github.com/classrooms/19305404-codecool_bp_2016_1/assignments/from-python-to-java-pokerclient/settings
Base repository (also public):
https://github.com/CodecoolBP20161/from-python-to-java-pokerclient

But when our students would like to accept the invitation for this assignment, they get an error:
"Cannot make this private assignment, your limit of 0 repositories has been reached. You can request a larger plan for free..."

tarebyte:
Closes #808 because I'm an idiot.

I was making public assignments during my tests, but my test org has a plan so I never ran into this issue.

/cc @tamastompa @immanuelfodor

### DIFF
--- a/app/models/assignment_repo/creator.rb
+++ b/app/models/assignment_repo/creator.rb
@@ -150,6 +150,8 @@ class AssignmentRepo
     #
     # Returns True or raises a Result::Error with a helpful message.
     def verify_organization_has_private_repos_available!
+      return true if assignment.public?
+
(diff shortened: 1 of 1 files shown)


tarebyte:
Fix confirmed, will follow up with tests after this is rolled out.

screen shot 2016-12-14 at 08 28 37

tamastompa:
Thanks a lot for this quick fix!

PR #815, opened 2016-12-27
Remove users from organization

### LINKED ISSUE #812: Allow staff to remove `Users` from `Organizations`

nwoodthorpe:
Closes #812 

Gives staff the ability to remove owners (users) from classrooms (organizations). 

Users are removed via a button in stafftools/classrooms/show beside the owners name. Clicking the button and confirming the action in the modal attempts to delete the user from the GitHub organization. If that succeeds, the association is destroyed from our database. Otherwise, don't destroy the association and we display a message. This maintains consistency between GitHub and us.

When there's only one admin left, the button is greyed out and unclickable. If a request is made when only one admin is left, it errors on GitHubs end and we display an error.

I had to use a bit of a hack to make the modals send the correct request for each user at /views/stafftools/organizations/show.html.erb:82. Is a way we can do this without making invisible modals for each user?

Happy Holidays :)


tarebyte:
Thanks of the PR @nwoodthorpe!

I'd like to noodle a bit on whether we _actually_ want to remove the user from the organization.

I think for now I'd rather just remove them from the Classroom.

Another thing to think about, if that user is the creator of an assignment we need to make sure that someone else is set as the creator (might need to change that name, maybe maintainer?) so that things still work smoothly.

nwoodthorpe:
I made it so the user is removed from the organization as well because otherwise the user can just visit the classroom page to be re-added as a user. I don't have 100% context though. We can just remove them for the classroom for now, yeah. 

Ah I didn't notice the assignment thing. I'll take a look when I get some free time

tarebyte:
> I made it so the user is removed from the organization as well because otherwise the user can just visit the classroom page to be re-added as a user.

That's technically incorrect, the user must be an admin of the Organization to be added to the Classroom for that org.

Here is the code: https://github.com/education/classroom/blob/211cb410b03c615f0331fe80dc8d59cd20cbb650/app/controllers/concerns/organization_authorization.rb#L5-L12 that figures who is an admin when they land on a page.

nwoodthorpe:
Hm so maybe the best solution would be to demote the user from admin to user in the organization instead of removing?

tarebyte:
Honestly as a first go, I'd just be happy with us just removing the user from classroom as long as they aren't an assignment creator.

tarebyte:
@nwoodthorpe feel free to ping me when you're ready for 👀

nwoodthorpe:
Will do! Should be able to find time to finish this on the weekend. 

If they do own assignments, should we just show an error? I was thinking we just say it can't be done because they own an assignment, then once #823 is implemented we can improve the error message.

nwoodthorpe:
@tarebyte ready for 👀 !

app/controllers/stafftools/organizations_controller.rb
```diff
@@ -5,6 +5,21 @@ class OrganizationsController < StafftoolsController
 
     def show; end
 
+    def remove_user
+      not_found unless true_user.try(:staff?)
+
+      user = User.find(params[:user_id])
+
+      if Assignment.find_by(creator: user)
```
tarebyte: I know that this will return nil if nothing is found, but would you mind adding a `.present?`

config/routes.rb
```diff
@@ -91,7 +91,11 @@
       end
     end
 
-    resources :organizations, path: 'classrooms', only: [:show]
+    resources :organizations, path: 'classrooms', only: [:show] do
+      member do
+        delete '/remove_user/:user_id', to: 'organizations#remove_user', as: 'remove_user'
```
tarebyte: Not a huge fan of this route, what do you think about something like this:

```ruby
delete '/users/:user_id', to: 'organizations#remove_user', as: 'remove_user'
```

It's a little more REST friendly. We really don't need the remove in the route since it's a `DELETE` request.

spec/controllers/stafftools/organizations_controller_spec.rb
```diff
@@ -31,4 +31,60 @@
...
+        end
+
+        it 'redirects to stafftools organization path' do
+          expect(response).to redirect_to(stafftools_organization_path(organization.id))
+        end
+      end
+
+      context 'when user owns at least one assignment' do
+        before do
+          Assignment.create(creator: user,
```
tarebyte: Could you update this to use FactoryGirl?

```ruby
create(:assignment, user: user, organization: organization, title: 'Title')
```

nwoodthorpe:
@tarebyte all done!

Good point about the route. For some reason I didn't want to do users/:user_id, I thought it'd be confused with a route that deletes actual users. Totally forgot we're namespaced under organizations. 👍

tarebyte:
Thanks @nwoodthorpe!

nwoodthorpe:
Cheers :)

PR #820, opened 2017-01-05
Revert use of Repository Invitations

### LINKED ISSUE #819: Students cannot accept assignment invitations
When we create a new assignment, we have the invitation link.
When a user clicks on that, the Github asks "Authorize application", and "Reauthorization required".
Then when the user clicks on the "Authorize application" button, it doesn't do anything, but redirects to the same page.
So finally the user cannot make a new repo for the assignment :-(

tarebyte:
Closes #819 

Because of the scope changes in #818 I borked some students ability to accept assignments. I'm still on vacation so for the time being I'm going to revert the use of the Beta API so that we can use the old scopes.

### DIFF
--- a/app/models/assignment_repo/creator.rb
+++ b/app/models/assignment_repo/creator.rb
@@ -87,23 +87,15 @@ class AssignmentRepo
     # Internal: Add the User to the GitHub repository
     # as a collaborator.
     #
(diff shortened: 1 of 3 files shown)


tarebyte:
@tamastompa would you mind trying out the assignment now?

I'd like to get a confirmation before I merge this in 😄

tamastompa:
Yes, it's working now. Thank you very much!

PR #825, opened 2017-01-14
Add tutorial video to logged out homepage

### LINKED ISSUE #713: Include tutorial video on logged out homepage
[screen shot 2016-08-19 at 5 00 06 pm](https://www.youtube.com/watch?v=ChA_zph7aao)

https://www.youtube.com/watch?v=ChA_zph7aao

brks:
Closes #713.

Adds an additional button to the homepage, which opens a lightbox and plays the Getting Started video as seen below.

Demonstration


app/assets/javascripts/pages.js
```diff
@@ -0,0 +1,35 @@
+//= require jquery/dist/jquery
+//= require jquery.turbolinks
```
tarebyte: Since we're not including turbolinks, we actually don't need this require 😃

app/views/layouts/pages.html.erb
```diff
@@ -32,5 +32,7 @@
     </main>
 
     <%= render 'shared/footer' %>
+
+    <%= javascript_include_tag 'pages', 'data-turbolinks-track' => true, 'data-turbolinks-eval' => false %>
```
tarebyte: Since we're not using turbolinks for this page, we can actually just make the script

```erb
<%= javascript_include_tag 'pages' %>
```

app/assets/stylesheets/layout.scss
```diff
@@ -39,14 +39,20 @@ main {
...
+  }
 }
 
 .site-logo {
   width: 188px;
   height: 20px;
   margin-top: 0.3rem;
 
-  @media (max-width: 34em) {
+  @media screen and (max-width: 40em) {
```
tarebyte: So my SCSS is a bit rusty, mind explaining what this change does?

tarebyte:
@brks this more I think about it, the more I'm not a fan of the buttons side by side of each other.

It distracts from the call to action of signing in.

What do you think about reworking it to be something like what @jasonlong suggested in https://github.com/education/classroom/issues/713#issuecomment-241409595?

app/assets/stylesheets/layout.scss
brks: @tarebyte I've added this in mainly by habit, but by adding `screen`, the `@media` query is simply more descriptive as it is now specified that these styles are for screens and not the other media types defined in the CSS3 specification, such as `print` or `speech`. 

If you were to print out a webpage, for example, then the navigation becomes useless therefore you can specify a media query to remove the navigation when the site is printed, like this:

``` css
@media print {
    nav { display: none; }
}
```

brks:
@tarebyte I'd based my implementation from what @johndbritton had suggested in https://github.com/education/classroom/issues/713#issuecomment-241409868 as it was the most recent comment on the matter. I'd be happy to try that out, though! 🙂

brks:
@tarebyte Ready for you to take a look! I reimplemented as per @jasonlong's mockup.

Preview

cc/ @johndbritton

tarebyte:
fast-modal

The animation is kind quick and abrupt, would you mind maybe slowing it down just a tad so it feels smoother?

brks:
@tarebyte I've pushed an update. Is that any better?

tarebyte:
The animation is better, one last thing and then I think this will be ready to go!

dark-flash

When you close out of the modal there is a weird flash of light to dark to light again. Do you see it on your browser as well?

brks:
@tarebyte Ah yes, I see that. That should be fixed now.

tarebyte:
⚡️

PR #835, opened 2017-02-04
Use find_by for true_user method

### LINKED ISSUE #834: Session error thrown when db is reset
Any time I reset the database, I get errors thrown from the app trying to load users the session insists exist. Not really an issue in production, but it'd be nice to add some error handling so we don't need to clear the session every time.

@tarebyte  worth fixing? I haven't thought through the edge cases yet, but switching to `#find_by(id: ` should work. `@true_user` would be set to nil as if the session key didn't exist at all. 

```
ActiveRecord::RecordNotFound (Couldn't find User with 'id'=1): app/controllers/application_controller.rb:111:in `true_user'
app/controllers/application_controller.rb:64:in `current_user'
app/controllers/application_controller.rb:87:in `logged_in?'
app/controllers/application_controller.rb:107:in `staff?'
app/controllers/application_controller.rb:20:in `peek_enabled?
```

Method:
```Ruby
  def true_user
    @true_user ||= User.find(session[:user_id]) if session[:user_id]
  end
```

tarebyte:
Closes #834 

If we've wiped the database in development let's just go ahead and redirect back to login instead of blowing up.

/cc @nwoodthorpe


app/controllers/application_controller.rb
```diff
@@ -108,6 +108,6 @@ def staff?
   end
 
   def true_user
-    @true_user ||= User.find(session[:user_id]) if session[:user_id]
+    @true_user ||= User.find_by(id: session[:user_id])
```
nwoodthorpe: It might be worth adding a small integration test to show we do get redirected. Code LGTM though!

PR #839, opened 2017-02-06
Stafftools delete groups and groupings

### LINKED ISSUE #813: Allow staff to delete `Groups` and `Grouping`

nwoodthorpe:
Closes #813 

Staff can now delete groups and groupings from stafftools.

Also fixed a small link bug similar to #837 

I made `Grouping#destroy` remove all group-assignments under it because some views we're assuming all group-assignments had a valid grouping, which seems valid. Instead of having this in the controller, should we move it to the model and give GroupAssignments a `dependant: :destroy` relationship with Grouping?

@tarebyte 

screen shot 2017-02-05 at 4 50 33 pm

screen shot 2017-02-05 at 4 50 52 pm


app/controllers/stafftools/groupings_controller.rb
```diff
@@ -5,6 +5,16 @@ class GroupingsController < StafftoolsController
 
     def show; end
 
+    def destroy
+      org = @grouping.organization
+
+      GroupAssignment.where(grouping: @grouping).each(&:destroy)
```
tarebyte: So this could actually be:

```ruby
GroupAssignment.where(grouping: @grouping).destroy_all
````

File: app/controllers/stafftools/groups_controller.rb [hunk 2]
```diff
@@ -5,6 +5,15 @@ class GroupsController < StafftoolsController
 
     def show; end
 
+    def destroy
+      grouping = @group.grouping
+
+      @group.destroy
```
tarebyte: I think we'd like to handle the case where the group actually didn't get destroyed. Maybe something like?

```ruby
if @group.destroy
  flash[:success] = 'Group was destroyed'
  redirect_to stafftools_grouping_path(grouping.id)
else
  flash[:error] = 'Group was not destroyed'
  render :show
end
```

app/views/stafftools/groups/show.html.erb
```diff
@@ -26,7 +26,7 @@
         </tr>
         <tr>
           <td>Grouping</td>
-          <td><%= link_to @group.grouping.title, stafftools_grouping_path(@group.grouping) %></td>
+          <td><%= link_to @group.grouping.title, stafftools_grouping_path(@group.grouping.id) %></td>
```
tarebyte: Nice 👍

spec/controllers/stafftools/groupings_controller_spec.rb
```diff
@@ -7,6 +7,15 @@
 
   let(:grouping) { Grouping.create(organization: organization, title: 'Grouping 1') }
 
+  let!(:group_assignment) {
```
tarebyte: Is there a reason we need the `!` here?

Also we can clean this up a bit:

```ruby
let(:group_assignment) do
  create(:group_assignment, creator: user, organization: organization, grouping: grouping)
end
```

app/controllers/stafftools/groupings_controller.rb
```diff
@@ -5,6 +5,16 @@ class GroupingsController < StafftoolsController
...
+      @grouping.destroy
```
tarebyte: Same comment as https://github.com/education/classroom/pull/839/files#r99513000

spec/controllers/stafftools/groupings_controller_spec.rb
nwoodthorpe: Did `let!` for eager-evaluation, so that the assignment is created before the delete request runs in the test. Is there a better way to do this? For some reason the test is failing, but passing on my machine. Maybe your tidied version will work

nwoodthorpe:
Thanks for the review, will make another commit in a bit

nwoodthorpe:
@tarebyte Any clue why this test is failing? `./script/cibuild` and `rspec spec/controllers/stafftools/groupings_controller_spec.rb` both pass locally, I can't figure out why Travis is having this error 😕

tarebyte:
> Any clue why this test is failing? `./script/cibuild` and `rspec spec/controllers/stafftools/groupings_controller_spec.rb` both pass locally, I can't figure out why Travis is having this error 😕

My best guess would be to try recreating the VCR cassettes and try pushing that up.

nwoodthorpe:
Weird, pushed new cassettes and it still fails. I'll look at it when I have a bit more time

nwoodthorpe:
@tarebyte finally got this sorted, sorry for the huuuuuge delay. I was still on a forked remote for some reason, so my changes were based off of a different repo and making weird things happen.

All good now (y)

nwoodthorpe:
Ping @tarebyte for when you have time

nwoodthorpe:
Tests fixed.

Issue #240, opened 2015-09-22
Enterprise


hartmamt:
Can this use GitHub enterprise?  We have the academic unlimited license and this would be a great service to offer on top of that.

johndbritton:
@hartmamt Classroom for GitHub doesn't support GitHub Enterprise and we don't have a plan to add support right now.

That said, it is open source and it probably wouldn't be too hard for you to take the code and stand up an instance that works with your GitHub Enterprise. We use Octokit for all our GitHub API calls, their [enterprise docs](https://github.com/octokit/octokit.rb#working-with-github-enterprise) would be a good place to start.

Let's say it's not officially supported, but we'd consider accepting pull requests to add the feature.

Marking this as closed as there's no concrete action for us to take right now, but feel free to continue discussion.

mdwheele:
@johndbritton I've implemented support for GitHub Enterprise to use Classroom at NC State. I've got some cleaning up to do, but the gist of the work was:

* New environment variables added to `.env.example`, `.env`, `config/secrets.yml`.
* Expand configuration of `GitHubClassroom.github_client` to pull environment variables that determine if integrated with GitHub Enterprise or not (i.e. `GITHUB_ENTERPRISE=true`, `GITHUB_ENTERPRISE_URL`)
* API / OAuth paths are resolved off the base URL of the instance according to GHE documentation.
* There were several instances of Classroom code instantiating the Octokit client directly. I patched all instances I could find to use the "singleton" reference from the `GitHubClassroom` module. There may have been specific reasons why code was doing this, so this will need extra scrutiny during code review.

For the most part, after updating Octokit configuration and wiring a few things to it, everything seems to work as expected. My next step is to verify tests are still passing and also verify that they are not passing vacuously. I intend to send a PR very soon; I just didn't want to waste your time on premature work.

Do you see anything obvious I'm missing? Not being a Rails / Ruby developer, I'm struggling a bit to find all the magic, haha.

/cc @hartmamt

tarebyte:
@mdwheele how are you handing the use of Starter Code?

The API for that the source code importing: https://developer.github.com/v3/migration/source_imports/ is still under preview and as far as I'm aware not available for Enterprise yet.

mdwheele:
[...]

PR #857, opened 2017-03-23
Add support for GitHub Enterprise

mdwheele:
Related: #240

This is my first contribution to the project and I am not a Ruby/Rails developer by trade. I would appreciate additional scrutiny in review and am eager to do what's necessary to move this forward.

This PR adds support for integrating Classroom with a GitHub Enterprise instance through the use of a "feature toggle" in `.env`. To enable the integration, `GITHUB_ENTERPRISE` must be set to `true` (`false` by default) and `GITHUB_ENTERPRISE_HOSTNAME` must be set to the domain name of the enterprise instance. These environment variables are then exposed to `config/secrets.yml`. This may or may not be necessary, but seemed to be where initializers were getting their configuration from.

**Basic Implementation**

The `github_client` and `omniauth` initializers are modified to further configure `Octokit::Client` if integrated with GitHub Enterprise. Primarily, this means configuration of `api_endpoint`. In `omniauth`'s case, I also had to specify alternative `site`, `authorize_url` and `token_url`. On top of this, I replaced all occurrences of `Octokit::Client` instantiation with `GitHubClassroom.github_client` to make SURE I was at all points using the same client configuration. This broke tests (VCR expectations) and I made adjustments to account for client code that was previously passing in custom configuration to the client. Tests are green (with a side of ಠ_ಠ).

**Source Code Import API missing on Enterprise**

To account for the fact that the source import API is not available on Enterprise, I simply hid the form element responsible for collecting the repository URL. I have an open email with one of our contacts at GitHub asking for a timeline on the addition of this API to Enterprise. Ideally, starter code should work as it currently does when the API is released to Enterprise. This is why I didn't do too much more than a superficial toggle here. Additionally, there was a notice about negotiating unlimited private repositories on the "Add Assignment" page. I hide this if integrated with Enterprise, where private repositories are readily available.

**Tests**

My goal was to maintain passing tests, but did not add any new ones, yet. As far as I can tell, the most important testable behaviours are:

- [ ] Make sure alternate configuration is handed to `Octokit::Client` when Enterprise is enabled
- [ ] Possibly generate new cassettes for integration tests against Enterprise. Personally, I don't see much value here as it's kinda double-covering the Octokit client.

I can add tests for either of these. I just want to make sure I'm on the right track before going further. I also welcome any advice.

**Further Work**

When we figure all this out, I want to send another PR to allow [multiple forms of "Starter Code"](https://github.com/education/classroom/issues/240#issuecomment-274633133). In this way, I can work around the source import API blocker and provide value to my faculty. Their current processes are to simply drop a static `README.md` in each assignment repository.

Thanks in advance for review and I appreciate your time!


johndbritton:
@mdwheele First off, thanks for the contribution. It's great to have folks passionate about making this possible.

I think the approach we want to take for Classroom to be used with GitHub Enterprise is that it should be possible, but we won't make it a recommended usage due to the overhead of running your own instance of Classroom. We're not planning to support other installs of this app than the one at `classroom.github.com`, but for the adventurous I don't have an issue with us making it easier for them.

We do have one major blocker here in my opinion:

> To account for the fact that the source import API is not available on Enterprise, I simply hid the form element responsible for collecting the repository URL. 

Without the ability to import starter code, Classroom is a lot less useful and I don't think we should implement another way of distributing the starter code.

Before the Import API, we implemented cloning and pushing directly, but that was very unreliable especially for larger repositories.

tarebyte:
> I would appreciate your additional consideration on this. In my opinion (and as an Enterprise administrator) I feel it to be an acceptable trade-off that source import is not supported in the short-term because at the moment, we cannot use this wonderful product at all. Additionally, as a user of this software, I think having additional options for initial state of the assignment repository could be very useful for some (speaking selfishly 😄 ).

I'm happy to give the code in this state review. Like @johndbritton I'm pretty 👎 on having multiple assignment support features for importing code. However I'm fine with having limited support for enterprise. 

> If the source code API was available today, would this work be considered acceptable?

_Most_ of what you have should work however there are issues with the code that I should be able to help you iron out.

mdwheele:
> I'm happy to give the code in this state review. Like @johndbritton I'm pretty 👎 on having multiple assignment support features for importing code. However I'm fine with having limited support for enterprise.

I think that's a reasonable compromise and would appreciate the review.

mdwheele:
Heyo! Has there been any progress towards review of this work? If not that's cool. I have time tomorrow to work with it again.

config/initializers/github_client.rb
```diff
@@ -2,10 +2,24 @@ module GitHubClassroom
   def self.github_client(options = {})
     client_options = {
       client_id:     Rails.application.secrets.github_client_id,
-      client_secret: Rails.application.secrets.github_client_secret,
-      auto_paginate: true
+      client_secret: Rails.application.secrets.github_client_secret
     }.merge!(options)
 
+    if client_options.has_key?(:access_token)
```
tarebyte: This is getting a bit hairy with all of these `merge!` calls, let's change it to:

```ruby
def self.github_client(options = {})
  client_options = {}.tap do |opts|
    opts[:user_agent]    = "GitHub Classroom" # This is new, but I want to add it 😄
    opts[:auto_paginate] = true # I want to put this back so I added it here.

    if options.has_key?(:access_token)
      opts[:access_token] = options[:access_token]
    else
      options[:client_id]     = Rails.application.secrets.github_client_id
      options[:client_secret] = Rails.application.secrets.github_client_secret
    end
    
    if enterprise_instance?
      api_endpoint = "https://#{Rails.application.secrets.github_enterprise_hostname}/api/v3"
      opts[:api_endpoint] = api_endpoint
    end

    Octokit::Client.new(client_options)
  end
end

File: config/initializers/omniauth.rb [hunk 2]
```diff
@@ -1,5 +1,21 @@
...
 Rails.application.config.middleware.use OmniAuth::Builder do
   provider :github, Rails.application.secrets.github_client_id,
-                    Rails.application.secrets.github_client_secret,
-                    scope: 'user:email,repo,delete_repo,admin:org,admin:org_hook'
-end
+           Rails.application.secrets.github_client_secret,
+           {
+               :client_options => options
+           }
+end
```
tarebyte: Can you fix the line ending here?

config/initializers/omniauth.rb
```diff
@@ -1,5 +1,21 @@
...
+
 Rails.application.config.middleware.use OmniAuth::Builder do
   provider :github, Rails.application.secrets.github_client_id,
-                    Rails.application.secrets.github_client_secret,
-                    scope: 'user:email,repo,delete_repo,admin:org,admin:org_hook'
-end
+           Rails.application.secrets.github_client_secret,
+           {
+               :client_options => options
```
tarebyte: Can you use the updated Ruby hash syntax here?

```ruby
{
  client_options: options
}
```

config/initializers/omniauth.rb
```diff
@@ -1,5 +1,21 @@
+options = {
```
tarebyte: Let's clean with up as well.

```ruby
options.tap do |opts|
  scope: 'user:email,repo,delete_repo,admin:org,admin:org_hook'
  
  if GitHubClassroom.enterprise_instance?
    hostname = Rails.application.secrets.github_enterprise_hostname
    opts[:site]          = "https://#{hostname}/api/v3"
    opts[:authorize_url] = "https://#{hostname}/login/oauth/authorize"
    opts[:token_url]     = "https://#{hostname}/oauth/access_token"
  end
  
  Rails.application.config.middleware.use OmniAuth::Builder do
    provider :github, Rails.application.secrets.github_client_id,
             Rails.application.secrets.github_client_secret,
             { client_options: options }

File: config/secrets.yml [hunk 5]
```diff
@@ -11,9 +11,11 @@
 # if you're sharing your code publicly.
 
 default: &default
-  github_client_id:     <%= ENV['GITHUB_CLIENT_ID']     %>
+  github_enterprise_enabled: <%= ENV['GITHUB_ENTERPRISE'] %>
```
tarebyte: Mind aligning these again?

.env-example
```diff
@@ -1,3 +1,6 @@
+GITHUB_ENTERPRISE=false
```
tarebyte: Let's update this to `ENTERPRISE_ENABLED`

app/views/assignments/_assignment_form_options.html.erb
```diff
@@ -27,11 +27,14 @@
   </div>
 </dl>
 
+
+<% if ENV['GITHUB_ENTERPRISE'] == false %>
```
tarebyte: Let's use our helper that we have.

```erb
<% if !GitHubClassroom.enterprise_instance? %>
  ...
<% end %>
```

app/views/assignments/new.html.erb
```diff
@@ -19,7 +19,9 @@
         <%= render partial: 'assignments/assignment_form_options', locals: { f: f } %>
       </div>
 
+      <% if ENV['GITHUB_ENTERPRISE'] == false %>
```
tarebyte: Same comment here: https://github.com/education/classroom/pull/857/files#r110310002

nwoodthorpe:
@mdwheele has there been any progress here? I'm closing right now assuming this isn't happening anymore. Feel free to re-open if you're still working on it.

hjmjohnson:
:-1  I am disappointed that this important feature is not being actively pursued.

tarebyte:
@hjmjohnson pull requests are welcome if you'd like to pick this back up 😄

hjmjohnson:
OK.  I want to use Github for my educational teaching purposes.  

 

Could you please provide the set of technical skills (python, ruby, javascript) that will be needed to make acceptable contributions?  I’ll try to find someone with the necessary skills to make contributions.

 

 

Thanks,

Hans

tarebyte:
> Could you please provide the set of technical skills (python, ruby, javascript) that will be needed to make acceptable contributions?  I’ll try to find someone with the necessary skills to make contributions.

@hjmjohnson you'll want someone with Ruby and Ruby on Rails knowledge, a bonus would be if they understand the [GitHub API](https://developer.github.com/) but it is is learnable as you go.

hjmjohnson:
Thank you.

 

Currently I do not personally have those skills.  I’ll try to identify others who may have those skills.

 

Hans

mdwheele:
I can also pick this up and finish off the requested changes. My wife and I had our first child and things got... hectic. 

Most of the requested changes seemed very straight-forward and looked to be mergeable (correct me if I'm wrong)

hjmjohnson:
Dustin,

 

You will be my personal hero if you can complete this.  

 

Thanks,

Hans

mdwheele:
I've opened another PR to continue work. I'm going to resolve problems found during review and if tests fail, will likely be looking for guidance 😅

mdwheele:
And... houndci-bot is amazing... hahaha...

PR #867, opened 2017-05-15
Add Error pages for RoutingError and RecordNotFound exceptions

### LINKED ISSUE #432: User friendly 404 page
The 404 page is still the Heroku default, which isn't super helpful for folks. Let's add some custom 404 pages. Specifically, we might consider a custom page for invalid invitation links that says the link is invalid and to contact the organization owner. For other pages, a generic 404 probably makes sense.

BelieveC:
Modify existing `404.html` to handle Routing Exception and add `invalid_link_error.html` to handle RecordNotFound exception.

Closes #432

Screenshots: 
- `invalid_link_error.html`

link

- `404.html`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,6 +19,7 @@ class ApplicationController < ActionController::Base
               GitHub::Forbidden,
               GitHub::NotFound,
               NotAuthorized, with: :flash_and_redirect_back_with_message
+  rescue_from ActionController::RoutingError, ActiveRecord::RecordNotFound, with: :render_404
 
   def peek_enabled?
     logged_in? && current_user.staff?
@@ -32,5 +33,14 @@ class ApplicationController < ActionController::Base
 
   def redirect_to_root
     redirect_to root_path
+  end
+
+  def render_404(exception)
+    case exception
+    when ActionController::RoutingError
+      render file: Rails.root.join('public', '404.html'), layout: false, status: :not_found
+    when ActiveRecord::RecordNotFound
+      render file: Rails.root.join('public', 'invalid_link_error.html'), layout: false, status: :not_found
+    end
   end
 end
--- a/public/404.html
+++ b/public/404.html
@@ -7,7 +7,7 @@
     html,body,div,h1,h3,a{margin:0;padding:0;border:0;font-size:100%;font:inherit;vertical-align:baseline;}
 
     a{
-      color:blue;
+      color: #4078c0;
       text-decoration: none;
     }
 
--- a/public/invalid_link_error.html
+++ b/public/invalid_link_error.html
@@ -79,7 +79,7 @@
 
     @media(max-width:1440px){
       .site-footer {
-        padding: 11em 5em 3em 0em;
+        padding: 12.2em 5em 3em 0em;
       }
     }
     @media(max-width:1366px){
@@ -251,7 +251,6 @@
     <div class="content">
       <h3>Invalid invitation link</h3>
       <h4>Recheck your invitation link or contact your organization.</h4>
-      <h4><a href="/">Go back to classroom dashboard</a></h4>
     </div>
     <div class="clear"></div>
   </div>
--- a/public/invalid_link_error.html
+++ b/public/invalid_link_error.html
@@ -7,7 +7,7 @@
     html,body,div,h1,h3,a{margin:0;padding:0;border:0;font-size:100%;font:inherit;vertical-align:baseline;}
 
     a{
-      color:blue;
+      color: blue;
       text-decoration: none;
     }
 
@@ -76,7 +76,6 @@
       font-weight: 400;
       color: #4078c0;
     }
-
 
     @media(max-width:1440px){
       .site-footer {
@@ -247,11 +246,11 @@
 <body>
   <div class="not-found-container">
     <div class="notfound-top">
-      <h1>404</h1>
+        <h1>404</h1>
     </div>
     <div class="content">
-      <h3>Page not found</h3>
-      <h4>The page you are trying to reach does not exist, or has been moved.</h4>
+      <h3>Invalid invitation link</h3>
+      <h4>Recheck your invitation link or contact your organization</h4>
       <h4><a href="/">Go back to classroom dashboard</a></h4>
     </div>
     <div class="clear"></div>
--- a/public/invalid_link_error.html
+++ b/public/invalid_link_error.html
@@ -7,7 +7,7 @@
     html,body,div,h1,h3,a{margin:0;padding:0;border:0;font-size:100%;font:inherit;vertical-align:baseline;}
 
     a{
-      color: blue;
+      color: #4078c0;
       text-decoration: none;
     }
 
@@ -250,7 +250,7 @@
     </div>
     <div class="content">
       <h3>Invalid invitation link</h3>
-      <h4>Recheck your invitation link or contact your organization</h4>
+      <h4>Recheck your invitation link or contact your organization.</h4>
       <h4><a href="/">Go back to classroom dashboard</a></h4>
     </div>
     <div class="clear"></div>
--- a/spec/controllers/groupings_controller_spec.rb
+++ b/spec/controllers/groupings_controller_spec.rb
@@ -60,32 +60,23 @@ RSpec.describe GroupingsController, type: :controller do
   context 'flipper is not enabled for the user' do
     describe 'GET #show', :vcr do
       it 'returns a 404' do
-        expect do
-          get :show, params: {
-            organization_id: organization.slug,
-            id: grouping.slug
-          }
-        end.to raise_error(ActionController::RoutingError)
+        get :show, params: { organization_id: organization.slug, id: grouping.slug }
+        expect(response.status).to eq(404)
       end
     end
 
     describe 'GET #edit', :vcr do
       it 'returns success status' do
-        expect do
-          get :edit, params: {
-            organization_id: organization.slug,
-            id: grouping.slug
-          }
-        end.to raise_error(ActionController::RoutingError)
+        get :edit, params: { organization_id: organization.slug, id: grouping.slug }
+        expect(response.status).to eq(404)
       end
     end
 
     describe 'PATCH #update', :vcr do
       it 'correctly updates the grouping' do
         update_options = { title: 'Fall 2015' }
-        expect do
-          patch :update, params: { organization_id: organization.slug, id: grouping.slug, grouping: update_options }
-        end.to raise_error(ActionController::RoutingError)
+        patch :update, params: { organization_id: organization.slug, id: grouping.slug, grouping: update_options }
+        expect(response.status).to eq(404)
       end
     end
   end
--- a/spec/controllers/groups_controller_spec.rb
+++ b/spec/controllers/groups_controller_spec.rb
@@ -75,27 +75,24 @@ RSpec.describe GroupsController, type: :controller do
   context 'flipper is not enabled for the user' do
     describe 'PATCH #remove_membership', :vcr do
       it 'returns a 404' do
-        expect do
-          patch :add_membership, params: {
-            organization_id: organization.slug,
-            grouping_id: grouping.slug,
-            id: group.slug,
-            user_id: user.id
-          }
-        end.to raise_error(ActionController::RoutingError)
+        patch :add_membership, params: {
+          organization_id: organization.slug,
+          grouping_id: grouping.slug,
+          id: group.slug, user_id: user.id
+        }
+        expect(response.status).to eq(404)
       end
     end
 
     describe 'DELETE #remove_membership', :vcr do
       it 'returns a 404' do
-        expect do
-          delete :remove_membership, params: {
-            organization_id: organization.slug,
-            grouping_id: grouping.slug,
-            id: group.slug,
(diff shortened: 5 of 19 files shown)


app/controllers/application_controller.rb
```diff
@@ -111,4 +112,13 @@ def staff?
   def true_user
     @true_user ||= User.find_by(id: session[:user_id])
   end
+
+  def display_error_page(exception)
```
nwoodthorpe: Can we make this method name more descriptive? The ApplicationController can throw a few different errors, this method is specific to 404 errors.

app/controllers/application_controller.rb
```diff
@@ -111,4 +112,13 @@ def staff?
...
+    case exception
+    when ActionController::RoutingError
+      render file: Rails.root.join('public', '404.html'), layout: false
```
nwoodthorpe: This render line and the one below should have `, :status => :not_found` added. Otherwise, I believe rails will return a 200 response code.

public/404.html
```diff
@@ -1,67 +1,213 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>The page you were looking for doesn't exist (404)</title>
+  <title>Page Not Found | Classroom</title>
   <meta name="viewport" content="width=device-width,initial-scale=1">
+  <link href='//fonts.googleapis.com/css?family=Open+Sans:300italic,400italic,600italic,700italic,800italic,400,300,600,700,800' rel='stylesheet' type='text/css'>
```
nwoodthorpe: Is there a reason you're using 100% inline styles? I've seen this technique used with LB-rendered 503's, but since this was rendered from the web-server, I think we can assume that styles can be pulled down from the server as well. I might be missing something though, let me know!

public/404.html
```diff
@@ -1,67 +1,213 @@
...
-      <p>You may have mistyped the address or the page may have moved.</p>
+  <div class="not-found-container">
+    <div class="notfound-top">
+      <h1>404</h1>
+    </div>
+    <div class="content">
+      <h2>Github Classroom</h2>
+      <h4>Page not Found</h4>
+      <h3>The page you are trying to reach does not exist, or has been moved.</h3>
+      <h4><a href="https://classroom.github.com/">Get back to Classroom</a></h4>
```
nwoodthorpe: 2 comments:

1. Can we change this to `Back to Classroom` or `Go back to Classroom`?
2. In general, we try to avoid hard-coding URLs in rails, since hosts can change easily. Especially in this case where our users may be running their own instance of GitHub classroom on their own server. Bringing them to our hosted version would be quite confusing.

This should return the current root URL:

`<%= link_to "Link Text", root_url %>`

public/invalid_link_error.html
```diff
@@ -0,0 +1,213 @@
...
+<body>
+  <div class="not-found-container">
+    <div class="notfound-top">
+        <h1>404</h1>
+    </div>
+    <div class="content">
+      <h2>Github Classroom</h2>
+      <h4>Invalid Invitation Link</h4>
+      <h3>Recheck your invitation link or contact your Organization!</h3>
+      <h4><a href="https://classroom.github.com/">Get back to Classroom</a></h4>
```
nwoodthorpe: Similar comment here as above regarding hard-coded URLs

public/404.html
```diff
@@ -1,67 +1,213 @@
...
-  <div class="dialog">
-    <div>
-      <h1>The page you were looking for doesn't exist.</h1>
-      <p>You may have mistyped the address or the page may have moved.</p>
+  <div class="not-found-container">
+    <div class="notfound-top">
+      <h1>404</h1>
+    </div>
+    <div class="content">
+      <h2>Github Classroom</h2>
```
nwoodthorpe: The 'H' in GitHub should be capitalized :)

public/invalid_link_error.html
```diff
@@ -0,0 +1,213 @@
...
+  </style>
+</head>
+
+<body>
+  <div class="not-found-container">
+    <div class="notfound-top">
+        <h1>404</h1>
+    </div>
+    <div class="content">
+      <h2>Github Classroom</h2>
```
nwoodthorpe: Same here as above

nwoodthorpe:
Thanks for the PR @BelieveC! I had a few comments, let me know what you think. I like this change though, hopefully we can get it merged.

I'm a bit iffy about the styles, we're kind of straying away from GitHub's normal styling. I think @tarebyte would be better to comment on that though.

public/404.html
BelieveC: Alright!

app/controllers/application_controller.rb
BelieveC: Sure! I will make the changes :+1:

public/404.html
BelieveC: > Can we change this to Back to Classroom or Go back to Classroom?

I'll change it to Go back to Classroom

> In general, we try to avoid hard-coding URLs in rails, since hosts can change easily. Especially in this case where our users may be running their own instance of GitHub classroom on their own server. Bringing them to our hosted version would be quite confusing.

Agree! But pages `404.html` and `invalid_link_error.html` are pure html pages. We can't add `link_to` unless they have `.erb`. I will try n search if there is a workaround to insert root_url instead of GitHub classroom link. If you have any will be glad to implement it.

nwoodthorpe: As I understand (correct me if I'm wrong), error pages are often pure HTML because they're rendered in the case of an error, and if a dynamic error page causes an error, theres's no error page to render and bad things happen. A 404 isn't rendered because of a server error, so it's safe to let it be an `erb`, while 5xx errors should be static. Does that make sense?

nwoodthorpe:
This is starting to look really good, nice work :) See my comments about inline styles, the 404 link_to, and the design.

public/404.html
BelieveC: > As I understand (correct me if I'm wrong), error pages are often pure HTML because they're rendered in the case of an error, and if a dynamic error page causes an error, theres's no error page to render and bad things happen. A 404 isn't rendered because of a server error, so it's safe to let it be an erb, while 5xx errors should be static. Does that make sense?

Yeah! It makes sense. Is it dynamic error pages that you want to convey instead of public HTML pages right?

nwoodthorpe: Yes, I'm ok with the 404 being a dynamic page.

nwoodthorpe: Sorry, after having a discussion with some other people, let's not make this page dynamic. </a href='/'></a> will work fine. :)

public/404.html
nwoodthorpe: `stylesheet_link_tag` may allow you to load app styles here. If not, we might just have to hardcode.

BelieveC: > stylesheet_link_tag may allow you to load app styles here. If not, we might just have to hardcode.

I don't think there is an easy way out to add assets pipeline in Public HTML pages without using Dynamic Error Pages(correct me if I'm wrong). I have used 100% inline styles as I didn't find any easy way out to add assets pipeline to it. Maybe I am unaware of the superior way to handle it unless we precompile HTML error pages like [this](https://www.icelab.com.au/notes/precompiled-rails-static-404-and-500-pages/).

nwoodthorpe: I think you're correct. Sorry about that, the asset pipeline is hard :laughing:

BelieveC:
@nwoodthorpe If you want any more changes, Please let me know :wink:

public/404.html
```diff
@@ -1,67 +1,213 @@
...
-      <p>You may have mistyped the address or the page may have moved.</p>
+  <div class="not-found-container">
+    <div class="notfound-top">
+      <h1>404</h1>
+    </div>
+    <div class="content">
+      <h2>GitHub Classroom</h2>
+      <h4>Page not Found</h4>
+      <h3>The page you are trying to reach does not exist, or has been moved.</h3>
+      <h4><a href="https://classroom.github.com/">Go back to Classroom</a></h4>
```
nwoodthorpe: This was the last thing I think, href='/' will work on hostnames that aren't github.com. Same in the other 404 view.

BelieveC: Alright!

public/invalid_link_error.html
```diff
@@ -0,0 +1,213 @@
...
+
+<body>
+  <div class="not-found-container">
+    <div class="notfound-top">
+        <h1>404</h1>
+    </div>
+    <div class="content">
+      <h2>GitHub Classroom</h2>
+      <h4>Invalid Invitation Link</h4>
+      <h3>Recheck your invitation link or contact your Organization!</h3>
```
nwoodthorpe: We don't normally capitalize Organization on GitHub

BelieveC: Sure!

nwoodthorpe:
Just made 1 more pass through, noticed a couple things. Besides those, I'm happy with this PR :)

Curious to see what John things of the styles. He'll check it out when he gets time.

Thanks for working on this!!

BelieveC:
@tarebyte @nwoodthorpe 
Updated Design

`404.html`
notfound

`invalid_link_error.html`

invalidinvitation

nwoodthorpe:
@BelieveC Why green? We don't use green anywhere in our app except to show success or an action. Maybe a black/gray of some sort would be better?

jasonlong:
Hey y'all, sorry for the drive-by feedback 🚗 , but I thought I'd jot down a couple thoughts you could take or leave.

* There's no need to pull in a new font (Open Sans) since it isn't used elsewhere.
* Could we keep the dark header at the top so the logo could live in the same place as every other page (I realize it might need to be recreated w/ inline styles)? Then the big "GitHub Classroom" heading could be removed here (it has a lot of visual weight here).
* I dig that there's some descriptive content explaining the problem, but it has the least visual weight. Maybe make the "404" text a muted gray and the description larger. Should probably make the link blue as well since it's the most useful action.
* Speaking of that link, I wonder if "Go back to Classroom" could be more clear since the user is already in Classroom. Maybe "Go back to your Classroom dashboard" or similar?

nwoodthorpe:
I agree with all of that ^, thanks for the feedback :)

BelieveC:
> @BelieveC Why green? We don't use green anywhere in our app except to show success or an action.

I used it because it's the color of our favicon. @nwoodthorpe

BelieveC:
@jasonlong @nwoodthorpe 

> Could we keep the dark header at the top so the logo could live in the same place as every other page (I realize it might need to be recreated w/ inline styles)? Then the big "GitHub Classroom" heading could be removed here (it has a lot of visual weight here)

I doubt do we really need the dark header and extra inline CSS? Maybe we can remove `GitHub Classroom` heading as we have that info in the footer. Just error description with increased font size. It will cover both 2 & 3 point. 

Looking forward to suggestions

nwoodthorpe:
The GitHub 404 page might be a good place to look for inspiration. They don't have a dark header, and they also don't have a "GitHub" header. So maybe that's the way we do this one. Think that would work?

jasonlong:
Our (GitHub's) 404 page is pretty dated at this point, but yeah it's probably not super important to have the header. As long as the user can tell what site they're on (e.g. they followed a link from somewhere else) that would be the main goal.

BelieveC:
@jasonlong @nwoodthorpe 

`404.html` 



`invalid_link_error.html`

invalidlink

jasonlong:
Nice, this is looking better to me @BelieveC. Just a few minor nits:

* GitHub's house style (which I assume we'd want to carry over here) is to use sentence case for labels and headings (e.g. "Page not found" and "Invalid invitation link").
* Probably don't need an exclamation point with "...contact your organization!".
* Be sure to match the existing link color for the back links.

BelieveC:
@jasonlong @nwoodthorpe 

Updated SS

`404.html`

screenshot from 2017-06-01 00-25-18

`invalid_link_error.html`

screenshot from 2017-06-01 00-25-26

nwoodthorpe:
I'm liking this look a lot more. I think the only thing left is to change the color on the link to be the same as our other links.

BelieveC:
@jasonlong 

> Be sure to match the existing link color for the back links.

I didn't get this part. ` existing link color` ??

BelieveC:
@nwoodthorpe 

> I'm liking this look a lot more. I think the only thing left is to change the color on the link to be the same as our other links.

`Go back ..` link have a blue color as suggested by @jasonlong. The color used for links in the footer are same as other pages of the classroom. Do you want those link color in the footer to be changed?

nwoodthorpe:
Throughout the app we use this shade of blue: #4078c0

Could we change the color of your link to match that?

BelieveC:
@nwoodthorpe 

`404.html`

screenshot from 2017-06-01 01-01-05

'invalid_link_error.html`

screenshot from 2017-06-01 01-00-56

tarebyte:
@BelieveC sorry one more bit, we should remove the `Go to classroom dashboard` for the invitation 404's.

Since only teacher's have a dashboard this could get confusing.

BelieveC:
@tarebyte, Done :+1:

tarebyte:
Thanks @BelieveC!!

BelieveC:
Thanks @tarebyte @nwoodthorpe @jasonlong for helping me out :smile:

nwoodthorpe:
@BelieveC it's in production! Nice work :) https://classroom.github.com/!!!

PR #869, opened 2017-05-23
More informative error message when trying to create a classroom with an Org that's in use

### LINKED ISSUE #845: Identify the owner who has added the organization to a classroom
Currently there is a 1:1 relationship between organizations and classrooms.

If a classroom has already been added to an organization the tooltip "Another owner has already added this organization to GitHub Classroom" appears.

another

This is confusing for both organizations with a single owner, where the owner may not be aware of the 1:1 relationship, and to organizations with multiple owners, where the identity of the owner who established the classroom could be useful.

Since non-owners will not see the tooltip there is no privacy issue if the name of the owner is stated rather than using "Another owner."

nwoodthorpe:
Closes #845 

Display a more informative message with the owners login when we cannot use an org for a classroom because it is already used by another Classroom

screen shot 2017-05-23 at 9 24 02 am

@johndbritton @tarebyte

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -108,16 +108,19 @@ class OrganizationsController < ApplicationController
     @organization = Organization.find_by!(slug: params[:id])
   end
 
+  # rubocop:disable AbcSize
   def set_users_github_organizations
     @users_github_organizations = current_user.github_user.organization_memberships.map do |membership|
       {
-        classroom: Organization.unscoped.includes(:users).find_by(github_id: membership.organization.id),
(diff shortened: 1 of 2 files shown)


nwoodthorpe:
```
app/controllers/organizations_controller.rb:111:3: C: Assignment Branch Condition size for set_users_github_organizations is too high. [16.03/15]
```

Any suggestions here? I'm pretty sure this belongs as a property of the organization object, I don't think figuring this out in the view is a good idea.

nwoodthorpe:
Went ahead and disabled the Abc check for that method: I don't see any unnecessary complexity.

PR #872, opened 2017-05-24
Add assignment title blacklist

### LINKED ISSUE #507: Assignments should not be named `new`

### LINKED ISSUE #871: Individual & Group assignments should use the same type of validation.
Was just tackling a quick win and noticed this. Individual assignments are validated at the model level with nothing on the frontend.  Group assignments are validated on the front end and at the model level.

It's odd because users see errors for group assignments as not being able to click submit, but for individual assignments it shows an error dialog. We should make this more uniform. Either nuke the front end validations on group assignments or add front end validations for individual assignments.

nwoodthorpe:
Closes #871
Closes #507

2 things: 

1. Added client side validation for assignments/new. Group assignments had them, but individual assignments didn't for whatever reason.

2. Added a title blacklist that disallows naming your assignment 'new' or 'edit'. These 2 words are RESTful keywords in rails routing. The rest of the REST methods are implemented using HTTP methods, so these are the only 2 we need to blacklist.

As validations are only run when writing new stuff to the database, this won't corrupt old assignments that decided to use these names. These assignments were literally impossible to view anyways.

@johndbritton


nwoodthorpe:
As John said in #871, we should be clear about what's preventing the user from submitting.

With our current strategy of graying out the button until everything is fine, that's hard unless we want the page to show errors on-load on all fields that aren't filled in yet, which is a shitty UX.

I'm going to let the submit button always be enabled, and then if they submit the form incorrectly, we can let them know exactly what's wrong. This will tidy the code quite a bit (Our query files for both assignment pages will disappear) and make the UI a bit easier to use imo. 

There's a small issue with group-assignments/new throwing a 500 if you submit without a team selected, but besides that, this will be pretty straightforward. :fire:

nwoodthorpe:
I got rid of all of our JQuery stuff for the reasons above. Now when the user tries to submit, they get a list of errors on the page, which is the way we've done this sort of flow in other forms.

There's a styling issue that has to do with our use of <p> tags in the flash message. I want to fix that in a seperate PR though incase it ends up being something that requires a bunch of changes in a bunch of files.

@johndbritton 

tldr:

1. Added name blacklist
2. Removed dynamic as-you-type error checking on assignment/new and group-assignment/new
3. Button isn't grayed out until your form is good anymore, it made showing what was wrong challenging
4. Submitting the form without a grouping doesn't 500 anymore
5. Submitting an errored form displays a nice list of errors

I'll squash when we're good to merge.

screen shot 2017-05-24 at 1 36 14 pm

app/models/assignment.rb
```diff
@@ -25,11 +25,13 @@ class Assignment < ApplicationRecord
...
   validates :title, length: { maximum: 60 }
   validates :title, uniqueness: { scope: :organization_id }
+  validates :title, exclusion: { in: %w[new edit],
+                                 message: 'should not be a reserved word' }
 
   validates :slug, uniqueness: { scope: :organization_id }
   validates :slug, presence: true
   validates :slug, length: { maximum: 60 }
-  validates :slug, format: { with: /\A[-a-zA-Z0-9_]+\z/,
+  validates :slug, format: { with: /\A[-a-zA-Z0-9_]*\z/,
```
nwoodthorpe: The reason I did this is sort of buried above. With the old regex, the user would be displayed this validation failure AND the presence validation failure if a slug wasn't present. Now we only tell them the format is incorrect if there's actually a slug to check.

app/services/group_assignment_service.rb
```diff
@@ -8,10 +8,17 @@ def initialize(new_group_assignment_params, new_grouping_params)
...
+    if grouping_info_valid?
+      group_assignment.grouping = Grouping.where(id: @grouping_id).first_or_create!(@new_grouping_params)
+    end
 
     group_assignment.build_group_assignment_invitation
     group_assignment
   end
+
+  def grouping_info_valid?
+    (@grouping_id != '') || (@new_grouping_params[:title] != '')
```
tarebyte: Is there a reason we are guarding against the id being nil?

In ActiveRecord `where` on `nil` will just return an `[]`

Also if we wanted to check to see if the id is there I'd rather do

```ruby
@grouping_id.present?
```

Rather than check if it's not equal to empty string

nwoodthorpe: Right, I keep forgetting about `present?`. The method checks to see if either an id is present, or a title. If neither is true, we try to create a Grouping with no title which throws an exception, which is why I added this check. The method reads much better with `.present?` now

nwoodthorpe:
@tarebyte done! I took your idea of a global blacklist and turned it into a concern that we can include into other models. Since you can't pass arguments to concerns, it exposes a validator that we pass the field name to.

nwoodthorpe:
:+1:

PR #883, opened 2017-05-30
Extract logic from ApplicationController into digestable dependency files

tarebyte:
Our controllers are just a bit unwieldy, this is my starting attempt to create a better pattern for them.

This PR _just_ updates the `ApplicationController`

This also adds a `feature_enabled?` helper method for all model that are `Flippable`

/cc @nwoodthorpe


nwoodthorpe:
Code looks a lot cleaner, I like this.

My only concern is that this change will make tracing code, especially callbacks, in our controllers way harder. E.g. if you're trying to figure out what code is running for an action, you need to check all the dependency files for callbacks instead of just looking at the top of the file.

Thoughts?

tarebyte:
I @nwoodthorpe I see your concern for sure, but since the files are pretty small and are grouped pretty easily together I personally found it easier to search through.

That may be because I wrote most if not all of that code though 😝 

We also only have one callback in the `ApplicationController` but I do see what you mean about the helper methods.

Maybe we want to define those in the main file?

nwoodthorpe:
Yeah that could work well. Maybe something like this:

```
class ApplicationController < ActionController::Base
  # Prevent CSRF attacks by raising an exception.
  # For APIs, you may want to use :null_session instead.
  protect_from_forgery with: :exception

  require 'application_controller/authentication_dependency'
  require 'application_controller/errors_dependency'
  require 'application_controller/feature_flags_dependency'

  before_action :authenticate_user! # authentication_dependency
```

PR #889, opened 2017-06-02
Update Group Assignment Invitations Controller

tarebyte:
This moves the Invitations controllers to a better unification or logic. I have

- Updated the `GroupAssignmentInvitationsController` to use the `InvitationsControllerMethods` module.
- Moved the `success` action rendering to one view with a view model.
- Removed identifier logic and views

## TODO
- Finish tests for view model
- Move logic from controller to the model (another PR)
- Verify code paths


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -1,70 +1,64 @@
...
-    @group ||= repo_access.groups.find_by(grouping: group_assignment.grouping)
+  def check_user_not_group_member
+    return if group.blank?
+    redirect_to accept_group_assignment_invitation_path
   end
-  helper_method :group
 
-  def create_group_assignment_repo(selected_group: group, new_group_title: nil)
-    users_group_assignment_repo = invitation.redeem_for(current_user, selected_group, new_group_title)
+  def create_submission(selected_group: group, new_group_title: nil)
```
nwoodthorpe: Any way we could re-word this? We're introducing the concept of assignment submissions with the deadline stuff, so I could see this getting a little confusing.

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -74,63 +68,50 @@ def create_group_assignment_repo(selected_group: group, new_group_title: nil)
...
   end
-  helper_method :group_assignment_repo
 
-  def group_params
-    params
-      .require(:group)
-      .permit(:id, :title)
-  end
+  def ensure_submission_repository_exists
+    return not_found unless current_submission
```
nwoodthorpe: `not_found` raises an exception, so I'm not sure returning here really does anything, since it'll never actually return.

nwoodthorpe:
Code lgtm so far :+1: I like what we're doing here, the more places we can generalize assignment/groupassignment code, the better

PR #892, opened 2017-06-02
Update app to Rails 5.1 and move JavaScript over to webpack

tarebyte:


tarebyte:
I'm definitely going to have to split this up.

PR #895, opened 2017-06-03
Move from bower to npm/yarn

tarebyte:


tarebyte:
According to [libraries.io](https://libraries.io/npm/include-fragment-element) include-fragment-element does not have a license.

That's not true it does, but npm isn't picking it up. I've opened up a PR https://github.com/github/include-fragment-element/pull/28 to fix that.

PR #901, opened 2017-06-06
Add short URLs to assignment invitations

nwoodthorpe:
This is something that's been asked for a few times now. Still not 100% sure if this is how we want to go about implementing it, but it was easy enough to throw together.

This implementation adds a short_key field on the invitation models. We generate them just like we generate the normal slugs.

There's 2 routes for short urls, `/a/<ID>` for assignments and `/g/<ID>` for group assignments. They can be the same thing, it just means we need to maintain ID uniqueness between the two tables.

If a request to one of these routes matches an invitation, we simply 302 to the long version of the invitation.

@tarebyte @johndbritton


db/schema.rb
```diff
@@ -56,12 +58,23 @@
     t.index ["slug"], name: "index_assignments_on_slug", using: :btree
   end
 
+  create_table "deadlines", force: :cascade do |t|
```
tarebyte: Looks like this snuck in here 😄 mind removing it?

app/helpers/invitation_helper.rb
```diff
@@ -1,28 +1,47 @@
...
+    #
+    # invitation - An AssignmentInvitation or GroupAssignmentInvitation
+    #
+    # Examples
+    #
+    # InvitationHelper.attributes(assignment.invitation, request.base_url)
+    # #=> {
+    #      :type => "assignment_invitation",
+    #      :key  => "7309ffd1f7b5e7af09dfcd11fab371d7",
+    #      :url  => "http://github.dev/assignment-invitations/7309ffd1f7b5e7af09dfcd11fab371d7"
```
tarebyte: Mind updating the TomDoc here?

app/models/group_assignment_invitation.rb
```diff
@@ -40,6 +43,10 @@ def assign_key
     self.key ||= SecureRandom.hex(16)
   end
 
+  def assign_short_key
+    self.short_key ||= SecureRandom.urlsafe_base64(6)
```
tarebyte: Looks like `assign_short_key` could be extracted out to a model concern, would you mind doing that?

app/controllers/short_url_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class ShortUrlController < ApplicationController
+  def accept_assignment
```
tarebyte: Can we update these actions to `assignment_invitation` and `group_assignment_invitation`

This action isn't actually accepting anything so it's kind of misleading.

db/schema.rb
nwoodthorpe: Whoops!

nwoodthorpe:
@tarebyte Addressed feedback + rebased and fixed merge conflicts.

tarebyte:
@nwoodthorpe would you mind also removing the `.swp` file?

Thanks!

nwoodthorpe:
Weird, not sure when I vim'd something :o Fixed!

db/migrate/20170525221134_add_short_key_to_assignment_invitation.rb
```diff
@@ -0,0 +1,6 @@
+class AddShortKeyToAssignmentInvitation < ActiveRecord::Migration[5.0]
+  def change
+    add_column :assignment_invitations, :short_key, :string
```
tarebyte: Can we add indexes on to these?

tarebyte: NVM I did it for you in 07c1695

tarebyte:
@nwoodthorpe in 28e1d83 I updated the code so that we show the short url if it's available instead of showing two different urls.

Let me know what you think!

nwoodthorpe:
Lgtm :+1:, tested it as well and looks like it works fine.

PR #902, opened 2017-06-06
Use form tooltips to show errors instead of flash messages

nwoodthorpe:
@tarebyte as we discussed yesterday, this changes our assignment form option errors from a flash message to Primer tooltips. 

We need to add `errored` to the form class + add the `dd` on error, so the code is a little messy. It's not too bad though.

I'm not 100% on leaving the errors up even after the user starts typing. It's kind of backwards to what most sites do, and it hides our hints underneath the fields. The only way the user can view the hint is to CMD-R.

We could make the errors disappear with a mutation observer (since ready selectors seem to be broken right now) that listens for `$('.form.errored input').onChange` that just removes the errored class from the field. I'm also OK leaving it like this though.


app/views/assignments/_assignment_form_options.html.erb
```diff
@@ -1,11 +1,19 @@
-<dl class="form">
+<% title_errors = @assignment.errors.full_messages_for(:title).join(", ") %>
```
tarebyte: @nwoodthorpe I'd love to extract this logic into a view model, would you mind taking a look into that?

nwoodthorpe: Never used a view model in rails before. Is this kind of what you were thinking?

nwoodthorpe:
@tarebyte  Never used a view model in rails before. Is this kind of what you were thinking?

nwoodthorpe:
Merged and finished up your work @tarebyte, thanks for the feedback, looks a lot better than before

tarebyte:
screen shot 2017-06-07 at 13 49 21

@nwoodthorpe mind doing one last thing? Would you mind changing the tooltip from all red to the one I have screen shotted above?

We're on an older (much older) version of primer that doesn't have this yet. The source for that tooltip is here: https://github.com/primer/primer-css/blob/v2.7.0/scss/_forms.scss#L284-L303

nwoodthorpe:
Done. For reference, the new source is [here](https://github.com/primer/primer-forms/blob/master/lib/form-group.scss), and I grabbed colors from [color-system.scss](https://github.com/primer/primer-support/blob/master/lib/variables/color-system.scss) and [colors.scss](https://github.com/primer/primer-support/blob/master/lib/variables/colors.scss).

I added an extra class `primer-new` to the error form to beat the existing primer styles' specificity. Otherwise we just get overridden. 

screen shot 2017-06-07 at 11 20 45 am

nwoodthorpe:
We were missing a few styles, looks like this now:

screen shot 2017-06-07 at 11 31 46 am

app/assets/stylesheets/layout.scss
```diff
@@ -895,3 +895,23 @@ hr {
 .flash.flash-error p:first-of-type {
   margin-top: 0;
 }
+
+dl.form.errored.primer-new {
```
tarebyte: @nwoodthorpe Could we just override the original classes?

That way when we look at the documentation we don't have to go back and remember that we made a new class?

Could we also move this into the [`forms.scss`](https://github.com/education/classroom/blob/1c9b7e1f23d2f8b4a010d1cb820749b1a14c2381/app/assets/stylesheets/components/forms.scss) file?

nwoodthorpe: Without the new class, our stye get's overridden by Primer. In CSS, if two selectors have equal specificity, the latest one gets used, and I guess application.scss is applied before imports. Adding a new class increases our specificity.

http://www.htmldog.com/guides/css/intermediate/specificity/

I've moved the stuff to `forms.scss`

tarebyte:
@nwoodthorpe could I talk you into updating the forms for editing an assignment and group assignment as well?

nwoodthorpe:
Thought it worked because they share the assignment_form_options / group_assignment_form_options partial, but the edit page was showing flash errors + our new tooltips. Removed the flash errors, all good. now :+1:

PR #903, opened 2017-06-06
Use require_dependency instead of require

tarebyte:
I've been getting weird loading errors and I couldn't figure out why. And then I realized I used the `require` instead of `require_dependency`...


nwoodthorpe:
:+1:

PR #905, opened 2017-06-07
Feedback for error restyling

tarebyte:
@nwoodthorpe instead of leaving a bunch of comments I'm making a PR to give you some idea about how I'd like the ViewModel to work.

I'd like it to be something more generic like I've described below. It'd be nice if it could just take in one object and respond to the field rather than making a bunch of objects in a view.

Let me know what you think!


nwoodthorpe:
Ah, this makes way more sense. Got it :+1:

I'm going to merge this then work on the other branch, I don't think I can add commits to your fork.

PR #906, opened 2017-06-07
Add functionality to watch AssignmentRepo and GroupAssignmentRepo

### LINKED ISSUE #775: Automatically add students to watch list on their repositories
When a student creates a repository from an assignment, they do not appear to be added to the watch list for their new repository.  I, as the actual "creator" of the repository, am automatically watching it, but it would make a lot of sense for the student to watch automatically as well.  If I understand correctly, this would mean they would be notified of comments I make on their commits, for example, which would be very useful.

BelieveC:
After accepting invitation user would be automatically added watchlist of created repo

Closes #775

@nwoodthorpe @tarebyte 

Will fix failing tests by recording cassettes once code to watch repos is approved.

### DIFF
--- a/app/controllers/group_assignment_invitations_controller.rb
+++ b/app/controllers/group_assignment_invitations_controller.rb
@@ -67,6 +67,8 @@ class GroupAssignmentInvitationsController < ApplicationController
     users_group_assignment_repo = invitation.redeem_for(current_user, selected_group, new_group_title)
 
     if users_group_assignment_repo.present?
+      options = { subscribed: true }
+      current_user.github_client.update_subscription(users_group_assignment_repo.github_repository.full_name, options)
       yield if block_given?
     else
       flash[:error] = 'An error has occurred, please refresh the page and try again.'
--- a/app/controllers/group_assignment_invitations_controller.rb
+++ b/app/controllers/group_assignment_invitations_controller.rb
@@ -67,8 +67,7 @@ class GroupAssignmentInvitationsController < ApplicationController
     users_group_assignment_repo = invitation.redeem_for(current_user, selected_group, new_group_title)
 
     if users_group_assignment_repo.present?
-      options = { subscribed: true }
-      current_user.github_client.update_subscription(users_group_assignment_repo.github_repository.full_name, options)
+      current_user.github_user.watch_repository(users_group_assignment_repo)
       yield if block_given?
     else
       flash[:error] = 'An error has occurred, please refresh the page and try again.'
--- a/app/models/assignment_repo/creator.rb
+++ b/app/models/assignment_repo/creator.rb
@@ -91,7 +91,6 @@ class AssignmentRepo
     # as a collaborator.
     #
     # Returns true if successful, otherwise raises a Result::Error
-
     def add_user_to_repository!(github_repository_id)
       options = {}.tap { |opt| opt[:permission] = 'admin' if assignment.students_are_repo_admins? }
 
--- a/app/models/assignment_repo/creator.rb
+++ b/app/models/assignment_repo/creator.rb
@@ -69,6 +69,8 @@ class AssignmentRepo
         push_starter_code!(assignment_repo.github_repo_id)
       end
 
+      add_user_to_repo_watchlist(assignment_repo)
+
       begin
         assignment_repo.save!
       rescue ActiveRecord::RecordInvalid
@@ -85,10 +87,16 @@ class AssignmentRepo
 
     private
 
+    def add_user_to_repo_watchlist(repo)
+      options = { subscribed: true }
+      user.github_client.update_subscription(repo.github_repository.full_name, options)
+    end
+
     # Internal: Add the User to the GitHub repository
     # as a collaborator.
     #
     # Returns true if successful, otherwise raises a Result::Error
+
     def add_user_to_repository!(github_repository_id)
       options = {}.tap { |opt| opt[:permission] = 'admin' if assignment.students_are_repo_admins? }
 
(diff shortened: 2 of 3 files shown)


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -67,6 +67,8 @@ def create_group_assignment_repo(selected_group: group, new_group_title: nil)
     users_group_assignment_repo = invitation.redeem_for(current_user, selected_group, new_group_title)
 
     if users_group_assignment_repo.present?
+      options = { subscribed: true }
+      current_user.github_client.update_subscription(users_group_assignment_repo.github_repository.full_name, options)
```
nwoodthorpe: So for actions on GitHub resources, we have a bunch of classes that inherit from `GitHubResource` where we put this logic. We're subscribing a user to a repository, and we need to be sure we're using the user's token, so I think this logic belongs in `GitHubUser`. `GitHubUser#accept_repository_invitation` is a good example of how this method could look.

app/models/assignment_repo/creator.rb
```diff
@@ -85,10 +87,16 @@ def perform
 
     private
 
+    def add_user_to_repo_watchlist(repo)
```
nwoodthorpe: Same comment as above.

nwoodthorpe:
I like the looks of this, 1 main comment though. Thanks for taking this on!

BelieveC:
@nwoodthorpe I have made the changes :+1:

app/models/github_user.rb
```diff
@@ -11,6 +11,11 @@ def accept_repository_invitation(invitation_id, **options)
     end
   end
 
+  def watch_repository(repo)
+    options = { subscribed: true }
+    @client.update_subscription(repo.github_repository.full_name, options)
```
nwoodthorpe: I think writing this as `@client.update_subscription(repo.github_repository.full_name, subscribed: true)` is a little neater

app/models/assignment_repo/creator.rb
```diff
@@ -89,6 +91,7 @@ def perform
     # as a collaborator.
     #
     # Returns true if successful, otherwise raises a Result::Error
+
```
nwoodthorpe: Looks like this slipped in! Could you remove it?

app/models/github_user.rb
nwoodthorpe: Methods in Octokit can throw a ton of different errors from the GitHub API, so in our GitHubResource classes, we usually wrap client calls. in `GitHub::Errors.with_error_handling do ... end`. The error handling manages all the different errors and raises `GitHub::Error` if something comes up. 

We should catch the error here and probably rescue and ignore the error, since it doesn't actually matter very much if we fail to follow the repo.

nwoodthorpe:
Couple of comments, besides that I :+1: this :)

app/models/github_user.rb
BelieveC: I will make the changes :+1:

tarebyte:
Hi @BelieveC thanks for opening this PR!

I'm going to be out of office until Monday, but when I get back I promise I'll give this some 👀

BelieveC:
Alright :smile:

BelieveC:
@nwoodthorpe
 I'm trying to record new cassettes with `record: :new_episodes`. It is kinda weird it throws invalid credentials error each time it tries to record new cassettes. 
I have followed https://github.com/education/classroom#testing-environment-variables to setup .env for testing and running new cassettes. Is anything else I need to setup?? :thinking:

nwoodthorpe:
Instead of using record: new_episodes, just delete the cassette for the specs you need to re-record. A spec with no cassette will record a new cassette.

If you're having authentication errors, you might be running into a bug I'm fixing in #907. If your personal access token has too many scopes, it'll fail to authenticate. 

Can you set your teacher scopes to "user:email repo delete_repo admin:org admin:org_hook" (and nothing else), and your student scopes to "user:email" and nothing else, and try again? You'll need to delete the old cassette each time you re-try the spec or else it won't actually make any requests.

BelieveC:
@nwoodthorpe first of all sorry for delay, was on vacation. I tried what you suggested. But it didn't work.

nwoodthorpe:
Hm strange. Can you push your failing specs anyways? I'll :eyes: when I get a bit of time.

BelieveC:
@nwoodthorpe I have fixed few of them but `group_assignment_invitations_controller_spec` tests still persist. After debugging, the main issue that persists is when test tries to subscribe user to created group_assignment_repo, group_assignment_repo is already deleted and have full_name as `deleted_repository`. So it throws `Invalid Repository. Use user/repo format`. Do you have any idea about why this is happening? Because repository created shows correct full name in just before saving repository but after saving it shows itself as deleted_repository.

nwoodthorpe:
@BelieveC What's the name or line # of the spec that's failing?

BelieveC:
@nwoodthorpe 

Few examples are failing due same reason as above. You can check this example

`spec/controllers/group_assignment_invitations_controller_spec.rb:68`

nwoodthorpe:
@BelieveC It looks like the GitHubRepository that the GroupAssignmentRepo is given is the `NullGitHubRepository`. The definition of that class says full_name is "Deleted Repository". We'll probably need to look into why .github_repository is giving the NullGitHubRepository instead of an actual GitHubRepository that's being vcr'd.

mine-cetinkaya-rundel:
Is this actually functional? It doesn't seem to be working for my course/organization, and I think it's a very useful feature.

PR #910, opened 2017-06-12
Model changes for Deadlines PR

nwoodthorpe:
All the model changes for the deadlines PR, + specs and the Deadline Job


app/models/deadline.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def passed?
+    deadline_at.past?
+  end
+
+  private
+
+  def deadline_in_future
+    errors.add(:deadline_at, 'must be in the future') if deadline_at && deadline_at < Time.zone.now
```
tarebyte: Can we make this a guard clause?

```ruby
def deadline_in_future
  return unless deadline_at && deadline_at < Time.zone.now
  errors.add(:deadline_at, 'must be in the future')
end
```

app/models/deadline/factory.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class Deadline
+  class Factory
+    # Accepts an optional datetime format string :datetime_format
+    # Default DateTime format is %m/%d/%Y %H:%M %z
+    # e.g. 05/25/2017 13:17-0800
+    def self.build_from_string(opts = {})
+      deadline = Deadline.new
+
+      format = opts[:datetime_format] || '%m/%d/%Y %H:%M %z'
```
tarebyte: Let's freeze this string and make it a constant.

```ruby
DATETIME_FORMAT = '%m/%d/%Y %H:%M %z'.freeze
```

nwoodthorpe:
@tarebyte done!

Rubocop yelled at me when I tried to freeze the string, probably because single quoted strings are frozen by default.

PR #881, opened 2017-05-30
Introduce Assignment deadlines 

nwoodthorpe:
~~[WIP - Not everything discussed below has been implemented yet]~~

This PR introduces a V1 assignment deadline functionality for Classroom. The initial implementation will be fairly minimal and lightweight, we plan on building on top of this in the future.

When creating an assignment, teachers will be able to select a deadline (date/time). At this time, Classroom will fetch the latest commit SHA from each assignment repo and save that as the "submission SHA". Teachers will be able to view commits that happened before and after the deadline in a clear and simple way.

Teachers will also be able to modify or remove the deadline via assignment settings. 

One thing to note: if a student force-pushes over their submission commit, all commits will be marked as being after the deadline. In the future, we may explore solutions to that problem such as protecting the default branch, or removing access from the student.

We'll also be exploring better submission evaluation methods after this PR goes through.

I'll upload some screenshots once I've done more of the front-end stuff.

If anyone has feedback on how this is being implemented, let us know :)

Relevant people:

@johndbritton @tarebyte @mozzadrella @sanicki @iamgoodbytes @ppannuto @samwhitlock @arjenpdevries @dhowe @M4urice

PR #920, opened 2017-06-14
Update repo/show UI to fit more information

nwoodthorpe:
This change gives us way more flexibility when it comes to adding stuff to this partial. The visuals for deadlines(#881) and commit statuses(#708) can both be easily accommodated by this new design.

Design taken from github.com/user/repositories, e.g. [here](https://github.com/nwoodthorpe?tab=repositories)

TODO:
- [x] Group Assignment repo/show redesign
- [x] Group Assignment view model
- [x] Assignment repo/show redesign
- [x] Assignment repo view model (hopefully it can be the same as group_assignment)


nwoodthorpe:
All done

@jasonlong would you be able to :eyes: and let us know what you think? We have a couple other planned properties for repos (commits, members, + more), but for now, this is what we have.

**Before**
**Assignment**
screen shot 2017-06-15 at 10 54 22 am

**Group Assignment**
screen shot 2017-06-15 at 10 54 09 am


**After**
**Assignment**
screen shot 2017-06-15 at 10 53 48 am

**Group Assignment**
screen shot 2017-06-15 at 10 52 55 am

jasonlong:
Hey @nwoodthorpe! I took a look at your screenshots (and thanks for including the before versions too) and here are some thoughts:

### Buttons

Primary (green) buttons ideally aren't used more than once on a page. Those "View repository" buttons should probably be gray.

Also, the button labels should use sentence case – "View repository" and not "View Repository"

### Assignment page

I think we can tweak the UI a bit to help with the imbalance of the big avatars and the new metadata line. A better visual pattern might be what we have on the [Followers page](https://github.com/nwoodthorpe?tab=followers):

screen shot 2017-06-15 at 4 18 52 pm

You could combine the username and real name in the top line and then have the metadata line below it so that both lines sit next to the avatar. Like

screen shot 2017-06-15 at 4 20 58 pm

but with these meta fields.

### Group assignment page

Likewise, there's existing GitHub UI patterns for listings of groups of people that we could pull from. The Teams overview page under an org might be a good example. A quick mock of using something like that:

screen shot 2017-06-15 at 4 35 05 pm

A nice little touch might be to show the `user` icon when there's just one member and the `organization` one (like above) when there's more.

Hope that helps!

nwoodthorpe:
Thanks @jasonlong 

Your idea's look great, I like them a lot.

Is this sort of what you were thinking?

screen shot 2017-06-15 at 4 11 12 pm

screen shot 2017-06-15 at 4 11 35 pm

jasonlong:
Yeah! I think that's looking pretty sharp. :zap:

jasonlong:
With the "View repository" buttons being gray, there's a lot of competing similar looking buttons (and maybe that's why you went with green), so another option might be to make them [outline buttons](http://primercss.io/buttons/#outline).

nwoodthorpe:
I'll play around with those tmrw morning, and tidy up the SASS. :+1:

nwoodthorpe:
I like it 🙌 


screen shot 2017-06-16 at 8 47 20 am

nwoodthorpe:
This is now all good to go

The assignment repo view is in the comment above.

This is the group assignment repo view:

screen shot 2017-06-16 at 10 19 44 am

nwoodthorpe:
@tarebyte Rebased on master and merged the deadlines changes in. Now we have another repo_detail for submissions. 

screen shot 2017-06-27 at 8 10 43 am

Moving forward from here, we can start implementing UX improvements based on Vanessa's work in #959

nwoodthorpe:
Now the full thing is a link, not just the icon. This should help with making sure users know these partials are clickable, like on GitHub.

(My screen recorder doesn't show the cursor changing to the click cursor, it actually does)

http://g.recordit.co/S3B2fZ9w4I.gif

app/assets/stylesheets/layout.scss
```diff
@@ -320,86 +338,49 @@ hr {
...
-  .assignment-repo-list-item:not(:first-child) {
-    padding-top: 1rem;
-    margin-top: 1rem;
-    border-top: 1px solid #e3e3e3;
-
-    @media screen and (max-width: 40em) {
-      margin-bottom: 0;
-    }
-  }
+.assignment-repo-info-container {
```
tarebyte: Instead of having our own custom container can we use the new stuff that Primer 6 gives us?

app/view_models/assignment_repo_view.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
tarebyte: Let's start organizing these a little more what do you think about a path like

`/app/view_models/assignment_repo/show_view.rb`

```ruby
# frozen_string_literal: true

class AssignmentRepo::ShowView < SharedAssignmentRepoView
  # ...
end
```

app/view_models/assignment_repo_view.rb
```diff
@@ -0,0 +1,19 @@
...
+class AssignmentRepoView < SharedAssignmentRepoView
+  attr_reader :assignment_repo
+
+  delegate :github_user, to: :assignment_repo
+
+  def avatar_url
+    avatar_url_for(github_user, 96)
+  end
+
+  def user_login
```
tarebyte: What do you think about updating this to use `github_login` instead of `user_login`

tarebyte: And all other places where a `github_` prefix would make sense 😄

nwoodthorpe:
@tarebyte Done!

I renamed `AssignmentRepo::ShowView` to `AssignmentRepoView::ShowView` because the model `AssignmentRepo` was causing name conflicts. I'm welcome to better name suggestions if you have one hehe

PR #928, opened 2017-06-19
Create GraphqlAssignmentsController for doing graphql work in

nwoodthorpe:
This PR makes a GraphqlAssignmentsController which is a copy of AssignmentsController, except it renders views from AssignmentsController and validates that the user is staff + has the flipper enabled.

From here, I'll be working on moving API requests from REST to GraphQL for this new controller only.

@tarebyte @johndbritton


nwoodthorpe:
Going to close this for now. Will re-open if we decide to move forward with this in the future.

PR #939, opened 2017-06-22
Allow staff to add users to feature preview group

tarebyte:
This is one implementation for allowing us to keep a group of "feature previewers" that we can add when we're ready for beta testing.

## Demo in stafftools 
enabling feature

## Todo
- [ ] Feedback
- [ ] Controller tests
- [ ] Extract migration into another PR to ship and run

/cc @mozzadrella as you may want this to look differently

### DIFF
--- a/config/initializers/flipper.rb
+++ b/config/initializers/flipper.rb
@@ -18,7 +18,5 @@ module GitHubClassroom
   end
 end
 
-# Flipper group for staff
-Flipper.register(:staff) do |user|
-  user.respond_to?(:staff?) && user.staff?
-end
+Flipper.register(:staff)              { |user| user.respond_to?(:staff?) && user.staff?                         }
+Flipper.register(:feature_previewers) { |user| user.respond_to?(:feature_previewer?) && user.feature_previewer? }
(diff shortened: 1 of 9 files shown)


tarebyte:
@wilfriedE I also took the time to update our use of `boxed-group` to use the new `Box` pattern in Primer 6.

We may want to look at updating the other instances, but it's _not_ a priority.

config/initializers/flipper.rb
```diff
@@ -18,7 +18,5 @@ def self.flipper!
   end
 end
 
-# Flipper group for staff
-Flipper.register(:staff) do |user|
-  user.respond_to?(:staff?) && user.staff?
-end
+Flipper.register(:staff)              { |user| user.respond_to?(:staff?) && user.staff?                         }
```
nwoodthorpe: Can we change these to use the `&.` operator?

nwoodthorpe:
This all lgtm so far :+1:

It would be nice to have a view of all features currently enabled for preview users. RIght now, I think the only way to do that is to go through each user in the Flipper menu and check the groups it's been added to.

PR #944, opened 2017-06-22
Revert "Private repos"

tarebyte:
Reverts education/classroom#924


tarebyte:
@BenEmdon I'm pretty sure it had something to do with errors that were being raised when this went out.

I wish I would've added more context 😢

PR #963, opened 2017-06-24
Add I18n support for autocomplete and videos

BelieveC:
Convert static text to symbols for autocomplete and videos views


BelieveC:
@wilfriedE, Can you merge it? or else stale Bot will close. You have already approved the PR.
@tarebyte

wilfriedE:
@BelieveC no, I can't merge this :sob:, but try keeping an eye on this for now and keep it open.

config/locales/views/autocomplete/en.yml
```diff
@@ -0,0 +1,4 @@
+en:
+  views:
+    autocomplete:
+      no_matching_repo: "We couldn't find any matching repositories"
```
d12: Can you change this to use single quotes for consistency?

d12:
@BelieveC There's one small change in the review above, then I can merge this. I can't commit to the PR because it's on your fork.

BelieveC:
@d12, Sorry I was away due to Holidays. I will get it done today :+1:

BelieveC:
@d12 , I have made the requested change :+1:

config/locales/views/autocomplete/en.yml
```diff
@@ -1,4 +1,4 @@
 en:
   views:
     autocomplete:
-      no_matching_repo: "We couldn't find any matching repositories"
+      no_matching_repo: 'We couldn''t find any matching repositories'
```
d12: @BelieveC looks like you added a `'` in `couldn''t`. Could you remove that? 😅

BelieveC: @d12 , Actually one extra `'` is required as an escape symbol. Therefore I have added one extra `'`.

d12: Ohh I see what's happening here. Correct me if I'm wrong, but I think the behavior of `''` in a string is concatenation, since all you're doing is terminating one string and beginning another, and `"a" "b"` in ruby is equal to `"ab"`

```
irb(main):002:0> 'foo''bar'
=> "foobar"
```

If that's the case, `\'` should work.

BelieveC: > I think the behavior of '' in a string is concatenation, since all you're doing is terminating one string and beginning another, and "a" "b" in ruby is equal to "ab"

Exactly! This is what I have done there.

> If that's the case, \' should work.

I tried \, but it works within " ". Here it didn't work as we have ' ' .

d12: `\'` should work within ''. You need to escape it with a backslash. 

The way you've done it actually loses the `'`.

```
irb(main):004:0> 'We couldn''t find any matching repositories'
=> "We couldnt find any matching repositories"
```

Notice the missing `'` in the string.

BelieveC: My Bad! Wait will fix it :+1:

BelieveC: @d12 , Actually `'couldn''t'` is parsed as `couldn't` in YAML. I checked it out here http://yaml-online-parser.appspot.com/

BelieveC: Lemme know if any other changes are required.

d12: Whoops, I forgot this is YAML and not ruby. Also, TIL `''` => `'` in YAML.

Sorry about the hassle!

PR #967, opened 2017-06-26
Memoize disabled? to save a ton of API requests

nwoodthorpe:
While doing some profiling last week, I noticed that our calls to `disabled?` are causing a TON of API requests on `group_assignment/show` and `assignment/show`, coming from the `group_assignment_repo/show` and `assignment_repo/show` partials.

On an assignment repo, we make 3 `disabled?` calls, which equals 6 API requests.

On a group assignment repo, we make between 3 and 8 `disabled?` calls depending on the number of team members, which equals 6 - 16 API requests.

The value of `disabled?` shouldn't change during the time we're rendering these partials, so memoizing won't have any visible impact. It will, though, bring both of these partials down to only 2 API requests.

@tarebyte @johndbritton

### DIFF
--- a/app/models/assignment_repo.rb
+++ b/app/models/assignment_repo.rb
@@ -24,7 +24,7 @@ class AssignmentRepo < ApplicationRecord
   # This should really be in a view model
   # but it'll live here for now.
   def disabled?
-    !github_repository.on_github? || !github_user.on_github?
+    @disabled ||= !github_repository.on_github? || !github_user.on_github?
   end
 
   def private?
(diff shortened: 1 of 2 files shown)


tarebyte:
@nwoodthorpe any chance you want to fix my TODO and move these to `ViewModel` logic while we're here?

nwoodthorpe:
@tarebyte We don't have view models for these two views yet. We'll probably want to extract more stuff than just this, should we do that in a seperate PR?

tarebyte:
> We don't have view models for these two views yet. We'll probably want to extract more stuff than just this, should we do that in a seperate PR?

Yeah it's a bit out of scope for this PR

johndbritton:
> I noticed that our calls to disabled? are causing a TON of API requests

Nice spotting ✨

PR #976, opened 2017-06-29
Fix deadline_updated? method of assignment::editor

BelieveC:
`deadline_updated?` method returns `true` even for unchanged deadline datetime leading to creation of new deadline record for every assignment update request.

### DIFF
--- a/app/models/assignment/editor.rb
+++ b/app/models/assignment/editor.rb
@@ -92,7 +92,7 @@ class Assignment
 
     def deadline_updated?
       return false unless @options[:deadline]
-      new_deadline_at = DateTime.strptime(@options[:deadline], '%m/%d/%Y %H:%M %z').utc
+      new_deadline_at = DateTime.strptime(@options[:deadline], DATETIME_FORMAT).utc
       new_deadline_at != @assignment.deadline&.deadline_at
     end
   end
--- a/app/models/assignment/editor.rb
+++ b/app/models/assignment/editor.rb
@@ -91,6 +91,7 @@ class Assignment
     end
 
     def deadline_updated?
+      return false unless @options[:deadline]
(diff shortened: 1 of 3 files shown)


app/models/assignment/editor.rb
```diff
@@ -91,7 +91,9 @@ def update_attribute_for_all_assignment_repos(attribute:, change:)
     end
 
     def deadline_updated?
-      @options[:deadline] != @assignment.deadline&.deadline_at
+      return false unless @options[:deadline]
+      new_deadline_at = DateTime.strptime(@options[:deadline], '%m/%d/%Y %H:%M %z').utc
```
nwoodthorpe: We also use this format string in Deadline::Factory. Would you be able to refactor them both out to a shared constant somewhere incase we ever change this format?

BelieveC: Sure thing :+1:

BelieveC:
@nwoodthorpe I have made the changes :+1:

app/models/assignment/editor.rb
```diff
@@ -91,7 +91,9 @@ def update_attribute_for_all_assignment_repos(attribute:, change:)
     end
 
     def deadline_updated?
-      @options[:deadline] != @assignment.deadline&.deadline_at
+      return false unless @options[:deadline]
+      new_deadline_at = DateTime.strptime(@options[:deadline], DATETIME_FORMAT).utc
```
tarebyte: I'd rather have this look something like

```ruby
new_deadline_at = DateTime.strptime(@options[:deadline], Deadline::Factory::DATETIME_FORMAT).utc
```

Instead of a constants file just hanging out as an initializer

BelieveC: Alright! I will make the changes :+1:

BelieveC:
@tarebyte I have made the changes :+1:

PR #977, opened 2017-06-29
Fix datetimepicker.js bug

BelieveC:
Due to incorrect format of deadline field in assignment update page, Incorrect datetime was selected as default. Correct datetime format to render correct datetime in selector popup.

Before: 
before

After: 
after


nwoodthorpe:
I'm not seeing this on my end. Do you have reproduction steps? What is your browser/OS? 
screen shot 2017-06-29 at 12 39 19 pm

BelieveC:
@nwoodthorpe 

> What is your browser/OS

Chromium 58.0.3029.110 Built on Ubuntu , running on Ubuntu 16.04

PR #978, opened 2017-06-29
Migrations to remove identifier stuff

nwoodthorpe:
A migration to remove the rows and tables in the schema related to the old implementation of identifiers.

The diff looks gnarly right now because this is based on #975, once that's merged, this'll look cleaner. The changes are in the latest commit.


tarebyte:
Migration has been run!

PR #979, opened 2017-06-30
Add missing association with group_assignment_repo for group model

BelieveC:
`has_many` association missing between `group` and `group_assignment_repos`


nwoodthorpe:
Why did we need to modify some of the cassettes?

BelieveC:
@tarebyte 

> Was there a use case you have in mind?

Yes! We do require this association in the next PR #980. When we delete a group we need to delete group_assignment_repos associated with it to avoid breaking of many pages associated with group_assignment_repos. For an example take this case

Delete a Group A -> Render Group Assignment Repo with A Group -> Link to group A breaks as no such group exists. 

Also In group_assignment_repo model, we have already declared one side of relation [link](https://github.com/education/classroom/blob/master/app/models/group_assignment_repo.rb#L10).

BelieveC:
@nwoodthorpe 

> Why did we need to modify some of the cassettes?

Few test in group_assignment_repo_spec were failing due to changes in associated group model.

Issue #970, opened 2017-06-28
Initialize assignment issues from private branch


wilfriedE:
Generate issues for Assignment repo if starter code has a `github-classroom` branch with issues. 

As part of Milestone 1 in #937.

PR #983, opened 2017-07-03
Initialize Assignment Issues

wilfriedE:
Work in progress for resolving #970 . 

Adds repository setup process into assignment repo creation flow.
Creates a set of issues on the student's assignment repo from the given starter code if it has a `github-classroom` branch with an issues folder with preloaded  issues. 
Removes `github-classroom` branch from student repo after successful repository setup and gives access to the student.


nwoodthorpe:
Ping me when this is :+1:!

app/models/github_blob.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require 'safe_yaml/load'
+
+class GitHubBlob
+  YAML_FRONT_MATTER_REGEXP = %r!\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)!m
```
tarebyte: It would be ✨ if you linked the source of this regex for context.

tarebyte:
> Ping me when this is 👍!

Better yet just add us a reviewers to the PR when you're ready 😄 unless you have specific questions that need out attention.

app/models/github_blob.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'safe_yaml/load'
+
+class GitHubBlob
+  # From jekyll/jekyll
+  # https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L13
```
tarebyte: Fun fact if go to a blob view and hit `y` you can see the `sha` for that file

```
https://github.com/jekyll/jekyll/blob/74373baa550282a8630368e7b609ca9370f6d560/lib/jekyll/document.rb#L13
```

app/models/assignment_repo/creator.rb
```diff
@@ -135,6 +135,7 @@ def push_starter_code!(github_repo_id)
       starter_code_repository = GitHubRepository.new(client, starter_code_repo_id)
 
       assignment_repository.get_starter_code_from(starter_code_repository)
+      setup_repository(starter_code_repository, assignment_repository)
```
nwoodthorpe: Two comments:
I haven't read past here in the diff, but `setup_repository` doesn't really say much about what exactly this method does.

Also I'm not sure this belongs in `push_starter_code`

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
       suffix = "-#{suffix_count}"
       repository_name.truncate(100 - suffix.length, omission: '') + suffix
     end
+
+    # Handles assignment repository setup
```
nwoodthorpe: I think in tomdoc you need to specify whether this is public or internal

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+    #
+    # configs_repo    - GitHubRepository containing the configuration files
+    # assignment_repo - GitHubRepository for which to perform the configuration
+    #                   setups
+    #
+    # Returns nothing
+    def setup_repository(configs_repo, assignment_repo)
+      return unless configs_repo.classroom_configs?
+      conf_tree = configs_repo.branch('github-classroom').commit.commit.tree.sha
```
nwoodthorpe: This is reaching pretty deep into the returned object from the API. Maybe we can make a method on `GitHubRepository` that handles this API-specific stuff for us.

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+      configs_repo.tree(conf_tree).tree.each do |conf|
```
nwoodthorpe: Could we assign this to a variable with a descriptive name? Iterating over the tree of the conf_tree of the conf_repo is pretty confusing

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+        process_config(configs_repo, assignment_repo, conf)
```
nwoodthorpe: Could we rename this method to say what sort of `processing` this method does?

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+      end
+
+      wait_import_completion assignment_repo
+
+      assignment_repo.remove_branch('github-classroom')
+    end
+
+    def wait_import_completion(github_repository)
```
nwoodthorpe: Is there no webhook for this? Or some way we can do this without tying up a web process? We don't have many web processes on Heroku.

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+      t = 1
+      loop do
+        break if github_repository.import_progress[:status] == 'complete' || t > 60
```
nwoodthorpe: `|| t > 60` makes it seem like this times out in 60 seconds, but that isn't actually what's happening here. If we do end up going with this sleep logic, can we make the timeout more clear?

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+        sleep t += (t *= 0.25).ceil
+      end
+    end
+
+    # Process the configurations separetly
+    #
+    # configs_repo    - GitHubRepository containing the configuration files
+    # assignment_repo - GitHubRepository for which to perform the configuration
+    #                   setups
+    # conf            - GitObjects from of the 'github-classroom' branch tree
```
nwoodthorpe: Can we make this more descriptive?

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,63 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+    #                   setups
+    # issues_tree     - sha of the 'issues' tree
+    #
+    # Returns nothing
+    def gen_issues(configs_repo, assignment_repo, issues_tree)
+      configs_repo.tree(issues_tree).tree.each do |issue|
+        blob = configs_repo.blob(issue.sha)
+        assignment_repo.add_issue(blob.data['title'], blob.body)
+      end
+    end
```
nwoodthorpe: There's a ton of logic in this file, I wonder if maybe this all belongs together in some sort of lib object or service object. I'm not sure though, still reading the code.

app/models/github_blob.rb
```diff
@@ -0,0 +1,49 @@
...
+  # From jekyll/jekyll
+  # https://github.com/jekyll/jekyll/blob/74373baa550282a8630368e7b609ca9370f6d560/lib/jekyll/document.rb#L13
+  YAML_FRONT_MATTER_REGEXP = /\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)/m
+
+  attr_reader :data, :body
+
+  def initialize(github_repository, sha, **options)
+    repo = github_repository
+    @blob = repo.client.blob(repo.full_name, sha, options)
+    @data = nil
```
nwoodthorpe: This and the line below aren't really necessary since we defined the attr_reader.

app/models/github_blob.rb
```diff
@@ -0,0 +1,49 @@
...
+    Base64.decode64(@blob.content)
+  end
+
+  def read_contents
+    match = YAML_FRONT_MATTER_REGEXP.match(decoded_content)
+    if match
+      @body = match.post_match
+      @data = SafeYAML.load(match.to_s)
+    end
+    @data
```
nwoodthorpe: Does this do anything? The only call I see is in `#initialize` and I don't think you can return values from an initializer.

app/models/github_blob.rb
```diff
@@ -0,0 +1,49 @@
...
+  #
+  # Returns a base64 string
+  def content
+    @blob.content
+  end
+
+  private
+
+  def decoded_content
+    return @blob.content unless @blob.content != 'utf-8'
```
nwoodthorpe: Can we use `if` here

app/models/github_repository.rb
```diff
@@ -22,6 +22,12 @@ def get_starter_code_from(source)
     end
   end
 
+  def import_progress(**options)
```
nwoodthorpe: This is awesome, super happy we can do this through the api 🙌 

It's still an API call each time though, so we should try to call as few times as we can.

app/models/github_repository.rb
```diff
@@ -34,6 +40,44 @@ def invite(user, **options)
...
+  # title    - The title of the issue.
+  # body     - The body of the issue.
+  #
+  # Returns the newly created issue, or raises a GitHub::Error.
+  def add_issue(title, body, **options)
+    GitHub::Errors.with_error_handling do
+      @client.create_issue(full_name, title, body, options)
+    end
+  rescue GitHub::Error
+    []
```
nwoodthorpe: Why the `[]` return value on error?

app/models/github_repository.rb
```diff
@@ -34,6 +40,44 @@ def invite(user, **options)
...
+  #
+  # sha    - sha of the tree.
+  #
+  # Returns a git Tree object, or raises a GitHub::Error.
+  def tree(sha, **options)
+    GitHub::Errors.with_error_handling do
+      @client.tree(full_name, sha, options)
+    end
+  rescue GitHub::Error
+    []
```
nwoodthorpe: Same comment here as above

app/models/github_repository.rb
```diff
@@ -34,6 +40,44 @@ def invite(user, **options)
...
+  end
+
+  # Get a tree object from the GitHub repository.
+  #
+  # sha    - sha of the tree.
+  #
+  # Returns a git Tree object, or raises a GitHub::Error.
```
nwoodthorpe: This doesn't raise an error, it returns `[]`

app/models/github_repository.rb
```diff
@@ -34,6 +40,44 @@ def invite(user, **options)
     end
   end
 
+  # Add issues to the GitHub repository.
+  #
+  # title    - The title of the issue.
+  # body     - The body of the issue.
+  #
+  # Returns the newly created issue, or raises a GitHub::Error.
```
nwoodthorpe: This doesn't raise an error, it returns `[]`

app/models/github_repository.rb
```diff
@@ -42,6 +86,20 @@ def default_branch
     end
   end
 
+  def branch(name, **options)
+    GitHub::Errors.with_error_handling do
+      @client.branch(full_name, name, options)
+    end
+  rescue GitHub::Error
+    []
```
nwoodthorpe: Same comment as above

app/models/github_repository.rb
```diff
@@ -58,6 +116,12 @@ def present?(**options)
     self.class.present?(@client, @id, options)
   end
 
+  def classroom_configs?(**options)
```
nwoodthorpe: What does this method do? `repository.classroom_configs?` is a little confusing.

nwoodthorpe: Maybe `config_branch_present?` or something like that

app/models/assignment_repo/creator.rb
wilfriedE: setup_repository is in push_starter_code for now since we're focusing on issues creation from starter code on a per assignment basis. @tarebyte  what do you think?

app/models/assignment_repo/creator.rb
wilfriedE: Yeah, I agree. One thing I would like to do is in the next milestone when all of this moves to a module to handle repository setup for both AssignmentRepo and GroupAssignmentRepo we handle this a little more differently.

app/models/assignment_repo/creator.rb
wilfriedE: I was trying to save space 😅

app/models/assignment_repo/creator.rb
wilfriedE: Not at the moment unfortunately. Might be doing something with ActionCable to better handle this in the future.

app/models/github_repository.rb
wilfriedE: yeah! good catch, this should be throwing an error.

app/models/github_repository.rb
wilfriedE: With this I think they could have other types of configuration branches in the future. I do think something like `branch_present?` instead would be a good idea.

app/models/assignment_repo/creator.rb
nwoodthorpe: @tarebyte Do our heroku dynos only process one web request at a time?

If so, with the number of dynos we have, I'm not sure we can be freezing requests for up to 2 minutes. If a class releases an assignment to a roster of 300 people, it'll take a very small number (<10) of imports freezing to freeze the entire app for 2 minutes. On top of that, only a few students will be able to accept an invitation at once because of how slow the importer is.

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,69 @@ def suffixed_repo_name(repository_name, suffix_count)
...
       repository_name.truncate(100 - suffix.length, omission: '') + suffix
     end
+
+    # Internal: Handles assignment repository setup
+    #
+    # configs_repo    - GitHubRepository containing the configuration files
+    # assignment_repo - GitHubRepository for which to perform the configuration
+    #                   setups
+    #
+    # Returns nothing
```
nwoodthorpe: Are there cases where something in here could fail and we would need to do something different?

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,69 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+
+    # Internal: Generates issues for the assignment_repo based on the configs
+    #
+    # configs_repo    - GitHubRepository containing the configuration files
+    # assignment_repo - GitHubRepository for which to perform the configuration
+    #                   setups
+    # tree_sha     - sha of the 'issues' tree
+    #
+    # Returns nothing
+    def generate_issues(configs_repo, assignment_repo, tree_sha)
```
nwoodthorpe: We need some error handling here. `add_issue` raises a GitHub::Error if the API call fails which will bubble all the way up to our ApplicationController and 500 I believe.

app/models/assignment_repo/creator.rb
```diff
@@ -189,5 +190,69 @@ def suffixed_repo_name(repository_name, suffix_count)
...
+    #
+    # github_repository - GitHubRepository awaiting import completion
+    #
+    # Returns nothing
+    def wait_import_completion(github_repository)
+      wait_limit = 60
+      increment_rate = 0.25
+      t = 1
+      loop do
+        break if github_repository.import_progress[:status] == 'complete' || t > wait_limit
```
nwoodthorpe: We'll need error handling here as well. import_progress can return an exception.

app/models/github_repository.rb
```diff
@@ -42,6 +84,29 @@ def default_branch
     end
   end
 
+  def branch(name, **options)
+    GitHub::Errors.with_error_handling do
+      @client.branch(full_name, name, options)
+    end
+  rescue GitHub::Error
+    {}
```
nwoodthorpe: Why a hash? What's the type of the success return value?

nwoodthorpe:
Few more comments. There are also these 2 comments from before that I'm still wondering about. I don't know what the solution is for the timeout thing is without setting up some endpoint for polling or something. Maybe @tarebyte has ideas.
https://github.com/education/classroom/pull/983#discussion_r125766125
https://github.com/education/classroom/pull/983#discussion_r125765277

app/models/github_repository.rb
wilfriedE: It returns a Hash if it's successful. More specifically though it's a Sawyer::Resource

app/models/assignment_repo/creator.rb
tarebyte: @nwoodthorpe I think this is fine to leave in here right now, and when we bring this to group assignment creation we can extract out duplicate logic then.

app/models/assignment_repo/creator.rb
tarebyte: [Article on how Dynos work](https://devcenter.heroku.com/articles/dynos) for reference.

@nwoodthorpe we have two web dynos running atm running puma. Each of those puma servers are running 4 different threads [`RAILS_MAX_THREADS`](https://github.com/education/classroom/blob/master/config/puma.rb#L9)

So we _should_ be ok hanging for a bit, I don't know about 2 minutes though.

wilfriedE: @tarebyte any recommendations for the wait_limit? Changing it to 30 gives us up to ~ 115 seconds worst case.

nwoodthorpe: I still think it should be more clear what the worst case is. `wait_limit=30` doesn't imply a worst case of 115 seconds without tracing through the implementation

wilfriedE: I see what you mean. I think I have a way of addressing this.

wilfriedE: @nwoodthorpe what do you think about the changes.

nwoodthorpe: How about:

```
def wait_import_completion(github_repository)
  timeout = 30.seconds
  begin_time = Time.zone.now

  while (Time.zone.now - begin_time) < timeout do
    break if github_repository.import_progress[:status] == 'complete'

    sleep 3
  end
end
```

wilfriedE: The reason for the incremental approach was to give it more time when checking for completion. I don't see anything wrong with this approach either, but let's have a discussion about this. /cc @nwoodthorpe  @tarebyte

app/assets/javascripts/setup-repo.js
```diff
@@ -0,0 +1,113 @@
...
+
+  show_success = function() {
+    location.href = success_path();
+  };
+
+  check_progress = function() {
+    var progress_indicator = $('.setup-progress');
+    var path = progress_path();
+
+    $.getJSON(path, function(data) {}).done(function(data){
```
nwoodthorpe: What happens here if we hit our rate limit, or we lose connectivity?

app/assets/javascripts/setup-repo.js
```diff
@@ -0,0 +1,113 @@
...
+    if(setup_progress.length != 0){
+      configuring = false;
+      (function() {
+        var import_indicator = $('#import-progress');
+        indicate_in_progress(import_indicator);
+
+        setInterval(function(){
+          check_progress();
+        }, 3000);
+      })();
```
nwoodthorpe: Whys this bit in an anonymous function?

app/assets/javascripts/setup-repo.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+  })
+
+  $(document).ready(ready);
+}).call(this);
```
nwoodthorpe: I really like the looks of this file :+1:

app/assets/stylesheets/components/setup-progress.scss
```diff
@@ -0,0 +1,15 @@
+.status {
+  display: inline-block;
+  min-width: 24px;
+  max-width: 24px;
+  min-height: 24px;
+  max-height: 24px;
```
nwoodthorpe: Are these 4 lines the same as `width: 24px; height: 24px;`?

app/controllers/assignment_invitations_controller.rb
```diff
@@ -2,16 +2,30 @@
...
 
   def accept
     create_submission do
-      redirect_to success_assignment_invitation_path
+      redirect_to setup_assignment_invitation_path
     end
   end
 
+  def setup
+    redirect_to success_assignment_invitation_path unless repo_setup_enabled?
```
nwoodthorpe: This should probably be a before_action

app/controllers/assignment_invitations_controller.rb
```diff
@@ -35,6 +49,19 @@ def check_user_not_previous_acceptee
     redirect_to success_assignment_invitation_path
   end
 
+  def classroom_config
```
nwoodthorpe: Is it possible to hit this method but not have a starter code repo?

app/controllers/concerns/setup_repo.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    progress.update(status: 'complete') unless config.github_repository.branch_present? 'github-classroom'
+    progress.update(status: 'configuring', message: CONFIGURATION_PROGRESS) if repo.branch_present? 'github-classroom'
+    progress.update(status: 'complete') unless config.finished_setup? repo
+    progress
+  end
+
+  def perform_setup(repo, config)
+    config.setup_repository(repo)
+    return true
```
nwoodthorpe: we don't need the two `return`s in this method :P

app/models/assignment_repo/creator.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class AssignmentRepo
+  # rubocop:disable Metrics/ClassLength
```
nwoodthorpe: You need to `rubocop:enable` this at the end of the class

app/services/classroom_config.rb
```diff
@@ -0,0 +1,45 @@
...
+  attr_reader :github_repository
+
+  def initialize(github_repository)
+    raise ArgumentError, 'Invalid configuration repo' unless github_repository.branch_present? 'github-classroom'
+    @github_repository = github_repository
+  end
+
+  def setup_repository(repo)
+    configs_tree = @github_repository.branch_tree('github-classroom')
+    GitHub::Errors.with_error_handling do
```
nwoodthorpe: We shouldn't need this, `#tree` returns {} in the error case I believe.

app/services/classroom_config.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  # Internal: Generates issues for the assignment_repository based on the configs
+  #
+  # repo - GitHubRepository for which to perform the configuration
+  #                   setups
+  # tree_sha     - sha of the 'issues' tree
+  #
+  # Returns nothing
+  def generate_issues(repo, tree_sha)
+    GitHub::Errors.with_error_handling do
```
nwoodthorpe: Same here. If a method we're using can raise a GitHub::Error, we can rescue that specifically instead of wrapping in this.

config/routes.rb
```diff
@@ -28,6 +28,8 @@
   resources :assignment_invitations, path: 'assignment-invitations', only: [:show] do
     member do
       patch :accept
+      get   :setup
+      match :setup_progress, via: %i[get patch]
```
nwoodthorpe: why do we need both here?

app/controllers/assignment_invitations_controller.rb
wilfriedE: Nope, it only gets called during setup. So if it doesn't have a starter_code this shouldn't be called

wilfriedE:
Closing this PR in favor of  #1030 , #1033 , #1036

PR #984, opened 2017-07-04
Migration to add tables for new roster models, + roster_id on org

nwoodthorpe:
Works towards #986 

As discussed here: https://github.com/education/classroom/issues/448#issuecomment-311786419

These are the new tables for rosters. Since an organization belongs to a roster, the organization now has a `roster_id` column

### DIFF
--- a/db/migrate/20170629203747_add_new_identifier_tables.rb
+++ b/db/migrate/20170629203747_add_new_identifier_tables.rb
@@ -1,12 +1,12 @@
 class AddNewIdentifierTables < ActiveRecord::Migration[5.1]
   def change
-    create_table :roster do |t|
+    create_table :rosters do |t|
       t.string :identifier_name, null: false
 
       t.timestamps null: false
     end
 
-    create_table :roster_entry do |t|
+    create_table :roster_entries do |t|
       t.string :identifier, null: false
       t.references :roster, null: false
       t.references :user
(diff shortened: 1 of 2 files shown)


db/migrate/20170629203747_add_new_identifier_tables.rb
```diff
@@ -0,0 +1,20 @@
+class AddNewIdentifierTables < ActiveRecord::Migration[5.1]
+  def change
+    create_table :rosters do |t|
+      t.string :identifier_name, null: false
```
tarebyte: Will we ever search using the identifier name?

db/migrate/20170629203747_add_new_identifier_tables.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      t.timestamps null: false
+    end
+
+    create_table :roster_entries do |t|
+      t.string :identifier, null: false
+      t.references :roster, null: false
+      t.references :user
```
tarebyte: Does a roster entry need to have a user? Since we don't specify I'm assuming not, but I just wanted to be sure.

db/migrate/20170629203747_add_new_identifier_tables.rb
nwoodthorpe: Almost certainly not, the identifier_name is just a human readable description of what the entry identifiers represent. e.g. 'your student email'

tarebyte:
Deployed and run in production

PR #985, opened 2017-07-04
Models for rosters

### LINKED ISSUE #986: Rosters: Create Roster and RosterEntry model
We can't have a roster if we don't have a Roster model 🐰

nwoodthorpe:
Closes #986 

Models for rosters: https://github.com/education/classroom/issues/448#issuecomment-311786419

Based on #984 so until that's merged, the diff will look weird. The last 2 commits are from this PR.

### DIFF
--- a/app/models/roster_entry.rb
+++ b/app/models/roster_entry.rb
@@ -1,0 +1,4 @@
+class RosterEntry < ApplicationRecord
+  belongs_to :roster
+  belongs_to :user, optional: true
+end
--- a/app/models/roster_entry.rb
+++ b/app/models/roster_entry.rb
@@ -1,4 +1,9 @@
+# frozen_string_literal: true
+
 class RosterEntry < ApplicationRecord
   belongs_to :roster
   belongs_to :user, optional: true
+
+  validates :identifier, presence: true
+  validates :roster,     presence: true
 end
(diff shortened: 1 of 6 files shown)


app/models/roster_entry.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RosterEntry < ApplicationRecord
+  belongs_to :roster
+  belongs_to :user, optional: true
+
+  validates :identifier, presence: true
+  validates :roster,     presence: true
```
tarebyte: Is this optional since `belongs_to` enforces presence?

spec/models/roster_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+require 'rails_helper'
+
+RSpec.describe Roster, type: :model do
+  subject { create(:roster) }
+
+  describe 'associations' do
+    it 'can have roster_entries' do
+      subject.roster_entries << create(:roster_entry)
+
+      expect(subject.valid?).to be_truthy
```
tarebyte: I really miss MiniTest for stuff like this.

```ruby
assert_predicate subject, :valid?
```

app/models/roster_entry.rb
nwoodthorpe: For some reason, not having the explicit validation makes the db throw an exception before the model

PR #999, opened 2017-07-06
p tag no longer covers lower half of button

### LINKED ISSUE #998: Lower half of "Manage Classroom" button isn't clickable
When hovering over the lower half of the Manage Classroom button, the button does not work (both click events and hover styling)

nwoodthorpe:
Closes #998 

The `p` tag was inheriting width from the parent and covering the lower half of the button. `display: inline-block` means it only takes as much width as it needs.


nwoodthorpe:
Lgtm in prod :+1:

PR #1005, opened 2017-07-08
Setup knapsack and hound

tarebyte:
This enables parallel testing on CI with [knapsack](https://github.com/ArturT/knapsack) and [Hound CI](https://houndci.com/repos) for doing out linting instead of running a script on Travis.


spec/support/knapsack.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+require 'knapsack'
```
nwoodthorpe: Looks like hound works 😆

PR #1006, opened 2017-07-08
Minitest spike for thoughts

tarebyte:


nwoodthorpe:
I prefer minitest assertions, I :+1: this. We can still namespace our specs by what we're testing though, right? 

```
describe '#staff?' do
  test 'does this'

  test 'does that'
end
```

tarebyte:
@nwoodthorpe not without minitest spec included as part of the gemfile

PR #1014, opened 2017-07-12
Don't run uniqueness validation for invitations with nil short keys

nwoodthorpe:
We should not be running uniqueness checks for invitations with nil short_keys (e.g. all existing invitations).

Also wrote regression specs for assignment and group assignment invitation.

### DIFF
--- a/spec/models/group_assignment_invitation_spec.rb
+++ b/spec/models/group_assignment_invitation_spec.rb
@@ -10,7 +10,8 @@ RSpec.describe GroupAssignmentInvitation, type: :model do
 
   describe 'short_key' do
     it 'allows multiple invitations with nil short_key' do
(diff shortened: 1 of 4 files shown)


spec/models/group_assignment_invitation_spec.rb
```diff
@@ -8,6 +8,17 @@
     expect(group_assignment_invitation.key).to_not be_nil
   end
 
+  describe 'short_key' do
+    it 'allows multiple invitations with nil short_key' do
+      first_inv, second_inv = create(:group_assignment_invitation), create(:group_assignment_invitation)
```
nwoodthorpe: Awe

nwoodthorpe:
Ran `script/linter` locally.

```
==> Running rubocop...
Inspecting 242 files
.....................................................................................................................................
.............................................................................................................

242 files inspected, no offenses detected
==> Running scss-lint...
```

PR #1018, opened 2017-07-14
Default deadline datepicker time to next hour

nwoodthorpe:
Thanks @mozzadrella for doing user testing and finding this UX issue ✨ 

Currently when you don't have a deadline, the datepicker time defaults to the current hour and 0 minutes, which means the default deadline is always in the past and is invalid.

In most cases, the teacher will be selecting a deadline from the datepicker, but if someone just accepts the default deadline (maybe to see how the deadlines feature works), we shouldn't error.

We now default to the NEXT hour.

screen shot 2017-07-14 at 10 18 47 am

Also: This is the last fix before everything has been addressed. After that, we should be 🍏 to full ship.

### DIFF
--- a/app/assets/javascripts/datetimepicker.js
+++ b/app/assets/javascripts/datetimepicker.js
@@ -10,6 +10,13 @@ function leftpad(val){
   return ("0" + val).slice(-2);
 }
 
+function timeRoundedUp() {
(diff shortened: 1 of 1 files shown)


nwoodthorpe:
Tested in prod, all good :+1:

Issue #1010, opened 2017-07-12
Stafftools button misaligned


nwoodthorpe:
The impersonate button is a little bugged out right now.

<img width="752" alt="screen shot 2017-07-12 at 9 44 45 am" src="https://user-images.githubusercontent.com/4149056/28129147-2d66468a-66e7-11e7-9d14-35ee63769cfe.png">

arolson:
@nwoodthorpe would you know how to replicate the issue?

nwoodthorpe:
@arolson creating a new user, then viewing the stafftools page for that user should be sufficient. You'll need to make sure the user you're logged in as is set as an admin in your `.env`.

localhost:5200/stafftools/users/<user_id>

Just checked in prod, this hasn't been fixed yet.

arolson:
Ok awesome I will take a look.

PR #1020, opened 2017-07-16
issue #1010, CSS update stafftools.scss, show.html.erb

arolson:
Repositioned the Impersonate button in stafftools. Changes were made to files `app/assets/stylesheets/stafftools.scss` and `app/views/stafftools/users/show.html.erb`.

### DIFF
--- a/app/assets/stylesheets/stafftools.scss
+++ b/app/assets/stylesheets/stafftools.scss
@@ -53,3 +53,9 @@
   overflow: hidden;
   line-height: 27px;
 }
+// Issue #1010: reposition Impersonate user button
+.imp-user{
+    position: absolute;
+    bottom: 10px;
+    right: 10px;
+}
--- a/app/views/stafftools/users/show.html.erb
+++ b/app/views/stafftools/users/show.html.erb
@@ -86,7 +86,7 @@
 
       <div class="boxed-group-inner">
         <h4>Impersonate <%= "@#{@user.github_user.login}" %></h4>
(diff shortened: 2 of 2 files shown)


nwoodthorpe:
Would we be able to do this without using `bottom`, `right` and `position: absolute`? Making things position absolute affects the flow of other elements, so styling things in the future might become more difficult because of this.

We have other buttons that are positioned correctly without using position: absolute, maybe look into those. Usually when we need to vertically center, my goto is flexbox.

Also, can you add a screenshot of the button after your change?

Thanks :)

arolson:
@nwoodthorpe yea actually I was able to reproduce similar results using the flex-box. Thanks for the suggestions! The first image is of the css modification, the second is using the flex-box. This was my first attempt at playing with ruby apps, so it was kind of fun 😆 

* CSS:
screen shot 2017-07-16 at 10 55 02 am

* Flexbox: 
screen shot 2017-07-16 at 12 19 53 pm
wrapped the `<h4>` header and the `<p>` tag in a `<div>` then did the same with the `<a>` tag.
Treated them as separate flex-items.

tarebyte:
👋

tarebyte:
👋 

I think my changes in #1021 will actually solve the issue here without the very specific CSS

https://github.com/education/classroom/pull/1021/files#diff-4cc49612e087993b6f5b7df19a58c7fcR83

nwoodthorpe:
Yeah, that looks better. Sorry @arolson, this bug was caused by something bigger. I'm going to close this in favor of #1021.

PR #1024, opened 2017-07-17
Dissallow empty rosters

nwoodthorpe:
We should show the user an error if they create a roster with no students


app/controllers/rosters_controller.rb
```diff
@@ -12,16 +12,21 @@ def new
 
   def create
     @roster = Roster.new(identifier_name: params[:identifier_name])
-    @roster.save!
 
     add_identifiers_to_roster
 
-    @organization.roster = @roster
-    @organization.save!
+    if @roster.roster_entries.empty?
```
tarebyte: This feels like something that should be part of the model, or a new creator. WDYT?

nwoodthorpe: Yeah you're right, I can enforce this on the model. :+1:

app/controllers/rosters_controller.rb
```diff
@@ -12,16 +12,21 @@ def new
...
+      @roster.errors.add(:roster_entries, "must include at least one student")
```
nwoodthorpe: Going to ignore this

nwoodthorpe:
@tarebyte updated

nwoodthorpe:
Also made sure we have no rosters with 0 entries on production: 

```
pry(main)> Roster.all.select{|r| r.roster_entries.empty?}
=> []
```

PR #1025, opened 2017-07-17
Enforce double quotes cause I'm getting tired of getting it wrong

tarebyte:
The default GitHub config uses double quotes for everything, and personally I really like it.

Does anyone have any strong feelings as to why I shouldn't merge this?


nwoodthorpe:
I've always been told that `'` is better for performance but

https://www.viget.com/articles/just-use-double-quoted-ruby-strings

> statistically insignificant difference

:+1:

PR #1027, opened 2017-07-17
Fix permissions error

nwoodthorpe:
When a user hits the short_url controller, the before action in `application_controller` asks for full teacher scopes. 

We should skip and authenticate after we redirect.

@tarebyte @johndbritton


nwoodthorpe:
Tested in prod, :+1:

PR #1028, opened 2017-07-18
Add bootsnap gem

tarebyte:
[Bootsnap](https://github.com/Shopify/bootsnap) by Shopify was just added to [Rails core](https://github.com/rails/rails/pull/29313), so I figured what the heck why not ¯\\_(ツ)_/¯


nwoodthorpe:
:+1:, bootsnap is great!

PR #1029, opened 2017-07-18
Allow users to delete roster entries from a roster

nwoodthorpe:
This change allows roster entries to be deleted.

screen shot 2017-07-18 at 12 32 20 pm

screen shot 2017-07-18 at 12 32 28 pm

screen shot 2017-07-18 at 12 32 45 pm


nwoodthorpe:
Looks good in prod :+1:

PR #1030, opened 2017-07-18
Add  GitHubBlob model

wilfriedE:
#983 is too big so breaking it down to smaller PRs.


app/models/github_blob.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(github_repository, sha, **options)
+    @blob = github_repository.client.blob(github_repository.full_name, sha, options)
+    read_contents
+  end
+
+  # Public: Get the utf-8 formated content of the blob
+  #
+  # Returns a string.
+  def utf_content
```
nwoodthorpe: This is 1-1 the same as `decoded_content`, we can probably just take the body from decoded content and throw it in here.

nwoodthorpe:
Small comment. Besides that, this lgtm

app/models/github_blob.rb
```diff
@@ -0,0 +1,47 @@
...
+  def decoded_content
+    return @blob.content if @blob.content == "utf-8"
+    Base64.decode64(@blob.content)
+  end
+
+  # Internal: process yaml front matter
+  #
+  # Returns nothing.
+  def read_contents
+    match = YAML_FRONT_MATTER_REGEXP.match(decoded_content)
```
tarebyte: Style question, non blocking WDYT about this?

```ruby
if (match = YAML_FRONT_MATTER_REGEXP.match(decoded_content))
  @body = match.post_match
  @data = SafeYAML.load(match.to_s)
end
```

app/models/github_repository.rb
```diff
@@ -34,6 +34,17 @@ def invite(user, **options)
     end
   end
 
+  # Public: Get a blob from the GitHub repository.
+  #
+  # sha    - sha of the blob.
```
tarebyte: Is `sha` of type `String`?

app/models/github_blob.rb
wilfriedE: The linter throws an Exception for this. It prefers using a GuardClause over wrapping it in a conditional

app/models/github_repository.rb
wilfriedE: yep, `String`.

PR #1032, opened 2017-07-19
Prompt users to join the roster when they accept an assignment

nwoodthorpe:
When a user opens an invitation, we need to prompt them to join the roster. If their class has a roster and they aren't already on it.

This PR adds a step before the accept step where the student can choose to link themself or skip for now. If they skip, they'll be asked again the next time they open an invitation.

The list of identifiers is a scrollable list with a max-height of 200px, so it looks OK with a large number of identifiers.

Idea: A search bar to find your identifier? I don't want that to be part of this PR, but it would be nice to have.

Things to do:
- [x] Users get prompted for assignment invitations
- [x] Users get prompted for group assignment invitations
- [x] Try to DRY up the code (lots of repeated stuff between the two)
- [x] Assignment invitation specs
- [x] Group assignment invitation specs

screen shot 2017-07-19 at 1 53 38 pm


nwoodthorpe:
Welp I can't remove you guys as reviewers since you're code owners. This is WIP, I'll ping when it's good to go

nwoodthorpe:
Going to deploy to make sure code is :+1: but I'm going to wait for @mozzadrella before I merge

nwoodthorpe:
Tested in prod, everything looks good, no exceptions. :+1:

PR #1033, opened 2017-07-20
Add Classroom config service

wilfriedE:
A piece of #983 .


spec/support/repository_factory.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+require_relative "vcr"
+require "safe_yaml/load"
```
wilfriedE: Remove "safe_yaml" dependency, use existing yaml parser

nwoodthorpe:
Lgtm, stub is a lot more descriptive than example to me :)

app/services/classroom_config.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class ClassroomConfig
+  CONFIGURABLES = %w[issues].freeze
+
+  attr_reader :github_repository
+
+  def initialize(github_repository)
+    raise ArgumentError, "Invalid configuration repo" unless github_repository.branch_present? "github-classroom"
```
tarebyte: @wilfriedE can we make `"github-classroom"` a constant?

spec/fixtures/repos/configured-repo/import_progress.json
```diff
@@ -0,0 +1,15 @@
+{
+  "vcs": "subversion",
```
tarebyte: Let's make this a git example since we'll actually never import from svn 😄

spec/support/repository_factory.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require_relative "vcr"
+
+class StubRepository
+  attr_reader :full_name, :branches
+
+  def repo_file(repo_name, path)
```
tarebyte: Could we rename this to be just `#file`? Since it's on the `StubRepository` already.

spec/support/repository_factory.rb
```diff
@@ -0,0 +1,91 @@
...
+    File.read(Rails.root.join("spec/fixtures/repos/#{repo_name}", path.to_s))
+  end
+
+  def repo_json(repo_name, path)
```
tarebyte: Is there a more descriptive name we can give this? I have a general idea what it does, but from the method name it's not clear.

spec/support/repository_factory.rb
```diff
@@ -0,0 +1,91 @@
...
+    @branches.select { |b| b.name == name }.last
+  end
+
+  def branch_tree(name)
+    return {} unless branch_present? name
+    Tree.new(self, branch(name).commit.sha)
+  end
+
+  def tree(sha)
+    Tree.new(self, sha)
```
tarebyte: Can we memoize this method and `#blob`?

app/models/github_repository.rb
```diff
@@ -34,6 +40,31 @@ def invite(user, **options)
     end
   end
 
+  # Public: Add issues to the GitHub repository.
+  #
+  # title    - The title of the issue.
+  # body     - The body of the issue.
+  #
+  # Returns the newly created issue, or raises a GitHub::Error.
+  def add_issue(title, body, **options)
```
tarebyte: Can we just call this `#create_issue`

app/services/classroom_config.rb
```diff
@@ -0,0 +1,48 @@
...
+    @github_repository = github_repository
+  end
+
+  def setup_repository(repo)
+    configs_tree = @github_repository.branch_tree("github-classroom")
+    configs_tree.tree.each do |config|
+      send("generate_#{config.path}", repo, config.sha) if CONFIGURABLES.include? config.path
```
tarebyte: I usually get nervous when we use `send` but it looks like this is only from GitHub API responses correct?

wilfriedE: yeah! this is only from the API responses.

spec/support/repository_factory.rb
tarebyte: Yeah that'd be fine I think

app/services/classroom_config.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+class ClassroomConfig
+  CONFIGURABLES   = %w[issues].freeze
```
nwoodthorpe: Strings are now frozen by default, so we don't need to freeze this manually

app/services/classroom_config.rb
```diff
@@ -0,0 +1,65 @@
...
+  CONFIGBRANCH    = "github-classroom"
```
nwoodthorpe: We still underscore our constants. E.g. `INVALID_SELECTION = "Invalid repository selection, please check it again."`

app/services/classroom_config.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  # Internal: Generates issues for the assignment_repository based on the configs
+  #
+  # repo - GitHubRepository for which to perform the configuration
+  #                   setups
+  # tree_sha     - sha of the "issues" tree
+  #
+  # Returns nothing
+  def generate_issues(repo, tree_sha)
+    @github_repository.tree(tree_sha).tree.each do |issue|
```
nwoodthorpe: Why do we need to do `.tree` after `github_repository.tree`? This seems a little strange

wilfriedE: There is a tree field that contains the list of items under the tree. I am going to make a helper to make this easier to read.

app/services/classroom_config.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+class ClassroomConfig
+  CONFIGURABLES   = %w[issues]
```
nwoodthorpe: Hrm... I'm almost positive that we don't need to freeze strings anymore. @tarebyte do you know?

PR #1035, opened 2017-07-20
Fix for PR #1030

wilfriedE:
#1030


nwoodthorpe:
This LGTM, we already use YAML.safe_load in the videos area so this should be ok.

PR #1036, opened 2017-07-20
Add repo setup and Issues initialization

wilfriedE:
Another piece of #983


nwoodthorpe:
Right now if they leave before config starts, they'll get access to the classroom-config branch. I think merging this without addressing that is OK since we're behind a flag, but that'll need to be addressed before we properly ship

nwoodthorpe:
Other possibilities: API goes down during import/config or users internet dies. We should fail cleanly in these cases

nwoodthorpe:
Also, we should consider the case where the classroom-config is invalid (missing files, bad yml, etc)

app/controllers/assignment_invitations_controller.rb
```diff
@@ -2,16 +2,29 @@
...
 
   def accept
     create_submission do
-      redirect_to success_assignment_invitation_path
+      redirect_to setup_assignment_invitation_path
     end
   end
 
+  def setup
+    starter_code_repo_id = current_submission.assignment.starter_code_repo_id
```
nwoodthorpe: Small style comment: Reading this code makes you think we need the id for something. Really, we need to know whether or not the id exists for something. We could change this line to:
```
starter_code_repo_present? = current_submission.assignment.starter_code_repo_id.present?
```
which I think reads a little better.

app/controllers/assignment_invitations_controller.rb
```diff
@@ -2,16 +2,29 @@
...
+    redirect_to success_assignment_invitation_path unless starter_code_repo_id
+  end
+
+  def setup_progress
+    perform_setup(current_submission, classroom_config) if configurable_submission?
```
nwoodthorpe: Why do we need to `perform_setup` when the user pings to inquire about setup progress?

app/controllers/assignment_invitations_controller.rb
```diff
@@ -35,6 +48,24 @@ def check_user_not_previous_acceptee
     redirect_to success_assignment_invitation_path
   end
 
+  def check_authorized_repo_setup
```
nwoodthorpe: I prefer `ensure` over `check` in these types of methods. Ensure sort of implies something will be done if the check fails.

app/controllers/assignment_invitations_controller.rb
```diff
@@ -35,6 +48,24 @@ def check_user_not_previous_acceptee
...
+    client               = current_submission.creator.github_client
+
+    starter_repo         = GitHubRepository.new(client, starter_code_repo_id)
+    ClassroomConfig.new(starter_repo)
+  end
+
+  def configurable_submission?
+    repo = current_submission.github_repository
+    configurable = classroom_config.configurable? repo
+    repo.import_progress[:status] == "complete" && configurable && params[:configure]
```
nwoodthorpe: What does params[:configure] here represent? It might be nice to wrap this in a method to describe what it represents

```
def something_meaningful
  params[:configure]
end
```

app/controllers/concerns/setup_repo.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module SetupRepo
+  extend ActiveSupport::Concern
+  IMPORT_IN_PROGRESS     = "Importing starter code to student repo"
+  CONFIGURATION_PROGRESS = "Configuring the student repo"
+
+  def setup_status(repo, config)
+    progress = { status: "importing", message: IMPORT_IN_PROGRESS }
```
nwoodthorpe: If the status is only being used for comparisons, these should be either symbols or strings referenced by a constant.

app/views/assignment_invitations/setup.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    <div id="config-progress" class="col-4 border rounded-1 mr-4 p-4">
+      <span class='status'><%= octicon("check", class: 'text-green complete-indicator', height: 30) %></span>
+      <p class="v-align-baseline f4 mt-4">2. Configuring...</p>
+    </div>
+    <div id="finalize-progress" class="col-4 border border-gray rounded-1  p-4">
+      <span class='status'><%= octicon("check", class: 'text-green complete-indicator', height: 30) %></span>
+      <p class="v-align-baseline f4 mt-4">3. Finalize</p>
+    </div>
+  </div>
+</div>
```
nwoodthorpe: This view looks great!

app/views/assignment_invitations/show.html.erb
```diff
@@ -21,7 +21,8 @@
     </div>
 
     <div class="form-actions">
-      <%= submit_tag 'Accept this assignment', class: 'btn btn-primary js-navigation' %>
+      <% loader = ''; loader = ' js-navigation' unless repo_setup_enabled? %>
```
nwoodthorpe: What about `<% loader = (repo_setup_enabled?) ? '' : ' js-navigation' %>`

db/schema.rb
```diff
@@ -37,6 +37,7 @@
     t.integer "user_id"
     t.string "submission_sha"
     t.string "global_relay_id"
+    t.integer "configuration_state", default: 0
```
tarebyte: We'll have to extract this out and run it separately once we're done

app/controllers/concerns/repo_setup.rb
```diff
@@ -0,0 +1,25 @@
...
+  extend ActiveSupport::Concern
+
+  def setup_status(assignment_repo)
+    repo     = assignment_repo.github_repository
+    progress = { status: :importing }
+
+    return progress unless repo.import_progress[:status] == "complete"
+
+    progress.update(status: :configuring) if assignment_repo.configuring?
+    progress.update(status: :complete) if assignment_repo.configured? || !repo.branch_present?("github-classroom")
```
tarebyte: Can we move the branch name to an application constant?

I'm not a huge fan of the rando strings hanging around.

```ruby
GitHubClassroom.default_branch
```

Or something of the like

app/controllers/concerns/repo_setup.rb
```diff
@@ -0,0 +1,29 @@
...
+  extend ActiveSupport::Concern
+
+  def setup_status(assignment_repo)
+    repo     = assignment_repo.github_repository
+    progress = { status: :importing }
+
+    return progress unless repo.import_progress[:status] == "complete"
+
+    progress.update(status: :configuring) if assignment_repo.configuring?
+    progress.update(status: :complete) if assignment_repo.configured? || !repo.branch_present?(config_branch)
```
tarebyte: I don't think we get a lot out of abstracting the constant to a method, let's go ahead and just use `ClassroomConfig::CONFIG_BRANCH`

wilfriedE: Do you have any recommendations for this? Would it be better to have a method under GitHubClassroom for it instead?

tarebyte: Oh no I just meant

```diff
-    progress.update(status: :complete) if assignment_repo.configured? || !repo.branch_present?(config_branch)
+    progress.update(status: :complete) if assignment_repo.configured? || !repo.branch_present?(ClassroomConfig::CONFIG_BRANCH)
 ```

wilfriedE: I see

PR #1037, opened 2017-07-20
New view for Assignment/show

### LINKED ISSUE #989: Rosters: Ship new Assignment/show view
https://github.com/education/classroom/issues/448#issuecomment-311516206

Assignment/show will need to be updated for classrooms that have a roster.

nwoodthorpe:
Closes #989

~~[code is super WIP, more looking for design feedback right now]~~

Now that we have rosters, we need a new view for assignment/show and group-assignment/show.

Users with no roster will still see the old view:

screen shot 2017-07-20 at 1 35 32 pm

The new view looks like this:

screen shot 2017-07-20 at 1 32 23 pm

The first tab is a list of all roster entries, instead of a list of all assignment repos like it used to be.

A roster entry can be in one of 3 different states:
 - Accepted and working on the assignment
 - In the classroom (linked to a GH account) but hasn't accepted the assignment
 - Not yet joined the classroom

This view shows which state each student is in.

The second tab is a list of Assignment repos that have started the assignment but haven't linked to a roster entry:

screen shot 2017-07-20 at 1 32 30 pm

This tab format is identical to the roster management UI.


JasonEtco:
Re: design, I think the tab-group bottom border should extend to the full width of the viewport. It looks funny break off into space like that, and its more consistent to keep going until it hits some kind of wall. Might be a pain with your current markup tho 😑

nwoodthorpe:
It will be done. 💖

nwoodthorpe:
Actually we're using a similar design on the roster page. Might be better to stick to this for consistency. If anyone is super for making it full width, I can make the change on both.

screen shot 2017-07-21 at 8 56 35 am

JasonEtco:
> making it full width

To be clear, I don't mean the whole thing, I just mean to have the line below "Unlinked GitHub Accounts" (probably a `border-top` or something) extend to be full width. I'm still 👍 for making the change on both 🌮

nwoodthorpe:
Yup. Oki, I'll do that in another PR :+1:

nwoodthorpe:
@tarebyte @wilfriedE this is good for review :+1:

app/controllers/roster_entries_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    elsif @roster_entry.user
+      render "roster_entries/assignment_repos/linked_not_accepted"
+    else
+      render "roster_entries/assignment_repos/not_in_classroom"
+    end
+  end
+
+  private
+
+  def ensure_authorized
```
tarebyte: Can we not just use `include OrganizationAuthorization` here?

app/services/roster_entry_sorter.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# Sorts roster entries by:
+# - Linked + accepted
+# - Linked + not accepted
+# - Not linked
+
+class RosterEntrySorter
```
tarebyte: Let's get some tests around this service

app/services/roster_entry_sorter.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(roster_entries, assignment)
+    @entries = roster_entries
+    @assignment = assignment
+  end
+
+  def sort
+    users_with_repo = @assignment.repos.map(&:user_id)
+
+    @entries.sort_by do |entry|
```
tarebyte: You could clean this up a bit using `next`

```ruby
@entries.sort_by do |entry|
  next 2 unless entry.user.present?
  users_with_repo.include?(entry.user.id) ? 0 : 1
end
```

app/services/roster_entry_sorter.rb
```diff
@@ -0,0 +1,27 @@
...
+class RosterEntrySorter
+  attr_reader :entries, :assignment
+
+  def initialize(roster_entries, assignment)
+    @entries = roster_entries
+    @assignment = assignment
+  end
+
+  def sort
+    users_with_repo = @assignment.repos.map(&:user_id)
```
tarebyte: Can use use [`#pluck`](http://guides.rubyonrails.org/active_record_querying.html#pluck) here? That way we don't have to load all of the data for each record?

```ruby
users_with_repo = @assignment.repos.pluck(:user_id)
```

app/views/assignment_repos/_unlinked_user.html.erb
```diff
@@ -0,0 +1,33 @@
+<% view = AssignmentRepoView::ShowView.new(assignment_repo: assignment_repo) %>
+
+
+<div class='d-flex flex-row flex-justify-between flex-items-center assignment-repo-list-item'>
+  <div class='d-flex'>
+    <div>
+      <%= link_to unlinked_user.github_user.html_url, class: "tooltipped tooltipped-n", target: '_blank', 'aria-label' => unlinked_user.github_user.login do %>
+        <%= image_tag unlinked_user.github_user.github_avatar_url(96), height: 48, width: 48, class: 'avatar avatar-large', alt: "@#{unlinked_user.github_user.login}" %>
```
tarebyte: Thought for another PR, this is becoming a pattern.

We should think about moving this out to a view helper of some kind.

app/services/roster_entry_sorter.rb
nwoodthorpe: TIL, I actually didn't know you could set the return value of a block using next O.o

app/views/assignment_repos/_unlinked_user.html.erb
nwoodthorpe: Totally agree. I want to do the group assignment view after this, then spend some time seeing what we can extract into shared views or view helpers. :+1:

nwoodthorpe:
@tarebyte all feedback addressed :)

nwoodthorpe:
Going to deploy to test out. @wilfriedE ill wait for your review before I merge

nwoodthorpe:
This looks good in production :+1:

PR #1045, opened 2017-07-27
Different hover color for delete roster entry button

nwoodthorpe:
The current delete button is sort of bugging me: Way too much red.

screen shot 2017-07-27 at 8 31 31 am

Idea 1:

screen shot 2017-07-27 at 8 29 03 am

Idea 2 (taken from the GitHub modal styles)

screen shot 2017-07-27 at 8 35 06 am

Preferences? I'm leaning towards 2, especially since it's what we use right now in our modals.

/cc @mozzadrella


tarebyte:
@nwoodthorpe I like 2 as well

nwoodthorpe:
Great 😄 the commit I pushed is for 2.

PR #1048, opened 2017-07-28
Add functionality to upload list of students to roster

### LINKED ISSUE #1044: Rosters: Create roster by uploading CSV
Note: in this context, a single column CSV and a newline seperated text file are the exact same thing.

nwoodthorpe:
Closes #1044 

Adds the ability for teachers to upload a list of students to their roster.

When they choose a file to upload, we add the list to the box above for manually entering in entries. This way, the teacher can verify what was read before submitting.

It's also a super minimalistic approach. We do no validation or processing on the input, partially because I don't think we should make assumptions like assuming the format of the entries.

Image


nwoodthorpe:
LGTM in prod :+1:

PR #1051, opened 2017-07-31
Datadog for Classroom

nwoodthorpe:
This PR pulls in the datadog statsd ruby client, and begins logging a single metric. I'd like to merge this before going forward to check our datadog stuff is setup right in Heroku.


config/initializers/statsd.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require "datadog/statsd"
+
+class StubStatsd
+  def increment; end
```
tarebyte: Could we make this match the [original definition](https://github.com/DataDog/dogstatsd-ruby/blob/master/lib/datadog/statsd.rb#L120-L131)?

```ruby
class StubStatsd
  # https://github.com/DataDog/dogstatsd-ruby/blob/d77b4bbaa734b2e7b02875c588fccc4221864276/lib/datadog/statsd.rb#L120-L131
  def increment(stat, opts={}); end
end
```

nwoodthorpe:
Whoops, total brainfart. Forgot ruby enforces number of arguments :P

nwoodthorpe:
Most of the stubbing/mocking work I've done has been with minitest, but we SHOULD be able to do something like

```Ruby
it 'sends classroom.created event to statsd' do
    expect do
      # Run the creator
    end.to call(GitHubClassroom.statsd.increment).with_arguments('classroom.created')
end
```

I'll look into it and ping you when I figure something out

nwoodthorpe:
@tarebyte This look good?

tarebyte:
> This look good?

Yup looks good to me, I'll add back in the buildpack and give you a ping

tarebyte:
Buildpack is added, we'll try this out later today

PR #1053, opened 2017-08-01
More instrumentation

nwoodthorpe:
This adds datadog instrumentation for:

`exercise.create`
`group_exercise.create`
`exercise.destroy`
`group_exercise.destroy`
`deadline.create`
`exercise_invitation.accept`
`group_exercise_invitation.accept`
`roster.create`
`exercise_invitation.fail`
`group_exercise_invitation.fail`

I'll do the webhook ones in another PR

Also did specs for all of the above


tarebyte:
@nwoodthorpe since the naming will change probably sooner than later, can we change `assignment` to be `exercise` ? That way we don't lose out on metrics?

nwoodthorpe:
Looks good in prod :+1:

PR #1054, opened 2017-08-01
Send exceptions to statsd

nwoodthorpe:
Increment `exception.raise` when an exception bubbles up to the ApplicationController. 

It'd be nice to test this but we don't really have anything in place to spec out ApplicationController behaviour. @tarebyte if you think it'd be useful, I can put together a spec file in another PR.


app/controllers/application_controller/errors_dependency.rb
```diff
@@ -17,4 +17,9 @@ def flash_and_redirect_back_with_message(exception)
 
     redirect_back(fallback_location: root_path)
   end
+
+  def send_to_statsd(exception)
+    GitHubClassroom.statsd.increment("exception.raise")
```
tarebyte: it would be ✨ if we could include the exception type as part of the stat

nwoodthorpe:
@tarebyte hows this? Made it as a tag instead of part of the event name so we can easily show the total number of exceptions

tarebyte:
Looks good to me @nwoodthorpe 🚢 💨

nwoodthorpe:
Just gave this another look in dev, everything looks like it's working fine... Exceptions are being caught and sent to statsd. Might have had something wrong on the Datadog end, going to try deploying again

nwoodthorpe:
This works!

PR #1056, opened 2017-08-02
Swallow deserialization errors in repository visibility job

nwoodthorpe:
We've been getting quite a few of these in Airbrake lately:

screen shot 2017-08-01 at 5 05 10 pm

I looked into it, serialization from ID to assignment object fails when the assignment no longer exists. The default behavior for these jobs is to retry after a period of time if they fail. But if the assignment no longer exists, there's no point in running this job.

Instead, let's just swallow the exception so the job does not run again.


nwoodthorpe:
I was looking at writing a test, but I'm not really sure how it'd go. The behavior happens when we enqueue the job for later, which serializes to redis, then when sidekiq pulls the job from redis later, it fails to deserialize because the assignment was deleted between serialization and deserialization.

If we just want to test that that exception is being caught, I can just force some line in the job to throw that exception via Rspec. does that seem fine?

tarebyte:
> If we just want to test that that exception is being caught, I can just force some line in the job to throw that exception via Rspec. does that seem fine?

Yeah that's fine by me, all I want to ensure is that we are rescuing the exception.

nwoodthorpe:
Looks good in prod :+1:

PR #1060, opened 2017-08-04
Perform Repo Setup for GroupAssignments

wilfriedE:
Adds repository setup support for group assignment repos.


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -11,16 +12,31 @@ class GroupAssignmentInvitationsController < ApplicationController
 
   before_action :authorize_group_access, only: [:accept_invitation]
 
-  before_action :ensure_github_repo_exists, only: [:successful_invitation]
+  before_action :ensure_github_repo_exists,    only: %i[setup setup_progress successful_invitation]
+  before_action :ensure_authorized_repo_setup, only: %i[setup setup_progress]
```
nwoodthorpe: We should do the authorized check before the ensure_github_repo_exists check. The github_repo_exists check has the possibility of causing side effects.

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -124,16 +140,43 @@ def organization
   end
   helper_method :organization
 
+  def classroom_config
+    starter_code_repo_id = group_assignment_repo.starter_code_repo_id
+
+    return nil unless starter_code_repo_id
```
nwoodthorpe: you can leave out the `nil`, `return unless starter_code_repo_id` will return `nil`.

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -124,16 +140,43 @@ def organization
...
+
+    client       = group_assignment_repo.creator.github_client
+    starter_repo = GitHubRepository.new(client, starter_code_repo_id)
+
+    ClassroomConfig.new(starter_repo)
```
nwoodthorpe: Can we memoize this? Right now we only call this method once, but in the future if we add another reference, we wouldn't want all these API calls to be made again.

`@classroom_config ||= ClassroomConfig.new(starter_repo)`

app/models/group_assignment_repo.rb
```diff
@@ -7,6 +7,8 @@ class GroupAssignmentRepo < ApplicationRecord
 
   update_index("stafftools#group_assignment_repo") { self }
 
+  enum configuration_state: %i[not_configured configuring configured]
```
nwoodthorpe: Wow this is really cool, I forgot rails 4 introduced this.

PR #1062, opened 2017-08-05
Backup RepoSetup

wilfriedE:
Handle cases where student accepts assignment but does not wait for assignment repo configuration to be completed.  

Schedules a RepoSetup job to be perform an hour later. 
Reschedules job if import not complete.
Configures assignment repo if not already configured.


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -31,7 +31,8 @@ def accept; end
 
   def accept_assignment
     create_group_assignment_repo do
-      if group_assignment_repo.starter_code_repo_id
+      if repo_setup_enabled? && group_assignment_repo.starter_code_repo_id
+        RepoSetupJob.set(wait: 1.hour).perform_later(GroupAssignmentRepo.name, group_assignment_repo.id)
```
nwoodthorpe: Instead of passing a type and an id, you can actually just pass an ActiveRecord object to an ActiveJob and it'll handle serialization and deserialization for you. Then in your job, you can check the type using `#is_a?`

app/jobs/repo_setup_job.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require "concerns/repo_setup.rb"
+
+class RepoSetupJob < ApplicationJob
+  queue_as :repo_setup
+
+  def perform(assignment_repo_type, assignment_repo_id)
+    @assignment_repo = get_assignment_repo assignment_repo_type, assignment_repo_id
```
nwoodthorpe: Can we use parenthesis in the method call here? If you have different opinions let me know, but I find method calls with multiple arguments are kind of hard to follow with no parenthesis'

nwoodthorpe: Though if we pass the object instead of the type and id, we won't need to do this at all

app/jobs/repo_setup_job.rb
```diff
@@ -0,0 +1,36 @@
...
+    if @assignment_repo.github_repository.import_progress[:status] != "complete"
+      RepoSetupJob.set(wait: 1.hour).perform_later(assignment_repo_type, assignment_repo_id)
+    else
+      perform_setup(@assignment_repo, classroom_config) unless @assignment_repo.configured?
+    end
+  end
+
+  private
+
+  def get_assignment_repo(class_name, id)
```
nwoodthorpe: This method will also go away if we pass in an object

app/jobs/repo_setup_job.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require "concerns/repo_setup.rb"
+
+class RepoSetupJob < ApplicationJob
```
nwoodthorpe: If you want to handle the case of serialization failing (the assignment repo was deleted from the db before the job was performed), you can use `rescue_from ActiveJob::DeserializationError`

app/jobs/repo_setup_job.rb
```diff
@@ -0,0 +1,36 @@
...
+  queue_as :repo_setup
+
+  def perform(assignment_repo_type, assignment_repo_id)
+    @assignment_repo = get_assignment_repo assignment_repo_type, assignment_repo_id
+
+    return unless @assignment_repo
+
+    if @assignment_repo.github_repository.import_progress[:status] != "complete"
+      RepoSetupJob.set(wait: 1.hour).perform_later(assignment_repo_type, assignment_repo_id)
```
nwoodthorpe: Are there any cases where import_progress could get stuck in a non complete status? That'd cause this job to run forever.

spec/jobs/repo_setup_job_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  let(:assignment)      { create(:assignment, organization: organization) }
+  let(:assignment_repo) { create(:assignment_repo, assignment: assignment, user: user, github_repo_id: 848) }
+
+  let(:unconfigured_repo) { stub_repository("template") }
+  let(:configured_repo)   { stub_repository("configured-repo") }
+
+  let(:repo_access)  { RepoAccess.create(user: user, organization: organization) }
+
+  let(:grouping)     { create(:grouping, organization: organization) }
```
nwoodthorpe: You don't need to create all of this stuff to create your group assignment. If you don't pass a thing to factorygirl that it needs, it'll create it for you. For the most part, anything you don't reference in your actual specs shouldn't be defined here in a `let`.

spec/jobs/repo_setup_job_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    group.repo_accesses << repo_access
+    GroupAssignmentRepo.create(group_assignment: group_assignment, group: group)
+  end
+
+  before do
+    Octokit.reset!
+  end
+
+  after(:each) do
+    group.destroy
```
nwoodthorpe: Why are we destroying everything after each spec? Rspec will recover the state of the application before the `it` block ran for each spec, they don't influence eachother.

app/jobs/repo_setup_job.rb
wilfriedE: This is a possibility, I am going to hold out on this PR for now and push out another PR with importer things.

spec/jobs/repo_setup_job_spec.rb
tarebyte: @nwoodthorpe I'm pretty sure this is so we don't keep the GitHub team around after the test is done.

nwoodthorpe: Will just a regular `after do` work?

PR #1066, opened 2017-08-07
Add Importable concern

wilfriedE:
Moved some of the importer stuff to GitHubImportable concern and added a few helper methods.


app/models/concerns/github_importable.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module GitHubImportable
```
tarebyte: Since this isn't used with an active record object I'm not 100% sure that this could be a model concern.

WDYT?

nwoodthorpe:
Are we going to be moving all of our importer stuff (even the old stuff) to this new module? Or just your logic for now?

app/models/concerns/github_importable.rb
wilfriedE: I agree, I don't think it should be a model concern. Any suggestions as to where this should go?

nwoodthorpe:
Wait nevermind, I see. You're just extracting it out of GitHubRepository. In that case, yeah I don't think this should be a `concern`, just a regular module.

nwoodthorpe:
This all looks OK to me, I think we should have some specs on the new methods you've added though. They're pretty simple, so the specs should be pretty easy.

spec/models/github_repository_spec.rb
```diff
@@ -150,6 +150,44 @@
       end
     end
 
+    context "importable" do
+      describe "#importing?", :vcr do
```
nwoodthorpe: You're testing the same thing in all 3 of these describes. I'd remove the unrelated expectations from each one. E.g. the #importing describe should only expect against the result of that method.

nwoodthorpe: Then you can have each different status in each one.

wilfriedE: I see what you mean. What I was going for here is to ensure that the other states are also not true.

nwoodthorpe: Right, it doesn't really make sense to test the other methods in the #importing block though. Instead, you can test the importing? method on all 3 cases in the #importing block, and the imported method on all 3 cases in the #Imported block, etc. Then you still get the ensure that the other states aren't true, but it's organized a little better. Does that make sense?

wilfriedE: I see what you mean.

wilfriedE:
Hi @tarebyte, can you let me know what happens when this is deployed? I want to merge this so I can use it with #1062

tarebyte:
:shipit: Done! :shipit:

PR #1096, opened 2017-08-23
Use repository invitations and fix OAuth scopes

tarebyte:
The preview is over https://developer.github.com/changes/2017-08-23-repository-invitation-api/

Which means we have to move 😅 

This PR also updates out look how we inspect OAuth token scopes.


tarebyte:
I'm going to deploy this and then change the rails secret to kick everyone out again 😄

PR #1114, opened 2017-08-29
Make OAuth Scope tree descendent finder work for arbitrarily deep trees

michaelrauh:
This addresses the comments on the `descendents` function within `token.rb` which explains that there is an assumption that OAuth scopes will never exceed three levels deep. It introduces some complexity into the code because it has to recursively traverse the tree, where the old version essentially just finds its place in the tree from three available levels.


michaelrauh:
It looks like hound has found some issues - I will address these.

PR #1115, opened 2017-08-30
Fix Roster Auth

tarebyte:
We didn't add the `OrganizationAuthorization` module to the `RostersController`. So in theory a user could modify another classrooms Roster if they had the proper URL.


app/controllers/rosters_controller.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class RostersController < ApplicationController
-  before_action :ensure_student_identifier_flipper_is_enabled, :set_organization
+  include OrganizationAuthorization
```
ptoomey3: Just to double check...are there any other controllers that should include this but don't currently?

tarebyte: @ptoomey3 nope that was the only one I found, I went through the rest of them when I did this PR.

ptoomey3: 👍 - I figured as much, but just had to ask the question :smile:.

PR #1133, opened 2017-09-05
Remove use of HABTM for Organizations and Users

tarebyte:
There are queries that may be less complex if we queried the join model instead of the relationships.

This PR adds those join models so that they are easily accessible.


d12:
Huh neat, too bad HABTM doesn't optionally expose a model.

:+1:

PR #1137, opened 2017-09-06
Fix incorrect group assignment invitation redirect

### LINKED ISSUE #1136: Create repository success, but show 404 not found
Students in the past few weeks have shown me the following error a few times.

1. I post a link to create a group project
2. Students click on the link, choose group members, and join the group.
3. This is supposed to create the repository.  Instead it sends students to a URL that has the word "success" at the end of it, which is not found, so the students get an "ERROR 404" page.

In fact, the repository was successfully created and everything is fine, but it causes much confusion.

wilfriedE:
resolves #1136 . 

/cc @tarebyte


wilfriedE:
Thanks @tarebyte

PR #1174, opened 2017-10-02
add job to remove a User from an Organization 

arolson:
add job to remove a User from an Organization when they are removed from the GitHub Organization


app/jobs/remove_user_event.rb
```diff
@@ -0,0 +1,14 @@
+# Documentation: https://developer.github.com/v3/activity/events/types/#membershipevent
+class RemoveUserJob < ApplicationJob
+  queue_as :github_event
+  def perform(payload)
+    return true unless payload_body.dig('action') == 'removed'
```
tarebyte: Would you mind using `payload_body["action"] == "removed"` instead since we're only going one level deep?

app/jobs/remove_user_event.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    github_user_id = payload_body.dig('member', 'id')
+    github_organization_id = payload_body.dig('organization','id')
+
+    organization = Organization.find_by(github_id: github_organization_id)
+    organization.users.delete(github_id: github_user_id)
```
tarebyte: Instead of deleting this in the job can you mark this for deletion and send it to the `DestroyResourceJob`?

```ruby
user = organization.users.find_by(id: github_user_id)
return true if user.nil?

user.update_attributes(deleted_at: Time.zone.now)
DestroyResourceJob.perform_later(current_organization)
```

lib/github/web_hook.rb
```diff
@@ -2,7 +2,7 @@
 
 module GitHub
   class WebHook
-    ACCEPTED_EVENTS = %w[ping repository].freeze
+    ACCEPTED_EVENTS = %w[ping repository, membership].freeze
```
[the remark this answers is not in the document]
tarebyte: It's just the comma we can remove here 😄

d12:
Hey @arolson , looks like you still have a failing test. If you need a hand, shoot me a ping. Also there's one outstanding hound comment about a magic comment.

arolson:
@d12 Sorry for the late reply, been catching up on a lot of class work. It looks like I did add the # frozen_string_literal: true into the `app/jobs/remove_user_event.rb`, kind of weird that it didn't pick it up maybe it's in the wrong spot? And yea I will take any help suggestions that you have to make this work 🙇

d12:
Your linter is fine, there's a name error in the travis build though:

```
NameError:
  uninitialized constant RemoveUserJob
# ./spec/jobs/remove_user_event_job_spec.rb:5:in `<top (required)>'
# ------------------
# --- Caused by: ---
# NameError:
#   uninitialized constant RemoveUserJob
#   ./spec/jobs/remove_user_event_job_spec.rb:5:in `<top (required)>'
```

The error is probably coming from the fact that your file name and your class name don't match. In Rails, if they're different, it won't be able to autoload your classes.

`MemberEventJob` will become `member_event_job.rb`. Your spec file will also need to be renamed.

See if that fixes the errors!

arolson:
@d12 Thank you for your help, I actually had to rename it form MemberEventJob => MembershipEventJob :man_facepalming:. 
Good news is it appears to be failing with one error now at: 
app/jobs/membership_event_job.rb
```
1) MembershipEventJob Action removed deletes user from organization
     Failure/Error: user = organization.users.find_by(id: github_user_id)
     
     NoMethodError:
       undefined method `users' for nil:NilClass
     # ./app/jobs/membership_event_job.rb:13:in `perform'
     # ./spec/jobs/membership_event_job_spec.rb:13:in `block (3 levels) in <top (required)>'
```
Out of curiosity Is it because I am using the wrong organization?
I am currently getting the organization this way:
 **organization = Organization.find_by(github_id: github_organization_id)**
Would I be using the one defined here?
app/spec/factories.rb
```
FactoryBot.define do
  factory :assignment do
    organization

    title        { "#{Faker::Company.name} Assignment" }
    slug         { title.parameterize                  }
    creator      { organization.users.first            }
  end
```
Otherwise how would you get the users from the organization, cause that continues to come up nil?

Kadaaran:
Hi @arolson ! Do you need a hand with this PR ? I'd be happy to help ✨

arolson:
Hi @Kadaaran Yea, any help you can provide would be great! Thank you!

app/jobs/membership_event_job.rb
```diff
@@ -0,0 +1,20 @@
...
+class MembershipEventJob < ApplicationJob
+  queue_as :github_event
+  def perform(payload_body)
+    return true unless payload_body["action"] == "removed"
+
+    github_user_id = payload_body.dig("member", "id")
+    github_organization_id = payload_body.dig("organization", "id")
+
+    organization = Organization.find_by(github_id: github_organization_id)
+    user = organization.users.find_by(id: github_user_id)
```
Kadaaran: It should be `user = organization.users.find_by(uid: github_user_id)` In the database uid is the github_id.

app/jobs/membership_event_job.rb
```diff
@@ -0,0 +1,20 @@
...
+    return true if user.nil?
+
+    user.update_attributes(deleted_at: Time.zone.now)
```
Kadaaran: You might have trouble here because there is no deleted_at in the database 🤔

Kadaaran:
Hi @arolson sorry for the delay ! 🙏 
I commented a few parts on your code in the `membership_event_jobs.rb` 🙌 

Maybe you should try something like this:
``` ruby
      github_user_id = payload.dig("member", "id")
      organization = create(:organization, github_id: payload["organization"]["id"])
      user = create(:user, uid: github_user_id, organization_ids: organization.id)
      organization.update(user_ids: user.id)

      MembershipEventJob.perform_now(payload)
      expect { organization.users.find_by(uid: user.uid) }.to raise_error(ActiveRecord::RecordNotFound)
```
Then you will create an organization and a user based on the ones you have in the payload (same paylod that you will run the job with) 

Also I'm not a 100% sure about this so we might ask @d12 or @tarebyte if I'm right but based on what you're trying to do I think that you might be actually removing a `member` for an `organization` when a member leaves a `team` and not an `organization` as MembershipEvent's scope is only `team`  (as you can see on the documentation 📖 )

To make sure you're removing a member for an `organization`, the right event to listen to would be `organization` and the action `member_removed`.(https://developer.github.com/v3/activity/events/types/#organizationevent)
Or you can just change your job so that a member is removed from a team in Classroom as well !

Kadaaran:
Hi @arolson ! 👋 

We really need to push this feature into production as soon as possible, so we pushed your branch into our remote branch in Classroom and finished what you started.
Don't worry we are keeping your work and your commits/contributions !

Thank you for all your work ✨ 

New PR can be found here: #1274

arolson:
✋ @Kadaaran  Thank you for your help, I apologies for not getting to this sooner. I appreciate your help! 🎉

Issue #847, opened 2017-02-11
Sort assignment repositories


janetlndavis:
When viewing the list of repositories associated with an assignment, there is no way to change the sort order. It would be helpful to be able to sort them alphabetically so the order in the Web interface matches the order of the cloned repos in my local filesystem.

janetlndavis:
See, for example, https://classroom.github.com/classrooms/22714409-whitmanswdesignspring2017/group-assignments/project1

danwallach:
Adding to this feature request: it would be handy to be able to sort on multiple fields, such as GitHub ID, lastname/firstname, or our university identifier if that's associated with a given GitHub Classroom instance.

athityakumar:
@tarebyte - I'd like to take up this issue, with ability to sort by name, firstname and lastname. I roughly understand the changes that would be required for this functionality, and will try to get back as soon as possible.

BU-AWolfe:
## MORE FEATURES:
- Folders and hierarchies for student repositories
- compact view with one student repository per line of text
- filter on "overdue" students
- look for pull requests

PR #1182, opened 2017-10-09
Adds feature of sorting assignment repos

athityakumar:
This Pull Request is intended to solve issue #847. By default, the list of assignment repositories is shown in the order of accepting the assignment. Through this Pull Request, I have added the ability to sort the assignment repositories by GitHub name & username of students. Here are a few screenshots of the patch. Suggestions are welcome. 😄 

- Default order
image

- Sorted by GitHub name
image

- Sorted by GitHub username
image


athityakumar:
Updated UI looks like this.

image

jasonlong:
@athityakumar The use of green buttons here probably isn't the best fit. Those are meant to be used for primary actions on a page, ideally with no more than one per page.

Fortunately, we have existing UI patterns for this sort of thing. Here are a couple of examples:

screen shot 2017-10-13 at 3 53 03 pm

screen shot 2017-10-13 at 3 52 32 pm

The second (button style) option is probably a better fit here since the first one works better in a table header.

The [Dropdown component in Primer](https://styleguide.github.com/primer/components/dropdown/) is what you'd want to check out.

athityakumar:
@houndci-bot - Nice catches there. Wanna play fetch, buddy? 😄 

@jasonlong - Thanks for reviewing. As per your suggestion, I've tried to replicate the GitHub's sort button. Definitely, it does "look" much better than the green buttons. I didn't have access to the primer-css dropdown link you had sent. But I was able to go through the CSS and JS aspects of GitHub that made the sort button (`select-menu`) work, and replicate it here as much as possible. Infact, I've made this as a partial so that it can be re-used easily wherever required. 😄 

image
image

Please let me know if anything else that has to be changed.

app/assets/stylesheets/components/select_menu.scss
```diff
@@ -0,0 +1,100 @@
...
+  cursor: pointer;
+  border-bottom: 1px solid #eaecef;
+
+  .selected .octicon-check {
+    color: inherit;
+    visibility: visible;
+  }
+}
+
+.select-menu-modal {
```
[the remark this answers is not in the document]
athityakumar: Ah, the diff fixing this issue is below. So, it isn't being shown as "outdated".

jasonlong:
Nice @athityakumar, that looks much better. :zap:

My only other minor suggestions would be on the dropdown options:

* Probably makes sense to group similar options, so "Student name" followed by "Student username".
* "Time of accepting assignment" may read better as "Assignment acceptance time".

athityakumar:
@jasonlong - Thanks! Regarding grouping of similar options, the dropdown was previously designed to always show the current selected option at the top of the dropdown, and the rest of the options in order. Being a shared partial, I thought that this would be a better way (UX) of showing dropdowns in general. This has been changed. 😄 

PS - I applied for Rails Developer Internship at GItHub, got shortlisted and had the first round of interview scheduled right now. This Pull Request was my attempt at familiarizing with the codebase before sitting down for the interview. But I was just informed an hour ago (before the interview), that a matching candidate has been found and hence I haven't even been given a chance for the interview. 😢

app/assets/javascripts/select_menu.js
```diff
@@ -0,0 +1,38 @@
...
+  function closeSelectMenu() {
+    select_menu_modal.removeClass("active");
+    select_menu_button.removeClass("active");
+    select_menu_button.attr("aria-expanded", "false");
+  }
+
+  $(document).click(function() {
+    closeSelectMenu();
+  });
+});
```
d12: Could you add a newline here?

jasonlong:
Ugh, sorry to hear about the interview situation. Don't give up though, hopefully you'll get another shot down the road!

app/assets/stylesheets/components/select_menu.scss
```diff
@@ -0,0 +1,101 @@
...
+  color: #fff;
+  background: #0366db;
+  /*
+    @note
+
+    The background color is set to this value and not $blue,
+    so as to closely resemble GitHub's other sort buttons.
+
+    Signed off by Athitya Kumar on 15th October 2017
+  */
```
d12: I don't think we need this comment, we have magic hexes in other places in our stylesheet 😛

app/controllers/assignments_controller.rb
```diff
@@ -7,6 +7,9 @@ class AssignmentsController < ApplicationController
   before_action :set_assignment, except: %i[new create]
   before_action :set_unlinked_users, only: [:show]
 
+  SORT_MODES        = ["Assignment acceptance time", "Student name", "Student username"].freeze
+  DEFAULT_SORT_MODE = "Assignment acceptance time"
```
d12: What do you think about changing this to `= SORT_MODES.first`? Then we don't need to have this string written twice.

app/controllers/assignments_controller.rb
```diff
@@ -110,4 +115,18 @@ def send_create_assignment_statsd_events
...
+
+  def sort_assignment_repositories(assignment_repos)
+    @current_sort_mode = params[:sort_assignment_repos_by] || DEFAULT_SORT_MODE
+    @all_sort_modes    = SORT_MODES
+
+    case @current_sort_mode
+    when "Assignment acceptance time"
+      assignment_repos.to_a
+    when "Student name"
+      assignment_repos.sort_by { |repo| repo.github_user.name }
```
d12: This makes an API call to GitHub for every single repo which is kind of unfortunate. There isn't really a better way to do this right now though.

app/controllers/assignments_controller.rb
```diff
@@ -110,4 +115,18 @@ def send_create_assignment_statsd_events
...
     GitHubClassroom.statsd.increment("deadline.create") if @assignment.deadline
   end
+
+  def sort_assignment_repositories(assignment_repos)
+    @current_sort_mode = params[:sort_assignment_repos_by] || DEFAULT_SORT_MODE
+    @all_sort_modes    = SORT_MODES
+
+    case @current_sort_mode
+    when "Assignment acceptance time"
+      assignment_repos.to_a
```
d12: Why do we need `.to_a`?

app/views/assignments/show.html.erb
```diff
@@ -73,12 +73,19 @@
...
-          <% @assignment_repos.each do |assignment_repo| %>
-            <%= render partial: 'orgs/assignment_repos/assignment_repo', locals: { url: organization_assignment_assignment_repo_path(@organization, @assignment, assignment_repo) } %>
-          <% end %>
-        </div>
+        <%= 
+          render partial: 'shared/select_menu',
+            locals: { 
+              title: 'Sort by:',
+              options: @all_sort_modes,
+              option_links: @all_sort_modes.map { |mode| "?sort_assignment_repos_by=#{mode}" },
```
d12: We don't really need to pass this in. We can remove `option_links` from the locals hash, then in `shared/select_menu` line 22, do:

```ruby
link_to "?sort_assignment_repos_by=#{options[i]}"
```

It'd be even nicer if we could generate that link in a view model, but I may be being too picky 😛

d12:
I agree with @jasonlong, on GitHub's other views, we don't change the ordering of sort options when we select something.

screen shot 2017-10-16 at 10 09 52 am

Re: styling: Would you be able to add a bit of padding above the button so it isn't pressed against the divider?

app/views/assignments/show.html.erb
athityakumar: Having select-menu as a shared partial, I was thinking of re-using this like `Save as: ` button with exporting feature for multiple formats like CSV, XLS, JSON, etc. as requested in #1184. Hence, I provided a way to generally redirect to other links. Let me know if this has to be changed. 😄

d12: Ah right on. Makes sense to me :+1:

athityakumar:
@d12 - Thanks for the review. I've added the suggested changes (except the shared_partial one). 😄 

Regarding the interview, yes - I'd definitely like to give another shot for GitHub next year. But unfortunately, I'll be graduating next year and will have to sit for placements rather than internships. This internship is something very ideal for me, that I've just missed despite my best intention. 😢

jasonlong:
@d12 

> Re: styling: Would you be able to add a bit of padding above the button so it isn't pressed against the divider?

I'd argue that the divider isn't even necessary here.

athityakumar:
@jasonlong - The divider does, divide the invitation part from the list of assignments being shown. So, it kinda makes sense for me to have a divider. However, let me know if the divider has to be removed.

jasonlong:
@athityakumar Yep, I get that you want to separate those two sections, but a line isn't always needed for that. Might want to try having some extra whitespace below the big input box at the top and the new sort button. The sort button appearing closer to the listings will make that relationship clear and will also visually separate the two sections.

athityakumar:
@jasonlong - Acknowledged. I've removed the divider UI that used to separate the invite link and the list of assignment repositories.

athityakumar:
@jasonlong @d12 - I've removed the divider UI that used to separate the invite link and the list of assignment repositories. Anything else that has to be changed? 😄

jasonlong:
:thumbsup: on my end

app/assets/javascripts/select_menu.js
athityakumar: Added 👍

d12:
Code LGTM. I'm going to pull it down locally later tonight, then I'll merge if all seems fine. Thanks for working on this :)

athityakumar:
@d12 - Sure, welcome. 😄 

The failing Travis CI build seems to be because of `KNAPSACK_GENERATE_REPORT=true ./bin/rake knapsack:rspec` command.

d12:
One more thing, would you be able to abstract this into something that works for both assignments and group-assignments? After that, we'd be happy to merge

Kadaaran:
Hi @athityakumar 👋 Do you need a hand with this ? Feel free to ask if you need help ✨

Kadaaran:
Hi @athityakumar 👋 Do you need a hand with this ? Feel free to ask if you need help ✨

athityakumar:
Hey @Kadaaran  👋

Sorry, I had just been caught up with other stuff and hadn't been able to work on the final touches. I'll try to work on this, in the coming weekend. 😃

Kadaaran:
Hi @athityakumar 👋 No worries ! Feel free to ask if you need help.

Kadaaran:
@athityakumar Friendly bump here ✋

peterdickx:
is this still being worked on?

BenEmdon:
This was an open source contribution that seems to have gone stale. @peterdickx if you're interested feel free to pick it up ✨

PR #1200, opened 2017-10-20
Revert "Remove use of HABTM for Organizations and Users"

d12:
Reverts education/classroom#1133

We've been getting exceptions, signup is broken right now.


d12:
I don't think we need this anymore.

PR #1210, opened 2017-11-09
Update travis config

PeterDaveHello:


d12:
@PeterDaveHello could you give some context? What's the purpose of this change?

PeterDaveHello:
@d12 sorry, I think I forgot to say that I mention them in the commit message, would that be enough?

d12:
Oh no worries, I see now. I'm not sure if theres a reason we're on this specific version. cc @tarebyte

PR #1218, opened 2017-12-04
added zh-CN locale files

yuxisteem:
Hi I just added zh-CN locale files for Simplified Chinese translation. Let me know if any more changes are needed. thanks


d12:
Hey @yuxisteem ,

Thanks for the pull request! We're going to need to do a bit of work checking the translations which may take some time. We'll get back to you if we have questions :)

PR #1228, opened 2018-01-11
[WIP] Pagination for assignment repos with rosters

d12:
A certain intern forgot to add pagination for the new assignment repo view 🙈

Also, the way we were fetching assignment repos and entries wasn't going to work with pagination (The pagination helpers require AR associations, but I was converting to an array then doing filtering/ordering.

So this PR changes a couple of things:

1. Remove RosterEntrySorter - No longer needed
2. Move the sorting into an `order` clause on the AR relation. This needs to be done on the relation, otherwise pagination won't work properly.
3. Add pagination controls for both tabs of the new assignment view
4. Give each tab it's own pagination control key. Paginating one tab shouldn't advance the other tab.
5. Move some other logic out of the view that really never should have been there

TODO:
1. ~~Some styling stuff~~
2. When we change pages on the second tab, the reload causes us to change back to the first tab. If you switch to the second tab again, the page will have changed, but the second tab should load automatically.
3. ~~Some code style stuff, like moving a lot of this stuff out of the controller.~~
4. ~~Tests the new ordering. I should be able to reuse the old tests from the sorting service.~~

cc @tarebyte


d12:
@tarebyte everything about this is good to go except the bug where changing page while on the second tab brings you back to the first tab, but does successfully change page. My javascript is super rusty, I'm not sure what the best way to solve is. Any ideas? I can give more info if you need.

d12:
screen shot 2018-01-12 at 2 52 49 pm

tarebyte:
@d12 best thing I can think of is that we add something to the URL on tab change so that we can keep track of the state. I'd have to look up how to do that again though 😆

d12:
Yup that was my idea, but apparently changing the URL in javascript forces a page reload unless you use a specific API that's in the new browsers. Not sure if we want to be going that route, I can look into it though.

d12:
I went ahead and used cookies, maybe overkill but it works 🤷‍♂️ 

I'll leave a couple comments on the code explaining why things are written how they are. I haven't written JS since my internship though, I might be doing everything wrong :laughing:

cc @tarebyte 

http://recordit.co/9FG8h9YBbw

(The page loads on the gif are crazy slow because local :/ )

app/assets/javascripts/roster_tabs.js
```diff
@@ -10,11 +10,25 @@ function selectTab(element){
...
     unlinkedSpan.classList.remove("hidden-tab");
 
     studentsTab.classList.remove("selected");
     unlinkedTab.classList.add("selected");
+
+    document.cookie = "unlinkedSet=true; path=/";
+  }
+}
+
+function setTabOnLoad(){
```
d12: This function runs every time we reload the partial (turbolinks, and also when a hard refresh is made on the page).

app/views/assignments/show.html.erb
```diff
@@ -89,3 +92,13 @@
     <% end %>
   </div>
 </div>
+
+<script>
```
d12: This script runs every time the partial loads. On a hard refresh, `roster_tab.js` hasn't loaded yet so we need to check to see if the function exists. On a turbolinks load, the script does exist.

tarebyte: Let's extract it to it's own partial cause we need it for the group assignments show page as well.

d12: Right, group assignments exist -_-

d12:
Alright thats assignments done, I'll do group-assignments later today or tomorrow, the tabs are a little different.

d12:
Looks like one tab of group assignments already has pagination. I just need to do the other tab, and then the logic to make sure changing page on the second tab doesn't switch to page 1.

d12:
screen shot 2018-01-19 at 3 04 53 pm

d12:
Alright, all done :+1:

http://recordit.co/QK9JZDg0cN

I'm travelling most of Monday, but we can look into shipping this next week.

PR #1233, opened 2018-01-17
Features rake task

Kadaaran:
Hi there !

So I'm trying to understand Classroom's code better and to achieve that I created a basic rake task that enables all features for staff users.

I'm not sure with the right way to do it, especially for the users.
Also do the rake task need to be run during the `./script/server` script or do we just add it into the README ?

Thanks for your help !

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,10 @@ And another terminal with:
 script/ngrok
 ```
 
+Then enable features with:
+
+```rake enable_features```
+
 That's it! You should have a working instance of GitHub Classroom located [here](http://localhost:5000)
 
 ## Deployment
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ And another terminal with:
 script/ngrok
 ```
 
-You might want some features to be enabled, if it is the case then just run:
+To enable optional features, run:
 
 ```rake enable_features```
 
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ And another terminal with:
 script/ngrok
 ```
 
-Then enable features with:
+You might want some features to be enabled, if it is the case then just run:
 
 ```rake enable_features```
 
--- a/lib/tasks/enable_features.rake
+++ b/lib/tasks/enable_features.rake
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
-desc 'Enable roster managment page or all features ?'
 task enable_features: :environment do
-
-  # get staff users or first user
-  # grant access to this user to get roster page
-
-  # do it for all or just one
-  # add to readme
+  User.all.each do |user|
+    if user.staff?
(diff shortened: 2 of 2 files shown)


lib/tasks/enable_features.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+task enable_features: :environment do
+  User.all.each do |user|
```
tarebyte: So technically this script isn't enabling this feature for the user you listed, it's enabling the feature for everyone for each user you have in your database.

You could just enable them all without the loop or if you want to enable the feature for just `staff` which we have defined as a group https://github.com/education/classroom/blob/44efed0852076877840d2dcd387389f165e0eaa5/config/initializers/flipper.rb#L21-L24

We could just make a call that enables it for the group. I would check out https://github.com/jnunemaker/flipper/blob/master/examples/enabled_for_actor.rb

Kadaaran: Great thanks I'll check the documentation !

README.md
```diff
@@ -230,6 +230,10 @@ And another terminal with:
 script/ngrok
 ```
 
+Then enable features with:
```
d12: I'd phrase this to make it sound optional, instead of mandatory. Production is only running with 2 of 4 features enabled, so a developer might not want to enable all the features.

lib/tasks/enable_features.rake
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+task enable_features: :environment do
+  GitHubClassroom.flipper[:team_managment].enable_group :staff
+  GitHubClassroom.flipper[:repo_setup].enable_group :staff
+  GitHubClassroom.flipper[:student_identifier].enable_group :staff
```
d12: There's one more! `:deadlines`

Kadaaran: Thanks ! But why isn't `:deadlines` in `feature_flags_dependency.rb` as the others ? 🤔

d12: Oh whoops, sorry thats my fault. The `:deadlines` flipper is still on production but the code was actually removed a long time ago. These 3 are fine :+1:

Kadaaran: It's alright ! Thanks for the information !

README.md
```diff
@@ -230,6 +230,10 @@ And another terminal with:
 script/ngrok
 ```
 
+You might want some features to be enabled, if it is the case then just run:
```
Kadaaran: @d12 What do you think of this part ? I might need some help with English with this one 😄

d12: How about `To enable optional features, run:`? 😄

Kadaaran: @d12 Oh. Yep. This one is more simple. Thanks and sorry 😅

PR #1237, opened 2018-01-19
Roster form validation

Kadaaran:
Hi there ! 👋 

I found a bad user experience while creating a roster.
The form validation is in server side, so if the user did not fill one field out of the two required, all data is lost (as the page reloads after hitting the submit button).

screen shot 2018-01-17 at 23 40 33
screen shot 2018-01-17 at 23 40 40

It can be solved by just adding `required: true` into the HTML but we lose the custom error messages. The data will be sent to the controller only if both fields aren't empty.
In order to keep those error messages we can add some JS but I don't know if it's a good thing to do.

screen shot 2018-01-19 at 11 30 55

Does someone have a thought about it ? 🤔 

Thanks a lot ! 🙇

### DIFF
--- a/app/views/orgs/rosters/new.html.erb
+++ b/app/views/orgs/rosters/new.html.erb
@@ -23,13 +23,11 @@
         <dl class="<%= view.form_class_for(:identifier_name) %> short-field">
           <dt><label>What information would you like to use to identify your students?</br> (emails, student ID #, etc.)</label></dt>
           <dd><%= text_field_tag(:identifier_name, '', class: 'form-control input-block', required: true) %></dd>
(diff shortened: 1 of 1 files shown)


tarebyte:
Honestly I'm fine with losing the custom error message for it not being blank if that means we don't have to make a round trip to the server and save people time.

Kadaaran:
@tarebyte Ok then I removed from the form the custom error messages but as we use them elsewhere, I'm leaving them in `FormView`. It might not be "pretty" for the user but indeed we don't make a round trip to the server side 👍

tarebyte:
@Kadaaran I know we technically don't need them, but just for consistency sake would you mind leaving the custom validation partials in?

That way if we add different validations that aren't just presence we don't have to go back and add them in.

Kadaaran:
@tarebyte Sure ! (I did not think about that 🙈 )

Issue #1216, opened 2017-11-22
Number of group members wrong


drbeco:
(feature-request, but also a _bug_ imho)

It should be important for classroom to keep updated with the real group teams after they are created.

For now I have a group assignment where one of the groups had a problem and a member simply left the "team", the other I removed.

Now the assignment states there are 8 members, while the team sees only 6.

This prevented another member to join, since the limit was seven (and I solved by raised it to 8).

Whenever I had this problems before, I asked an admin here ( @tarebyte ) to help and he was always so kind to do so. This time I didn't want to bother him yet again with such problems to be "manually" handed. So I ended up with this solution that breaks the coherence of the system.

It would be very nice if we could manage classroom (or it be aware of the changes) by simply and only going to our github account and manage teams there. I reckon classroom has reading rights to see the teams, how it goes, what it changed, and update itself on that base, right? 

It does not have to be "instantaneous", mind you. It may be a daily cron-job, or half-a-day, or two days, no problem. That depends on how you see the demand for this task would increase the load on the server. But still, a nice feature to have, if not a bug to solve.

Thanks!

---

(edited: typo)

Kadaaran:
Hi @drbeco  , thank you for reporting this issue.

I've been trying to find a fix for this issue this week and I think I have found how to do so.

I'll try to ship a first try this week 🚀

drbeco:
Merci

Kadaaran:
Hi @drbeco  👋 

We finally merged the PR #1254 that should fix this issue 🚀 
Next one will be #1274 

Thank you for again your patience !

Kadaaran:
Merged #1274 & #1254

drbeco:
Great news, thank you and all the team for the effort.

Issue #1215, opened 2017-11-20
I cannot seem to fully remove stranger member from an assignment team


bvelez:
I was able to remove team member AmbujaAK from team:

https://github.com/orgs/UPRM-CIIC4010-F17/teams/tem-240db.

![image](https://user-images.githubusercontent.com/5605641/33033548-199059ea-cdfb-11e7-85f9-105a92166161.png)

But it still appears in the list of teams in the assignment screen:

![image](https://user-images.githubusercontent.com/5605641/33033506-edefc2c6-cdfa-11e7-826b-a9c51f92c195.png)

The assignment does not allow the other team member to join the team

This person is not part of my classroom. I have no idea how he managed to join the team.

drbeco:
This request seems to be related to my new issue (feature-request), so I'm linking here ( #1216 ) for organization. Please take a look if that would address this matter. Thanks.

PR #1241, opened 2018-01-25
[WIP] Synchronize users removed from teams

Kadaaran:
Hi folks ! 👋 

First fix for the issue #1216 / #1215 
TODO:
- [ ] Write tests
- [ ] Removed from organization as well ?

It seems Classroom does not accept some events, especially when a user is removed from an organization or leaves a team in GitHub.
The thing is that in Classroom the number of members was wrong all this time (the informations were not updated at all).

I found a way to fix it. Not sure if it's a good way to do it but it works.

Next step would be to do the same when someone leaves an organization or is removed by an admin. 
I realized after I've finished it that someone was already working on the organization/teams issue #1174 🙈 

screen shot 2018-01-25 at 15 13 51
screen shot 2018-01-25 at 18 10 57

Happy to hear your comments about it ✨

### DIFF
--- a/app/jobs/membership_event_job.rb
+++ b/app/jobs/membership_event_job.rb
@@ -9,14 +9,14 @@ class MembershipEventJob < ApplicationJob
 
     user_id = payload_body.dig("member", "id")
     team_id = payload_body.dig("team", "id")
+
     user = User.find_by(uid: user_id)
-    group_id = Group.find_by(github_team_id: team_id).id
+    return true unless user.present?
 
-    repo_accesses = GroupAssignmentRepo.find_by(group_id: group_id).repo_accesses
+    group = Group.find_by(github_team_id: team_id)
+    return true unless group.present?
 
-    repo_access = repo_accesses.find_by(user_id: user.id)
-    repo_access&.delete
-
-    # Do we remove github_repo as well ?
+    group.repo_accesses.find_by(user_id: user.id)
+    repo_access&.destroy
   end
 end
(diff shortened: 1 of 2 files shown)


app/jobs/membership_event_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def perform(payload_body)
+    return true unless payload_body.dig("action") == "removed"
+
+    user_id = payload_body.dig("member", "id")
+    team_id = payload_body.dig("team", "id")
+    user = User.find_by(uid: user_id)
+    group_id = Group.find_by(github_team_id: team_id).id
+
+    repo_accesses = GroupAssignmentRepo.find_by(group_id: group_id).repo_accesses
```
tarebyte: So `Group` has `repo_accesses` directly https://github.com/education/classroom/blob/44efed0852076877840d2dcd387389f165e0eaa5/app/models/group.rb#L13-L14

So you could just do:

```ruby
def perform(payload_body)
  return true unless payload_body.dig("action") == "removed"
  
  user_id = payload_body.dig("member", "id")
  team_id = payload_body.dig("team", "id")

  user  = User.find_by(uid: user_id)
  return true unless user.present?

  group = Group.find_by(github_team_id: team_id)
  return true unless group.present?

  group.repo_accesses.find_by(user_id: user.id)
  repo_access&.destroy
end
```

Kadaaran: I get it now thanks ! 👍

Kadaaran:
Duplicate of #1174

Kadaaran:
Duplicate of #1174

Issue #1239, opened 2018-01-25
Sort ids in roster alphabetical for first log-in


amueller:
When a student first logs in, they are shown a roster to select their ID from.
That roster is not sorted alphabetically, which makes it really painful for users to find their ID.
It should probably be easy to sort it, and that would make students life much easier.

I'm using student IDs, which is usually two characters followed by 4 numbers. Imagine scrolling through 150 of those trying to find yours. And that's not a big class.

ashleytqy:
I'd like to contribute to this issue, if possible!

Kadaaran:
@ashleytqy Sure, go ahead ! ✨ In the meantime feel free to ask if you have any questions.

PR #1242, opened 2018-01-26
Sort roster entries based on identifier

ashleytqy:
Based on this: https://github.com/education/classroom/issues/1239

The roster shown to a student when they first log in to choose their IDs is now **sorted** ✨ !

screen shot 2018-01-25 at 8 21 58 pm


d12:
Thanks for the submission @ashleytqy !

I pulled this down locally and it looks good to me. In the future we might want more customizable search but this seems :100: for now.

Issue #1243, opened 2018-01-26
Problems running the server locally.


Losnen:
I'm trying to run the server locally for contribute but I have some problems with elasticsearch after the authorization with github.

got this in the terminal

`17:44:53 web.1           | 2018-01-26 17:44:53 +0000: Rack app error handling request { GET /auth/github/callback }
17:44:53 web.1           | #<Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [string] declared on field [group_title]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [group_assignment_repo]: No handler for type [string] declared on field [group_title]","caused_by":{"type":"mapper_parsing_exception","reason":"No handler for type [string] declared on field [group_title]"}},"status":400}>
`

and got this on the browser

`Puma caught this error: [400] {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [string] declared on field [group_title]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [group_assignment_repo]: No handler for type [string] declared on field [group_title]","caused_by":{"type":"mapper_parsing_exception","reason":"No handler for type [string] declared on field [group_title]"}},"status":400} (Elasticsearch::Transport::Transport::Errors::BadRequest)`

Can you help me please?

Kadaaran:
Hi @Losnen did you have any trouble while running `./script/setup` ?

Losnen:
`$ ./script/setup 
==> Bootstrapping Homebrew OSX environment
==> Installing gem dependencies…
Using rake 12.0.0
Using concurrent-ruby 1.0.5
Using i18n 0.8.6
Using minitest 5.10.3
Using thread_safe 0.3.6
Using tzinfo 1.2.3
Using activesupport 5.1.3
Using builder 3.2.3
Using erubi 1.6.1
Using mini_portile2 2.2.0
Using nokogiri 1.8.0
Using rails-dom-testing 2.0.3
Using loofah 2.0.3
Using rails-html-sanitizer 1.0.3
Using actionview 5.1.3
Using rack 2.0.3
Using rack-test 0.6.3
Using actionpack 5.1.3
Using nio4r 2.1.0
Using websocket-extensions 0.1.2
Using websocket-driver 0.6.5
Using actioncable 5.1.3
Using globalid 0.4.0
Using activejob 5.1.3
[...]

PR #1251, opened 2018-02-01
Use Docker Compose for dependencies instead of Homebrew services

tarebyte:
We've had plenty of setup issues in the past such as https://github.com/education/classroom/issues/1243 and it's really difficult to know what you need in order to get classroom working since there are so many parts.

I'd like to propose that we move the project from using Homebrew services to using Docker Compose for our external dependencies.

**What does this mean?**
All it means is that instead of using brew services and pointing our app to the default ports for each respective service we turn them on using docker-compose and then point them to a custom port we have configured in `docker-compose.yml`

**How does this change my workflow?**
Hopefully it really shouldn't apart from the fact that you now need to have docker installed on your computer.

All of the files in `script` have been or will be updated to use docker-compose.

**Why not just fully containerize it**
I thought about it, and we've had Pull Requests in the past to do this. But I really didn't like how much it interfered with my development flow using things like pry or running rails commands that we had to prefix `docker-compose` on to the command.

Hopefully this is a happy medium

**What can I do to help?**
I'm glad you asked! I'd love for folks to pull this down and see if it works for them, and if not leave feedback so that I can get it working.

## TODO
- [x] Get tests working with new configuration
- [ ] Get feedback from active contributors whose workflow I would mess up 😄 
- [x] Update the README and other related documents to reflect the new world order

<br><br>
💖 **Thanks** 💖


trombonekenny:
@tarebyte Did you see https://github.com/mdwheele/classroom-docker by @mdwheele ? That seemed pretty useful to me for wrapping Classroom in docker-compose. Of course, anything that changes in Classroom to support that better is even nicer!

tarebyte:
> Did you see https://github.com/mdwheele/classroom-docker by @mdwheele ? That seemed pretty useful to me for wrapping Classroom in docker-compose. Of course, anything that changes in Classroom to support that better is even nicer!

I hadn't, that would've been nice to know when I started 😅

tarebyte:
@Kadaaran it looks like your issue is that your node install didn't work, maybe retry install `nodejs`?

talsafran:
Just did a fresh run of the codebase with docker and worked! 🐳

Issue #603, opened 2016-05-26
Group Assignment - Minimum number of members


drbeco:
Following the same line as #288 this is also a feature request.

Please, consider to add an optional field in the settings during creation for the minumum number of members per team.

If edited, only in the direction to open the range (that is, allowing a lower minimum).

What this setting would do: groups that are not in the range (minimum-maximum) would be **froze** until they grab the number of members or the range change to allow the current number of members.

Also, members should be able to move from one group to another as they wish (or at least if there is another option that the teacher set allowing the move -- this can be another feature request in itself, but it is here because it is just a simple statement on how this could work).

johndbritton:
> What this setting would do: groups that are not in the range (minimum-maximum) would be froze until they grab the number of members or the range change to allow the current number of members.

The way Classroom works, we create the repository for the group when the first member accepts an assignment invitation. There's no real way for us to enforce that a group have a minimum number of people.

Do you have some example use cases where this feature would be helpful? Are there any other teachers using Classroom that want this?

I think this is relatively low priority for us if we decide to do it. I'm not convinced it's needed but we won't rule it out entirely. Let's see if other people request the same.

drbeco:
Hello @johndbritton ,

I can't tell on behalf of other teachers, and I hope more teachers comment.

An example, in my case, would be the recurring situation we have when we tell the whole class that there is a task to be done in group, lets say, group of 5, and they ask, of course:

- Teacher, can we work in pairs?
- No, group of five please.
- But teacher, what if there is not enough students?
- Ok, groups of five or six, just to balance, but preferred of five.
- Teacher, can we have groups of 4?
- No, five or six please.
- Teacher, I don't want to work in group, can I work alone?
- No, minimum 5, maximum 6.
- Err, teacher, what if we had a group of 5 and one drop the course?
- That is ok, finish with four.
- Teacher, can we work in couples?
...

And so on... 

I think you got the idea! :laughing:

johndbritton:
Yeah, definitely understand the request. I think we'll keep this at low priority for now given the feedback we've received on the app. If other folks think this is important, please speak up.

drbeco:
Hey @tarebyte , @tabinks , @cyhsutw , @dinever and @johndbritton , give me some support here! Need you! :smile: 

I rather use the only **minimum** than the maximum! Because if I set only the minimum, and I only invite students in a class that obviously have a limited number of sits, I can with only one number set also the approximated number of groups the class would be divided into!

That is really hand!

[...]

PR #1254, opened 2018-02-02
Create Membership Event Job

Kadaaran:
What happens now: When a member is removed from a team, Classroom still displays this member and do not let users join the team(because it appears to be full in some cases).

What it changes: Now we display the right numbers of members ✨ 

35402000-7faac1aa-01fb-11e8-804d-4606498a9e7a

35402001-7fe0fcc0-01fb-11e8-935c-ececee039496

Why I need help: I think Travis is failing because when I run RSPEC locally and I create  `RepoAccess`, it tries to add the user I created from the payload to my organization and apparently the API is not happy with that and returns a `401` but for now I don't know why. I'm still trying to figure out why but if someone has an idea I'd be happy to hear more about it ✨ 

screen shot 2018-02-02 at 23 06 37

To do next: 
- [x] Same thing with people being removed from `organization` 👍 
- [x] Fix this massive dependencies in the rspec file 
- [x] Bug fixed thanks to @d12 🍰


app/jobs/membership_event_job.rb
```diff
@@ -0,0 +1,22 @@
...
+  queue_as :github_event
+
+  def perform(payload_body)
+    return true unless payload_body.dig("action") == "removed"
+
+    user_id = payload_body.dig("member", "id")
+    team_id = payload_body.dig("team", "id")
+
+    user = User.find_by(uid: user_id)
+    return true if user.present?
```
koenyskout: I think the switches between if/unless and blank/present reversed the logic of this line (and the one for group below).

Kadaaran: But it still works though 🤔

fjorgemota:
Well...so ready for review? (just to remember to change the labels)

Kadaaran:
@fjorgemota Not yet, just a little change and we will be good to change for review 👍

fjorgemota:
Guys, Sorry to bother again, but...is there any plans to review this?

I'm going to use Github Classroom on a class this semester and this PR is very important to fix problems where a student join a wrong group.

Thanks.

d12:
Hey @fjorgemota ,

We're ramping up some more engineers to work on this project, but for now we don't have a ton of bandwidth. We'll get this shipped as soon as we can.

> I'm going to use Github Classroom on a class this semester and this PR is very important to fix problems where a student join a wrong group.

Are you saying there's a student who's joined the wrong group? Or you're trying to prevent that from happening in the future? If something's gone wrong with your Classroom, can you email `education@github.com`? We can try to get you sorted while you wait for this fix. Thanks :)

app/jobs/membership_event_job.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Documentation: https://developer.github.com/v3/activity/events/types/#membershipevent
+class MembershipEventJob < ApplicationJob
+  queue_as :github_event
+
+  def perform(payload_body)
+    return true unless payload_body.dig("action") == "removed"
```
d12: We probably don't need `dig` here, we can just do `payload_body["action"]`

fjorgemota:
Hey @d12! Nice to know that Github Classroom will receive more attention! 

About your question, I'm trying to prevent that from happening on the class this semester, because for now the use of Github Classroom will be just a test with a class where i'm a teaching assistant, to see if it helps with the verification of the activities (just by having a simple `git diff` I think it will), and I did not know that you're available to receive small problems manually on the database (I mean, I think that this problem is not too frequent, but it's good to know anyway).

About the test, I think that this is the most import important issue we detected right now that may affect us. We will use Github Classroom together with the Moodle of the university (not a direct integration, tho) so for us #603 will not be a big problem right now (because we will request the student to enter a limited group on Moodle (which have these restrictions) and on the same group on Github Classroom), so it's not a big problem right now.

Anyway, here's a "how-to document" that me and the teacher developed to allow students to use git and github classroom more easily: https://github.com/nanvix/teaching/tree/master/howtos/github-classroom (it's in Portuguese, but ... okay)

Thanks and keep the good work, guys. =D

d12:
@fjorgemota Ah I see. We're done with reviewing now, so this will get deployed likely sometime today or tomorrow. :)

Glad to hear you're trying out Classroom for your course, if you run into any problems or any features you think would help you, we take feature requests [here](https://github.com/education/classroom/issues/new)!

fjorgemota:
@d12 Nice! Thanks for the support, and like I said before, sorry for bothering. =)

app/jobs/membership_event_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    user_id = payload_body.dig("member", "id")
+    team_id = payload_body.dig("team", "id")
+
+    user = User.find_by(uid: user_id)
+    group = Group.find_by(github_team_id: team_id)
+
+    return true unless group.present? && user.present?
+
+    repo_access = group.repo_accesses.find_by(user_id: user.id)
+    repo_access&.destroy
```
tarebyte: So I don't think we want to delete the repo_access here, just the fact `has_and_belongs_to_many` relationship exists.

So it would be something like 

```ruby
repo_access = group.repo_accesses.find_by(user_id: user.id)
group.repo_accesses.delete(repo_access)
```

PR #1265, opened 2018-02-06
Fix for pagination bug

d12:
Apparently `IN ()` is invalid syntax in postgres and none of my specs caught that -_-

Instead of `()`, we use `(NULL)` as the empty collection, which works in both cases.

I also added a regression test.

@tarebyte

### DIFF
--- a/app/models/roster_entry.rb
+++ b/app/models/roster_entry.rb
@@ -16,7 +16,7 @@ class RosterEntry < ApplicationRecord
   # with a secondary sort on ID to ensure ties are always handled in the same way
   def self.order_for_view(assignment)
     users_with_repo = assignment.repos.pluck(:user_id)
-    sql_formatted_users = "(#{users_with_repo.join(',')})"
+    sql_formatted_users = users_with_repo.empty? ? "(NULL)" : "(#{users_with_repo.join(',')})"
 
     order <<~SQL
       CASE
@@ -31,7 +31,7 @@ class RosterEntry < ApplicationRecord
   # Restrict relation to only entries that have not joined a team
   def self.students_not_on_team(group_assignment)
     students_on_team = group_assignment.repos.map(&:repo_accesses).flatten.map(&:user).map(&:id).uniq
-    sql_formatted_students_on_team = "(#{students_on_team.join(',')})"
+    sql_formatted_students_on_team = students_on_team.empty? ? "(NULL)" : "(#{students_on_team.join(',')})"
 
     where <<~SQL
       roster_entries.user_id IS NULL OR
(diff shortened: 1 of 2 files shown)


d12:
Whoops. I pushed them together so we didn't get the build for the failing test. This is the failing spec:

```
➜  classroom git:(fix_for_pagination_bug) ✗ rspec spec/models/roster_entry_spec.rb
DEPRECATION WARNING: ActiveSupport.halt_callback_chains_on_return_false= is deprecated and will be removed in Rails 5.2. (called from <top (required)> at /Users/d12/dev/classroom/config/initializers/new_framework_defaults.rb:25)
Knapsack time offset warning enabled!
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}

Randomized with seed 50176

RosterEntry
  associations
    can have a user
    must have a roster
    can have no user
  order_for_view
    orders correctly (FAILED - 1)
    orders correctly, even with no accepted students

========= Knapsack Time Offset Warning ==========
Time offset: 30s
Max allowed node time execution: 01m 07s
Exceeded time: -01m 07s

Global time execution for this CI node is fine.
Happy testing!

Need explanation? See FAQ:
https://github.com/ArturT/knapsack#faq
=================================================
Sign up for Knapsack Pro here:
http://knapsackpro.com
=================================================

Knapsack global time execution for tests: 0.40532994270324707s

Failures:

  1) RosterEntry order_for_view orders correctly
     Failure/Error: expect(RosterEntry.where(roster: roster).order_for_view(assignment).to_a).to eq(expected_ordering)

     ActiveRecord::StatementInvalid:
       PG::SyntaxError: ERROR:  syntax error at or near "4"
       LINE 3:   WHEN roster_entries.user_id IN 4) THEN 0 /* Accepted */
                                                ^
       : SELECT "roster_entries".* FROM "roster_entries" WHERE "roster_entries"."roster_id" = 4 ORDER BY CASE
         WHEN roster_entries.user_id IS NULL THEN 2                   /* Not linked */
         WHEN roster_entries.user_id IN 4) THEN 0 /* Accepted */
         ELSE 1                                                       /* Linked but not accepted */
       END
       , id
     # ./spec/models/roster_entry_spec.rb:48:in `block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # PG::SyntaxError:
     #   ERROR:  syntax error at or near "4"
     #   LINE 3:   WHEN roster_entries.user_id IN 4) THEN 0 /* Accepted */
     #                                            ^
     #   ./spec/models/roster_entry_spec.rb:48:in `block (3 levels) in <top (required)>'
```

PR #1274, opened 2018-02-15
Remove user from Organization

Kadaaran:
This PR is intended to fix bugs due to the fact that we don't keep organization updates in Classroom.

When a member is removed from an organization in GitHub, it is removed from Classroom as well 🎉 

Thanks to @arolson for all of his work in the previous PR #1174 👏 

TODO:
- [x] Transfer assignments to another owner
- [x] Write tests for assignments ownership transfer


app/jobs/organization_event_job.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    github_user_id = payload_body.dig("membership", "user", "id")
+    github_organization_id = payload_body.dig("organization", "id")
+    organization = Organization.find_by(github_id: github_organization_id)
+    user = organization.users.find_by(uid: github_user_id)
+
+    return true unless user
+
+    organization.users.delete(user)
+    organization.save
```
tarebyte: Since this is a `has_and_belongs_to_many` relationship, the organization doesn't need to be saved here

app/jobs/organization_event_job.rb
```diff
@@ -0,0 +1,20 @@
...
+    return true unless payload_body.dig("action") == "member_removed"
+
+    github_user_id = payload_body.dig("membership", "user", "id")
+    github_organization_id = payload_body.dig("organization", "id")
+    organization = Organization.find_by(github_id: github_organization_id)
+    user = organization.users.find_by(uid: github_user_id)
+
+    return true unless user
+
+    organization.users.delete(user)
```
tarebyte: We should probably think about what the behavior should be if the user in question was the author of an assignment.

tarebyte:
> After a student leaves, is their work still visible in Classroom?

This only applies to teachers since a student doesn't belong to our version of an organization.

app/jobs/organization_event_job.rb
Kadaaran: Can we then update the `creator_id` with the owner id of the organization ?

Kadaaran:
> Also will deploying automatically subscribe to the new hook? Or will we need to re-auth teachers?

The new hook ? For repositories ?

tarebyte:
> Also will deploying automatically subscribe to the new hook? Or will we need to re-auth teachers?

@d12 when we create a hook for organizations we ask for every event, so this should just start working as soon as a we deploy

app/jobs/organization_event_job.rb
Kadaaran: We could also keep the author name and precise that the author is no longer part of the organization.

tarebyte: Yeah it would be like a transfer

app/jobs/organization_event_job.rb
```diff
@@ -0,0 +1,32 @@
...
+    @user = @organization.users.find_by(uid: github_user_id)
+
+    return true unless @user
+
+    transfer_assignments if user_owns_assignments?
+    @organization.users.delete(@user)
+  end
+
+  def user_owns_assignments?
+    @organization.all_assignments.map{ |a| a.creator_id }.include? @user.id
```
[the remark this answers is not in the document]
tarebyte: This could just be `@organization.all_assignments.map(&:creator_id).include?(@user.id)`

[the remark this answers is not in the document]
Kadaaran: Yes ! Sorry I pushed too fast, I did not see your comment 🙈

PR #1276, opened 2018-02-16
Report swallowed errors to Haystack

tarebyte:
Because we are swallowing the GitHub Errors in the Controller we aren't reporting them to Haystack which means we have no idea what's going on.

This fixes that by adding the `failbot` after the `Octokit::Error` capture.


BenEmdon:
## Update
I took a jab at this myself. The reason why `failbot($ERROR_INFO)` was throwing `NoMethodError`s was because the convenience method `failbot` is in a `ActionControllerUtilities`.

The fix was to reference the singleton module `Failbot` and call it's `report!` method.

PR #1288, opened 2018-02-22
Fix accept assignment view

Kadaaran:
Hi there 👋 

> Based on user bug reported by our support team. 

I found out that if the assignment was already accepted and a user goes on the invitation link:
- the user is redirected to the setup view even though the repository already existed on GitHub and the assignment was already accepted.

screen shot 2018-02-22 at 17 53 30

- If the assignment had no starter code it returns this view, even if the repository exists (404 because we check the import status on an assignment that was not imported because there is no starter code to import)

screen shot 2018-02-22 at 11 59 58

I added a few methods to check if the repository exists on GitHub and added a check for starter-code if there is one.
I applied it for `Assignment` and `GroupAssignment` as well even if the logic is different.
The user is now redirected to the success view if the repo is already created on GH and if the starter code is imported if there is one.

screen shot 2018-02-22 at 14 50 40

Also I could really use the community help to test this new feature because I can reproduce this bug in dev mode but not in production anymore 😞

### DIFF
--- a/app/controllers/assignment_invitations_controller.rb
+++ b/app/controllers/assignment_invitations_controller.rb
@@ -4,7 +4,7 @@ class AssignmentInvitationsController < ApplicationController
   include InvitationsControllerMethods
   include RepoSetup
 
-  before_action :check_repository_already_created_on_github, :check_user_not_previous_acceptee,
+  before_action :check_repo_already_on_github, :check_user_not_previous_acceptee,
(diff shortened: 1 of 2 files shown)


Kadaaran:
After talking with @johndbritton (👋) it appears that this feature works only for the staff and improves the `repo_setup` feature if it's enabled.

I'm closing my PR for now just the time to check with the team what we are going to do with this feature 🙏

PR #1290, opened 2018-02-23
Add option to disable assignment invitations

d12:
As requested at SIGCSE, me and @johndbritton agreed this'd be easy to do and useful for teachers looking to archive their classroom and prevent students from joining the classroom after the course has ended.

~~Still need to write specs~~

http://recordit.co/A6TXSKA6wI


db/migrate/20180223160042_add_inivitation_enabled_to_assignment.rb
```diff
@@ -0,0 +1,5 @@
+class AddInivitationEnabledToAssignment < ActiveRecord::Migration[5.1]
+  def change
+    add_column :assignments, :invitations_are_enabled, :boolean, default: true
```
tarebyte: I'm not a big fan or `invitation_are_enabled`, maybe just `invtiations_enabled`

spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -130,6 +130,16 @@
         patch :accept_invitation, params: { id: invitation.key, group: { title: "Code Squad" } }
       end
 
+      it "fails if assignment invitations are disabled" do
+        group_assignment.invitations_are_enabled = false
```
tarebyte: Can we just make this

```ruby
group_assignment.update_attributes(invitations_enabled: false)
```

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -100,16 +100,21 @@ def group
...
 
   def create_group_assignment_repo(selected_group: group, new_group_title: nil)
-    users_group_assignment_repo = invitation.redeem_for(current_user, selected_group, new_group_title)
-
-    if users_group_assignment_repo.present?
-      GitHubClassroom.statsd.increment("group_exercise_invitation.accept")
-      yield if block_given?
+    if !group_assignment.invitations_are_enabled?
+      flash[:error] = "Invitations for this assignment have been disabled."
+      render :show
```
tarebyte: Should we be redirecting here like down below?

app/models/assignment_invitation.rb
```diff
@@ -34,6 +34,10 @@ def redeem_for(invitee)
     assignment_repo = AssignmentRepo.find_by(assignment: assignment, user: invitee)
     return AssignmentRepo::Creator::Result.success(assignment_repo) if assignment_repo.present?
 
+    unless assignment.invitations_are_enabled?
```
tarebyte: Totally a style nit, but what do you think about delegating `Assignment#invitations_enabled?` to `AssignmentInvitations#enabled?`.

Just a little nicer DSL

PR #1294, opened 2018-03-02
Remove homebrew/versions

Kadaaran:
`homebrew/versions`[here](https://github.com/Homebrew/homebrew-versions) is now deprecated (it's now part of homebrew/core) and might cause troubles during the `./server/setup`

This PR is the same than @talsafran 's ( 👋) one for the education's website 👍

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -2,7 +2,6 @@
 
 # Helpers
 tap "caskroom/cask"
-tap "homebrew/versions"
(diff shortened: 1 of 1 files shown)


talsafran:
Looks like we can close this since https://github.com/education/classroom/commit/e2f61251f4c3d4029fd8abdefb18d548a47b793c is on master meow?

Issue #1291, opened 2018-02-23
Roster Download


robertcduvall:
Only downloads the first page (20) of students --- it needs to do all the pages.

johndbritton:
@Kadaaran Can you verify this bug, write a failing test, and fix?

Kadaaran:
Hi @robertcduvall , the PR is now merged this issue should be fixed.
Please keep me updated if it's (or not) the case.

Thank you for your patience and for reporting this issue.

markpolyak:
I had the same issue when I tried to download the rooster a couple of weeks ago. I can confirm that now the "Download" button on "Rooster management" page downloads a CSV file with all students, not just 20 shown on the web page.

PR #1296, opened 2018-03-03
Remove pagination for download

Kadaaran:
Fix for #1291 
I forgot to remove pagination for the download button on my last PR 🙈 
The download button returned only the first 20 students not all of them as it should have.

I removed pagination so it returns now all of the students from the roster.

TODO: 
- [x] Tests for all entries
- [x] Tests with pagination

### DIFF
--- a/app/controllers/orgs/rosters_controller.rb
+++ b/app/controllers/orgs/rosters_controller.rb
@@ -110,6 +110,7 @@ module Orgs
     end
 
     def download_roster
+      @roster_entries = @current_roster.roster_entries.includes(:user).order(:identifier)
       respond_to do |format|
         format.csv { send_data @roster_entries.to_csv, filename: "classroom_roster.csv", disposition: "attachment" }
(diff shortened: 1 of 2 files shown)


Kadaaran:
Sure ! Sorry I forgot the `WIP` 🙈

robertcduvall:
I am still getting the same behavior when I tried it just now (I do not know how to verify if I am using your latest version).  Perhaps you can check it directly at:
    https://classroom.github.com/classrooms/35479176-duke-compsci-web-development/roster

Also, when talking to John B he suggested including GitHub IDs in the CSV data in case they change their github_username during the semester.  I can create another issue for this if you prefer.

Kadaaran:
Ooh my bad, it seems that the deploy did not actually pass at all 🙈
Classroom is still running with the old version. I'm going to fix it and I'll let you know as soon as it's fixed !

Thank you for your patience 🙏

Issue #1297, opened 2018-03-06
script/setup fails because of java8 in Brewfile


talsafran:
When I run `script/setup` it yells at me that the `java8` cask is not available:

```bash
15:29 $ script/setup
==> Ensuring Homebrew macOS dependencies...
Updated 1 tap (caskroom/cask).
No changes to formulae.
==> Installing Homebrew dependencies...
Using caskroom/cask
Using github/bootstrap

Error: Cask 'java8' is unavailable: No Cask with this name exists.
Installing java8 has failed!

...

Homebrew Bundle failed! 1 Brewfile dependency failed to install.
```

I was able to run `script/setup` successfully by using `java` instead of `java8` in the Brewfile, but noticed that in https://github.com/education/classroom/pull/1227 we specifically wanted `java8`. Not sure how to proceed here, thoughts?

---

cc @Kadaaran @tarebyte

Kadaaran:
It appears that ElasticSearch *finally* [supports](https://www.elastic.co/blog/elasticsearch-java-9-and-beyond) Java9 🎉 

I was able to run successfully `script/setup` and `script/server` as well. I think we can definitely remove `java8` and put back `java` 👍

talsafran:
w00t thanks @Kadaaran!

PR #1298, opened 2018-03-06
Put back java as ElasticSearch now supports Java9

Kadaaran:
Fix to #1297 

It appears that ElasticSearch finally supports Java9 🎉
We're going to put back `Java` and remove `Java8` from the Brewfile.

I uninstalled `java8` and I'm running with `java9` and all is good for now ✨ 
> ➜  classroom git:(update_Brewfile_java) brew cask info java
java: 9.0.4,11:c2514751926b4512b076cc82f959763f

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -5,7 +5,7 @@ tap "caskroom/cask"
 tap "homebrew/versions"
 tap "github/bootstrap"
 
-cask "java8"
+cask "java"
 cask "ngrok"
 
 brew "nodejs"


d12:
Do we need this dependency now that we're on Docker? The old `java8` didn't work for me so I just removed the line and it's worked just fine.

Kadaaran:
No indeed. Now that we're on Docker we don't need this one anymore 👍

d12:
@Kadaaran Does `java8` still cause brew crashes? I was thinking we could just remove the `java8` dependency all together.

PR #1309, opened 2018-03-14
Update GitHub client token for organization

### LINKED ISSUE #1308: Bad credentials when running RSPEC
Thanks to @d12 👋for this one.

I was getting `401 - Bad credentials` all over again and again, it seems that it comes from this line.

https://github.com/education/classroom/blob/a0f29d2742e315f7e093dd2270b1e28d98d2119a/app/models/organization.rb#L39

When running test we might use the token of a test account whom do not have rights over the `organization` which results with `401 - Bad credentials`.

A quick fix would be to  get the token from the first user in the database (who is the owner of the `organization`) to make sure we don't get 401, of course it runs only if we are in a test environment.

Kadaaran:
Intend to fix #1308 

If we are in test environment then we get the first user's token to make sure we don't get `401 - Bad credentials` when doing API calls.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -36,7 +36,7 @@ class Organization < ApplicationRecord
   end
 
   def github_client
-    token = Rails.env.test? ? users.limit(1).order("RANDOM()").pluck(:token)[0] : users.first.token
+    token = Rails.env.test? ? users.first.token : users.limit(1).order("RANDOM()").pluck(:token)[0]
 
     Octokit::Client.new(access_token: token)
   end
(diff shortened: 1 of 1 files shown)


d12:
I agree with ^. This should apply to every spec, so maybe something in `/spec/support`?

Kadaaran:
> What do you think about just stubbing the github_client method in the tests?

What do you mean by that ? Like moving it to another place ? Turn it into a macro ?

d12:
Mark is talking about RSpec stubs, there are some reasources online for learning about how they work!

Kadaaran:
I'm going to take a look thank you !

PR #1314, opened 2018-03-19
Add an issue template

Kadaaran:
> Suggestion: Add an issue template for Classroom.

It's difficult sometimes to really understand and figure out how users got the issue they reported.
I think it will be valuable to have an issue template, it could really help us and the community to better understand what happened and then be able to reproduce easily a bug to fix it.

If you have any ideas or suggestions on the content I'll take it 🙌


ashleytqy:
Hi! What about the case when the issue is a feature request?

Kadaaran:
We can add a feature request part under or we keep the feature-request label or both together 🤔

ashleytqy:
I think adding a first question like `is this a bug or a feature request?` would be helpful! reference: https://github.com/facebook/Docusaurus/blob/master/.github/ISSUE_TEMPLATE.md

Kadaaran:
It would if we add a _feature request_ part of the template. 
By adding the question and letting the issue template alone won't help that much. 
We could add a new section like the [Homebrew](https://github.com/Homebrew/brew/blob/master/.github/ISSUE_TEMPLATE.md) team did

johndbritton:
FYI, GitHub supports multiple issue templates: https://blog.github.com/2018-01-25-multiple-issue-and-pull-request-templates/

Kadaaran:
Yes but we can't precise which one we want to use except on the URL by using `&template=`

PR #1318, opened 2018-03-22
Fix jobs/organization_event_job.rb

Kadaaran:
Even if in GitHub the `Organization` contains two members or more, classroom might only have one recorded in the database if the users never signed in into Classroom.

We now check if the removed member is the last member, if it's the case then the job stops right away.


Kadaaran:
Woops I'm writing it, forgot to add the `in-progress` sorry 🙏

sapu76:
[spec/jobs/organization_event_job_spec.rb](url)
if job event closed so what we do. review is ready to in progress.

Issue #1301, opened 2018-03-06
script/server fails


Kadaaran:
All good when running `script/setup` but when I run `script/server` it yells at me this:

```
Users/kassandre/GitHub/classroom/lib/docker_compose/parser.rb:51:in `service_attributes': undefined method `[]' for nil:NilClass (NoMethodError)
	from /Users/kassandre/GitHub/classroom/lib/docker_compose/parser.rb:38:in `block in parse'
	from /Users/kassandre/GitHub/classroom/lib/docker_compose/parser.rb:37:in `map'
	from /Users/kassandre/GitHub/classroom/lib/docker_compose/parser.rb:37:in `parse'
	from /Users/kassandre/GitHub/classroom/lib/docker_compose.rb:21:in `services_up?'
	from /Users/kassandre/GitHub/classroom/lib/docker_compose.rb:39:in `all_services_up?'
	from ./script/docker-services:16:in `block in <main>'
	from /Users/kassandre/.rvm/rubies/ruby-2.4.2/lib/ruby/2.4.0/fileutils.rb:120:in `chdir'
	from /Users/kassandre/.rvm/rubies/ruby-2.4.2/lib/ruby/2.4.0/fileutils.rb:120:in `cd'
	from ./script/docker-services:15:in `<main>'
```

I've been adding debug messages and it fails eventually on `lib/docker_compose/parser.rb:55` , I think we try to pop an array which was already under `.pop` earlier or something like that ?

talsafran:
+1, I get this when I run `script/test`:

```
14:49 $ script/test
==> Ensuring Homebrew macOS dependencies...
Already up-to-date.
The Gemfile's dependencies are satisfied
yarn check v1.5.1
success Folder in sync.
✨  Done in 0.41s.
/Users/talsafran/github/classroom/lib/docker_compose/parser.rb:50:in `service_attributes': undefined method `[]' for nil:NilClass (NoMethodError)
	from /Users/talsafran/github/classroom/lib/docker_compose/parser.rb:38:in `block in parse'
	from /Users/talsafran/github/classroom/lib/docker_compose/parser.rb:37:in `map'
	from /Users/talsafran/github/classroom/lib/docker_compose/parser.rb:37:in `parse'
	from /Users/talsafran/github/classroom/lib/docker_compose.rb:21:in `services_up?'
	from /Users/talsafran/github/classroom/lib/docker_compose.rb:39:in `all_services_up?'
	from ./script/docker-services:16:in `block in <main>'
[...]

PR #1320, opened 2018-03-23
script/setup and script/server fix

Kadaaran:
Related to #1301 

After pairing with @d12 it appears that he had trouble to run the scripts as well so after digging a bit I tried to find a solution for both scripts.

**script/server**: we check if there is any existing containers, we stop them so we can restart them properly with the script.
**script/setup**: what happens is that we try to access the database when it's not ready yet. It solves everything by just waiting for the containers to be ready but I admit that it's not a pretty fix 🙈

#

cc @tarebyte as you worked, any suggestions would help me a lot (also this is a first version that I will improve 😤)
@d12 @damaneice @talsafran If you have time next week to try it on your end that would be awesome 🙏

### DIFF
--- a/script/server
+++ b/script/server
@@ -7,6 +7,11 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+if [ -f "$(which docker-compose)" ]; then
+  echo "===> Stopping existing containers..."
+  docker-compose stop
+fi
+
 # ensure everything in the app is up to date.
 script/update
 
--- a/script/setup
+++ b/script/setup
@@ -14,6 +14,9 @@ fi
 
 script/bootstrap
 
+# We make sure the containers are mounted before any further action
+sleep 5
+  
 echo "===> Preparing database..."
 bin/rails db:setup
 
--- a/script/setup
+++ b/script/setup
@@ -14,9 +14,9 @@ fi
 
 script/bootstrap
 
-# We make sure the containers are mounted before any further action
+echo "===> Waiting for the containers to mount..."
 sleep 5
-  
+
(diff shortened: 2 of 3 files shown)


script/server
```diff
@@ -7,6 +7,11 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+if [ -f "$(which docker-compose)" ]; then
+  echo "===> Stopping existing containers..."
+  docker-compose stop
+fi
```
d12: Maybe here, if docker-compose isn't installed, we can print a helpful message and `exit 1`?

script/setup
```diff
@@ -14,6 +14,9 @@ fi
 
 script/bootstrap
 
+# We make sure the containers are mounted before any further action
+sleep 5
```
d12: Small nitpick, it might be nice to log something before sleeping here, like `echo "Waiting for containers to mount..."` or something like that.

script/server
Kadaaran: Technically, this script is run after `script/setup` which by running `script/bootstrap` checks if docker-compose is installed and returns a message if it's not installed. 
https://github.com/education/classroom/blob/f0ca7f98d62c8c2a92f12a748b03a6f9ad4b770c/script/bootstrap#L53

It can happen if someone doesn't run script in the right order though 🤔 I can add it if you think it'll be helpful.

script/setup
Kadaaran: I agree !

script/server
d12: Ooh ok, no this is probably ok then.

Issue #1303, opened 2018-03-13
assignment repository web page "next" (page 2, 3, ...) buttons don't work


cindyanorris:
I'm unable to get to the second, third, etc page of classroom assignment repositories.  Clicking on next or manually changing the URL doesn't work.  It only returns the first page. 

![page1](https://user-images.githubusercontent.com/6600260/37316186-100d7a2e-2634-11e8-83df-420f3feca9d5.png)
![page2](https://user-images.githubusercontent.com/6600260/37316187-101ffac8-2634-11e8-9ee1-a3f04ea424a6.png)

Kadaaran:
Hi @cindyanorris 👋 First thank you for reporting this issue.

Unfortunately for now after trying a few times, I can't reproduce your issue on my end. 
Did it happen suddenly or did you already have this issue before ?

As soon as I can reproduce it I'll fix it.
Thank you for your patience.

sfikas:
Hi, I'm having the same problem.

The 'next page' button takes me back to the same page.
The list is a set of *group* assignments:
https://classroom.github.com/classrooms/23520047/group-assignments/psee-heim-2017-2i-seira-askiseon

I see the problem for @cindyanorris 's list is also with a list of *group* assignments, this likely makes the difference (?)

Kadaaran:
Hi @sfikas 
Yes it does you're right it's only on _group assignments_ ! 

I started to dig more into it but I'm still not able to reproduce the issue on my end in dev 😞

Kadaaran:
Hi everyone 👋 
I *finally* found what causes this issue and I'm working on a fix right away.

Thank you for your patience 🙏

Kadaaran:
Hello everyone,

Just to let you know that the fix is now merged and the issue should be fixed 🎉

sfikas:
It works, thanks!

PR #1326, opened 2018-03-26
Fix pagination for non rosters orgs

Kadaaran:
Fix to #1303 
Found out that the bug is linked to #1228 @d12 🙏 

This issue occurs in a particular situation: in the group_assignment page for organizations without roster.
This line was sending to the partial another page parameter than the regular one we should have.

https://github.com/education/classroom/blob/3b01242b9c5f769a1fa12234c83723713851f8fd/app/controllers/group_assignments_controller.rb#L32
https://github.com/education/classroom/blob/3b01242b9c5f769a1fa12234c83723713851f8fd/app/views/group_assignments/show.html.erb#L99

**TODO**:  Write test (which are not overkill) for pagination

### DIFF
--- a/app/controllers/group_assignments_controller.rb
+++ b/app/controllers/group_assignments_controller.rb
@@ -29,10 +29,11 @@ class GroupAssignmentsController < ApplicationController
   end
 
   def show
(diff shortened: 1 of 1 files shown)


d12:
Thanks for taking care of this @Kadaaran ! Ping me when you add the test and I'll take a look :eyes:

Issue #1319, opened 2018-03-22
Admin list in Classroom settings


Kadaaran:
## Feature Request

**What happens now**:  Classroom is not fully synchronised with GitHub, teachers can't track who has admin access on Classroom or not.
Ex: An `organization` has 2 owners on GitHub, but only one of them signed in into Classroom. Classroom tells us that the `organization` has only one member.

Even if we keep track of whom is no longer part of an organization because we merged #1274 , I really like the idea for the teachers to have a page where they can check who has admin access over their Classroom and be able to revoke accesses when necessary (which we don't have for now).

**How**: The idea would be to add a section under the admin invitation link a page where the owner can revoke directly teachers access over the Classroom and check who has admin access.

<img width="800" alt="screen shot 2018-03-22 at 16 48 22" src="https://user-images.githubusercontent.com/11220823/37781381-e270aaa4-2df0-11e8-9b2b-c68541353df5.png">

Feel free to comment if you have any suggestions 🙏

johndbritton:
@Kadaaran For the implementation, would removing a classroom administrator also remove the user as an owner of the GitHub organization?

Kadaaran:
@johndbritton No it just removes access on Classroom side. 
But a few weeks ago we merged a PR that gets the event of when an owner leaves an org so we can update the org in Classroom as well.

PR #1331, opened 2018-03-27
Administrators list

Kadaaran:
Fix to #1319

I've added the administrators list in the invite_administrators section.
Now teachers are able to revoke access to other administrators as well.

screen shot 2018-03-27 at 19 31 52

I added some safety checks in the code (in case someone wants to mess with us with the params) and of course we can't remove an owner if this is the last one.

screen shot 2018-03-29 at 00 11 38

TODO:

- [x] RSPEC and final safety checks
- [x] refactoring


app/controllers/organizations_controller.rb
```diff
@@ -143,4 +156,17 @@ def update_organization_params
       .require(:organization)
       .permit(:title)
   end
+
+  def verify_organization_user
+    @removed_user = User.find(params[:user_id])
+    not_found unless @removed_user && current_organization.users.map(&:id).include?(@removed_user.id)
+  end
```
tarebyte: If `find` fails this will cause a `404` anyway. So the first part of the conditional is not needed

```ruby
@removed_user = User.find(params[:user_id])
not_found unless current_organization.users.pluck(:id).include?(@removed_user.id)
```

app/controllers/organizations_controller.rb
```diff
@@ -71,6 +72,18 @@ def destroy
     end
   end
 
+  def remove_user
+    if current_organization.users.count < 2
```
tarebyte: Should we be enforcing this at the model level?

app/controllers/organizations_controller.rb
```diff
@@ -7,6 +7,7 @@ class OrganizationsController < Orgs::Controller
   before_action :set_users_github_organizations,      only: %i[index new create]
   before_action :add_current_user_to_organizations,   only: [:index]
   before_action :paginate_users_github_organizations, only: %i[new create]
+  before_action :verify_organization_user,            only: [:remove_user]
```
tarebyte: Perhaps `verify_user_belongs_to_organization` ?

app/controllers/organizations_controller.rb
```diff
@@ -143,4 +156,17 @@ def update_organization_params
...
+  def verify_organization_user
+    @removed_user = User.find(params[:user_id])
+    not_found unless @removed_user && current_organization.users.map(&:id).include?(@removed_user.id)
+  end
+
+  def transfer_assignments
+    new_owner = current_organization.users.where.not(id: @removed_user.id).first
+    current_organization.all_assignments.map do |a|
+      a.creator_id = new_owner.id if a.creator_id == @removed_user.id
+      a.save
```
tarebyte: You could clean this up a bit with `next`

```ruby
next unless assignment.creator_id == @removed_user.id
assignment.update_attributes(creator_id: new_owner_id)
```

app/controllers/organizations_controller.rb
```diff
@@ -143,4 +156,17 @@ def update_organization_params
...
   end
+
+  def verify_organization_user
+    @removed_user = User.find(params[:user_id])
+    not_found unless @removed_user && current_organization.users.map(&:id).include?(@removed_user.id)
+  end
+
+  def transfer_assignments
+    new_owner = current_organization.users.where.not(id: @removed_user.id).first
+    current_organization.all_assignments.map do |a|
```
tarebyte: Instead of using `a` can we use `assignment` here?

app/models/user.rb
```diff
@@ -54,6 +54,10 @@ def staff?
     site_admin
   end
 
+  def owns_assignments?(organization)
```
tarebyte: The naming here doesn't really fit the method, could we update this to something like `owns_all_assignments_for?(organization)` ?

app/controllers/organizations_controller.rb
Kadaaran: We can add a new method called `one_owner_remains?` to organization 👌

app/controllers/organizations_controller.rb
Kadaaran: Awesome ! I'm not used to `next` but it seems great (gonna keep this for the next time it happens)

PR #1345, opened 2018-04-12
Update to rails 5.2

Kadaaran:
Update from Rails `5.1` to `5.2` 🔥 

About why I skipped some test with Bullet: https://github.com/flyerhzm/bullet/issues/382
There is an issue with `has_one` association with the gem reported in January, a fix will be implemented in the next months I believe.


d12:
Looks like there's still CI failiures on this related to eager loading. Let me know when you've got it sorted!

Kadaaran:
@d12 I think it's fixed ✨

config/initializers/bullet.rb
```diff
@@ -3,6 +3,9 @@
 Rails.application.configure do
   config.after_initialize do
     if defined? Bullet
+      Bullet.add_whitelist(type: :n_plus_one_query, class_name: "GroupAssignment", association: :deadline)
+      Bullet.add_whitelist(type: :n_plus_one_query, class_name: "Assignment", association: :deadline)
```
johndbritton: While this does make the errors go away, this doesn't actually solve the underlying problem. I think it's worth spending the time to find the source of the bug in Bullet and get it fixed upstream if it is in fact an issue with Bullet.

@Kadaaran @d12 My concern is that these exceptions will be forgotten and be here forever, even after a potential fix.

Kadaaran: @johndbritton I started to check the code on Bullet and the `spec` are written for `has_many` relationships and there is no sign of `has_one` relationship. 
It seems tricky for me but I can give it a try 👍

johndbritton:
Replaced by https://github.com/education/classroom/pull/1369

PR #1346, opened 2018-04-12
Delete team when it's deleted in GitHub

Kadaaran:
Hi 👋 

We don't track when teams are deleted which led to unwanted behaviours this week 🙈 
The repo access is deleted along with the team. 

**Question**: Should we delete the repo on GitHub as well ? 🤔 

Thanks 🌊


d12:
This all makes sense to me, I'd like to hear what @tarebyte thinks. 

> Should we delete the repo on GitHub as well ?

I don't think we should delete GitHub resources here.

Thanks for doing this @Kadaaran :)

BenEmdon:
Can we get a second approval? CC: @d12 @srinjoym

Issue #1193, opened 2017-10-16
GitHub Classroom Roster Assignment Paging Error


ccannon94:
When selecting "Unlinked GitHub accounts" tab on the Class Roster Management Page, changing pages of the list will send the user back to the "All students" tab. Then, when "Unlinked GitHub accounts" is selected again, it is on the correct page. Not impossible to use, just weird.

For example, when on page 1 of "Unlinked GitHub accounts", when I select page 2, the page reload brings me back to the "All students" tab. However, when I click the "Unlinked Student accounts" tab, I am on page 2 as requested.

Preferably, selecting a page in "Unlinked GitHub accounts" would just refresh on the desired page.

sethb157:
Hi! I want to start contributing to Classroom, and I figured this might be a good place to start. I've been looking at the code--seems like the JavaScript this is relying on is acting a bit interesting across pages. Was wondering if there's someone I can send a few questions to. @mozzadrella, could you point me to the right person? Thanks!

Kadaaran:
Hi @sethb157 👋 
Awesome ! Feel free to ask me any questions I'll try so answer to all of them 💪

sethb157:
Ok awesome! There are two places where the All Students/Unlinked Students tabs show up: once on the Assignment Detail view and once in the Roster Management view. The way things work now, the two are connected. That is, the following happens: 1) Start on Assignment Detail view, with All Students tab open 2) Go to Roster Management view, click on Unlinked Students tab 3) Go back to Assignment Detail view, Unlinked Students tab is now open, because you clicked on it in Roster Management.

I’m assuming this isn’t expected behavior??😬 if it is okay/expected, then this bug is a one-line fix. If not, It’ll be a little more refactoring to make the two act independently. Thoughts?

sethb157:
If you need more clarification, please let me know!

Kadaaran:
@sethb157 Oops. The expected behaviour is that the two paginations should work independently from each other.

What happens is that in the view, the Javascript we load is the roster one
https://github.com/education/classroom/blob/4346984e6ca87e555329777b3df0e9ff926ab015/app/views/assignments/show.html.erb#L96

That's why I _think_ the targeted tabs are the same no matter page you change. One solution could be to set re-init the variables every time we load or something like that 🤔

sethb157:
[...]

PR #1351, opened 2018-04-22
Fix tabs not restoring selected tab after refresh

sethb157:
This bug caused the roster tabs on the management page
to forget the currently selected tab on page refresh.
In order to fix this, the roster_tabs js file
has been replaced with a more generic manage-tabs
implementation, which both fixes the bug and makes the addition
of more tabs a little easier. All groups of tabs have been updated
to use manage-tabs.

NOTE: There are no tests attached to this yet. @Kadaaran suggested that I wait till the update to 5.2 is done, in order to test with capybara.


d12:
Closes https://github.com/education/classroom/issues/1193

BenEmdon:
Hey @sethb157 looks like we are on Rails 5.1 https://github.com/education/classroom/blob/d82ea459cc21a3f6d1e5d678402c3439a63ea633/Gemfile#L11

sethb157:
@BenEmdon Shoot, just saw this. I'll add in testing and make another PR. Thanks for the reminder!

BenEmdon:
This has become too stale for us to consider. If you decide to rework this PR please open an issue and tag `@education/classroom` with the feature request.

PR #1367, opened 2018-06-14
[WIP] Reject user with invalid access token

srinjoym:
[WIP] We were only adding to the current scopes when updating from Github. When the user rejects the token, the scopes come back as empty from Github but the old scopes are still returned so the user can keep using classroom. Tests coming...


app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -20,6 +20,7 @@ def adequate_scopes?
   end
 
   def authenticate_user!
+    # require 'byebug'; byebug
```
BenEmdon: 🔪 hehe

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    context "user access token is no longer valid" do
+      before do
+        User.any_instance.stub(:github_client_scopes).and_return([])
+      end
+      it "should logout user and delete session" do
+        get :index
+        expect(response).to redirect_to(login_path)
+        expect(session[:pre_login_destination]).to eq("http://test.host/anonymous")
+        expect(session[:required_scopes]).to eq(GitHubClassroom::Scopes::TEACHER.join(","))
```
johndbritton: usually we try to have one `expect` per `it` block. You can split these up into different `it` blocks describing what you're testing.

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  let(:user)    { classroom_teacher }
+
+  before(:each) do
+    sign_in_as(user)
+  end
+
+  describe "user authentication tests" do
+    context "user access token is valid", :vcr do
+      it "should render action" do
```
johndbritton: don't use `should` in descriptions. write as if it's the truth.

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        get :index
+        expect(response.status).to eq(200)
+      end
+    end
+
+    context "user access token is no longer valid" do
+      before do
+        User.any_instance.stub(:github_client_scopes).and_return([])
+      end
+      it "should logout user and delete session" do
```
johndbritton: same thing with `should` here.

app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -5,7 +5,7 @@ class ApplicationController
 
   def current_scopes
     return [] unless logged_in?
-    session[:current_scopes] ||= current_user.github_client_scopes
+    session[:current_scopes] = current_user.github_client_scopes
```
johndbritton: This seems like it is more correct. The `||=` here implies we were maybe trying to save an API call, but I don't think that's something we need / want.

@tarebyte can you double check this since I think you wrote it originally.

tarebyte: This seems fine 👍

johndbritton: Do you know why we're putting `current_scopes` into the session in the first place?

BenEmdon:
I see you get some whitespace barks from _the hound_. To trim whitespace on save:

> Under your Atom `Preferences` go to `Packages` tab and search for `whitespace`. Click on the `whitespace` package and uncheck `Ensure Single Trailing Newline` option

app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -20,7 +20,10 @@ def adequate_scopes?
   end
 
   def authenticate_user!
```
johndbritton: Might be worth adding a comment here explaining why this is our authentication logic.

app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -48,6 +51,11 @@ def logged_in?
     !current_user.nil?
   end
 
+  def log_out
```
johndbritton: We might want to make use of the flash to tell the user that they were logged out of Classroom.

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  let(:user)    { classroom_teacher }
+
+  before(:each) do
+    sign_in_as(user)
+  end
+
+  describe "user authentication tests" do
```
johndbritton: ~~tests~~

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    context "user access token is not authorized" do
+      before do
+        User.any_instance.stub(:authorized_access_token?).and_return(false)
+      end
+      
+      it "redirects to home page" do
+        get :index
+        expect(response).to redirect_to(root_path)
+      end
+      it "clears the session" do
```
johndbritton: missing newline

Issue #1253, opened 2018-02-02
Making the import process more resilient


tarebyte:
Yesterday we had a lot of folks not be able to import repositories because of an issue with the service backing the [Source Imports API](https://developer.github.com/v3/migration/source_imports/)

This tells me that we need to make this process more resilient on our end in order to make it more clear to the end user what is going on.

I propose moving the import process to the background and put the assignments a state of sorts.

These states could be
- `pending`
  - The assignment isn't ready to go
- `ready`
  - The assignment can be worked on by students
- `errored`
  - Something has gone horribly wrong and intervention is probably needed

Unfortunately the only way to know if the import was successful is the [continuously pull from the Source Imports API](https://developer.github.com/v3/migration/source_imports/#get-import-progress). So we'll need create timeouts and retries.

/cc @Kadaaran @d12 for thoughts

tarebyte:
Another thought is that we could add on we could add on to the work that @wilfriedE did with repository imports to make the UI experience even better.

I'm thinking instead of pulling in the foreground we can use action cable and let it listen for the status of the background job.

Kadaaran:
Could we maybe add a time limit when it's `pending` and then update it to `errored` if it takes too long so users know what's happening ?

Kadaaran:
> **TLDR;** This is the new import process design I came up with. We'll back up the process with state machine and secure actions for each state so we won't have issues when Porter is down 💪 

### What exactly is happens
### After a few research here's what I suggest:



With @johndbritton's help , based on talks with @tarebyte and some research I've come to this new design:

<img width="518" alt="screen shot 2018-04-12 at 18 04 02" src="https://user-images.githubusercontent.com/11220823/38989006-9622d186-43d5-11e8-94ca-3c9b2a7207df.png">

I was inspired by the [education-web](https://github.com/github/education-web/blob/master/app/models/discount_request.rb#L33-L65) way of using state machine.
So to help us redesign the process I've thought about using the state-machine gem [Statesman](https://github.com/gocardless/statesman) which seems perfect for what we are trying to do and is still maintained ✨ 


Each state is represented in green.
We're going to have a new class which will keep track of the transition: `ImportStateMachine

Kadaaran:
> TLDR; This is the new import process design I came up with. We'll back up the process with state machine and secure actions for each state so we won't have issues when Porter is down 💪

### What exactly happens

This bug is happening because of the way the action are ordered and *mainly* because we don't take actions when Porter triggers an `Error`.
[...]

PR #1368, opened 2018-06-18
Import resiliency (1/3)

BenEmdon:
## What
This PR (1/3) proposes a different architecture to improve repository creation/import resiliency for the purpose of addressing #1253.
The architecture I am proposing is to move repository creation/import into a background job and communicate with the client asynchronously via ActionCable. Moving creation/import to a background job is a easy first step to resiliency since ActiveJobs can be retried.

I've decided to break this refactor into 3 parts:
1. Create `CreateGitHubRepositoryJob` job which is more resilient than the current offering
1. Create some form of porter polling job that can watch the status of the starter code import
1. Implement the new jobs in place of the current offering. Use ActionCable to asynchronously message the client with status.

## How
In this PR I created `CreateGitHubRepositoryJob` which essentially performs the same logic as [what is currently in place](https://github.com/education/classroom/blob/0638f37ce1ebc978926899cb8f64cc2b7fd4e8bf/app/models/assignment_repo/creator.rb#L58-L87). The big difference is that if at any point the repo creation/import fails, the GitHub repository is deleted and the creation/import job is retried. (Edit: the retry is done with exponential backoff).

Could I get some 👀 from @johndbritton or @tarebyte


app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class AssignmentRepo
+  class CreateGitHubRepositoryJob < ApplicationJob
+    queue_as :create_repository
+    retry_on Creator::Result::Error, queue: :create_repository
```
BenEmdon: `retry_on` defaults to 3s wait and 5 attempts. 
This should be enough for what we are trying to do, right?

Also if we wanted we could pass a `wait: :exponentially_longer` which would perform an exponential backoff algorithm.

tarebyte: I'd vote for the exponential backoff.

johndbritton: Yeah, especially because the main reason we'd hit a problem like this is GitHub API downtime / issues.

BenEmdon: I'll add 🚩

spec/jobs/create_github_repository_job_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+    end
+
+    it "creates an AssignmentRepo as an outside_collaborator" do
+      subject.perform_now(assignment, student)
+
+      result = assignment.assignment_repos.first
+      expect(result.nil?).to be_falsy
+      expect(result.assignment).to eql(assignment)
+      expect(result.user).to eql(student)
```
johndbritton: Style point on specs: Aim to have one `expect` per it block. If you need multiple `expects` split them up into separate it blocks and group them together logically in a `context`block.

BenEmdon:
## Update
* I added the exponential backoff 🚩
* I changed spec style to use shared contexts

## Response
> I would expect a side effect to happen if it fails and doesn't proceed to the next step. 
  @johndbritton 

When ActionCable is set up we can broadcast the progress or failure of the repo create/import. Having a progress bar would also be a nice user experience.

PR #1375, opened 2018-06-21
Octopoller - asking are we there yet - until we are

BenEmdon:
## What
To make repo import more resilient I proposed (in #1368) that we have a cascading job after `CreateGitHubRepositroyJob` that polls the [porter api](https://developer.github.com/v3/migrations/source_imports/) for repo import status.

The problem I encountered was that there weren't any modern polling ruby gems that looked well maintained/used so I took it upon myself to write my own.

## How does it work
Octopoller has a simple but feature rich api inspired by gems [atomicobject/fire_poll](https://github.com/atomicobject/fire_poll) and [ooyala/retries](https://github.com/ooyala/retries) (both under MIT license). Here's how you would use it:

```ruby
result = Octopoller.poll(wait: 1, timeout: 10) do 
  if we_are_there_yet() # waits 1 second between asking for a max of 10 seconds 
    "We are here, finally!"
  else
    raise "We aren't there yet!"
  end
end

# result => "We are here, finally!"
```

It also has the ability to record and handle each error individually:
```ruby 
error_handler = proc { |error| puts error }
Octopoller.poll(error_handler: error_handler) do 
  raise "O c t o p o l l e r 🦑"
end

# "O c t o p o l l e r 🦑"
# "O c t o p o l l e r 🦑"
# ... for 15 seconds
# Polling timed out paitently (Octopoller::TimeoutError)
```

## Shouldn't this be a ruby gem
Yes. For now this could live inside education/classroom and in my free time I'll pull it out into a ruby gem. If this proposed gem is accepted should it live under the GitHub or Education org, or my user?


BenEmdon:
> I'd be interested in having this inside of https://github.com/octokit/octokit.rb 😄
  @tarebyte 

I like this idea 🙌 
Should I extract it out of classroom now or after I improve import resiliency?

tarebyte:
> or after I improve import resiliency?

Definitely after, let's get this out the door and battle test it a little bit

BenEmdon:
> Definitely after, let's get this out the door and battle test it a little bit
  @tarebyte 

Sounds good. Can I get a 👍 👎 ?

d12:
@BenEmdon Do you and @srinjoym have permissions to merge your own PRs after they get :+1:'d?

BenEmdon:
> There might be a use-case for raising an exception inside of the polling block that doesn't force a retry (something bad went wrong and we don't want to retry), but we can cross that bridge if we ever get there.
  @d12 

You raise a good point. This will definitely need some refining before we try move it to oktokit.

Maybe we should pass an error class to `poll` as an argument that specifies which errors to catch?

PR #1378, opened 2018-06-29
Import resiliency (2/3)

BenEmdon:
## What
This PR is the second part of improving import resiliency. Import resiliency was raised as enhancement in https://github.com/education/classroom/issues/1253. Part (1/3) was merged in https://github.com/education/classroom/pull/1368.

### This PR introduces 2 things
This PR sets up ActionCable for web socket communication with a client. ActionCable is standard in Rails 5. Along with setting up ActionCable it introduces an ActionCable channel and various client side generated channel related code.
ActionCable files to look out for:
* `app/assets/javascripts/cable.js` (ActionCable setup)
* `app/assets/javascripts/channels/repository_creation_status.js` (scaffolded by Rails)
* `app/channels/application_cable/channel.rb` (base channel)
* `app/channels/application_cable/connection.rb` (channel authentication)
* `app/channels/repository_creation_status_channel.rb` (channel to broadcast repo creation status on)
* `config/routes.rb` (where the `/cable` route is mounted)
* `app/views/layouts/application.html.erb` (aded a new meta tag to all html)

This PR adds a `PorterStatusJob` which checks the status of an ongoing porter import. It does this by polling the [source imports API](https://developer.github.com/v3/migrations/source_imports/) using `Octopoller` (introduced in #1375). The `PorterStatusJob`'s purpose is to broadcast the status of importing to the `RepositoryCreationChannel`.

This change also adds channel broadcasting to `CreateGithubRepsitoryJob`.


Gemfile
```diff
@@ -98,6 +98,7 @@ group :production do
 end
 
 group :test do
+  gem "action-cable-testing"
```
BenEmdon: Added this gem because there aren't ways to test ActionCable channels currently in Rails. This gem is also currently a PR on Rails/Rails so it's probably our best option.

Should I set a static version?

app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -5,10 +5,17 @@ class CreateGitHubRepositoryJob < ApplicationJob
     queue_as :create_repository
     retry_on Creator::Result::Error, wait: :exponentially_longer, queue: :create_repository
 
-    # Create an AssignmentRepo.
+    class Status
+      CREATE_REPO = "Creating repository"
+      ADDING_COLLABORATOR = "Adding collaborator"
+      IMPORT_STARTER_CODE = "Importing starter code"
```
BenEmdon: The idea behind these statuses was to display something to the user as the repo creation job proceeds. 👍 👎 ❓

app/jobs/assignment_repo/porter_status_job.rb
```diff
@@ -0,0 +1,42 @@
...
+  class PorterStatusJob < ApplicationJob
+    class ImportInProgress < StandardError; end
+    queue_as :porter_status
+    retry_on Octopoller::TimeoutError, queue: :porter_status
+
+    # rubocop:disable MethodLength
+    def perform(assignment_repo, user)
+      github_repository = assignment_repo.github_repository
+
+      result = Octopoller.poll do # TODO: log errors
```
BenEmdon: We should probably log any errors that come our way here even if they are about to be swallowed. Do we log errors anywhere else in the app?

BenEmdon: Should the `Octopoller` API accept a `retry_on_error:` argument that accepts an error to re-poll on? Right now `Octopoller` swallows any error and re-polls.

app/models/github_repository/import_dependency.rb
```diff
@@ -19,6 +20,7 @@ def get_starter_code_from(source)
 
   def import_progress(**options)
     GitHub::Errors.with_error_handling do
+      options[:accept] = "application/vnd.github.barred-rock-preview"
```
BenEmdon: This was added to silence warnings about how the API is still in beta.

config/cable.yml
```diff
@@ -2,7 +2,7 @@ development:
   adapter: async
 
 test:
-  adapter: async
+  adapter: test
```
BenEmdon: Changed this so we can use `action-cable-testing` gem.

app/models/assignment_repo/creator.rb
```diff
@@ -6,6 +6,8 @@ class Creator
     REPOSITORY_CREATION_FAILED              = "GitHub repository could not be created, please try again"
     REPOSITORY_STARTER_CODE_IMPORT_FAILED   = "We were not able to import you the starter code to your assignment, please try again." # rubocop:disable LineLength
     REPOSITORY_COLLABORATOR_ADDITION_FAILED = "We were not able to add you to the Assignment as a collaborator, please try again." # rubocop:disable LineLength
+    REPOSITORY_CREATION_COMPLETE            = "Your GitHub repositroy was created."
```
ashleytqy: `repositroy` -> `repository`

BenEmdon: Good catch! Thanks, will fix 😄

Gemfile
d12: :+1: , and yeah lets specify a minor version here.

`"~> 0.3"`

app/channels/application_cable/connection.rb
```diff
@@ -0,0 +1,17 @@
...
+    identified_by :current_user
+
+    def connect
+      self.current_user = find_verified_user
+    end
+
+    protected
+
+    def find_verified_user
+      User.find_by(id: request.session[:user_id]) || reject_unauthorized_connection
```
d12: :+1:

app/jobs/assignment_repo/create_github_repository_job.rb
d12: If this class has nothing but these status messages, we can just define them on the job class. `CREATE_REPO_STATUS = "...`

app/jobs/assignment_repo/porter_status_job.rb
d12: If you need logging, you can use `Rails.logger`

app/jobs/assignment_repo/porter_status_job.rb
```diff
@@ -0,0 +1,42 @@
...
+        begin
+          progress = github_repository.import_progress[:status]
+          case progress
+          when GitHubRepository::IMPORT_COMPLETE
+            Creator::REPOSITORY_CREATION_COMPLETE
+          when *GitHubRepository::IMPORT_ONGOING
```
d12: Is this `*` supposed to be here?

BenEmdon: The `*` lets the `case`  [match against arrays](https://bengribaudo.com/blog/2013/12/31/2530/ruby-case-testing-against-arrays-of-values)

BenEmdon:
> I left some comments, but this looks good! Just confirming, nothing about this PR should affect how imports happen in production yet, right?
  @d12 

This does not change how imports happen in production. That will happen in a follow up PR.

spec/jobs/assignment_repo/porter_status_job_spec.rb
```diff
@@ -98,12 +114,12 @@
             headers: { "Content-Type": "application/json" }
           ).times(2).then
           .to_return(
-            status: 500,
+            status: 201,
```
d12: Why is the status here 201 instead of 500 or some other error code?

PR #1380, opened 2018-07-03
Reject user with invalid access token + Cassette fixes

srinjoym:
Continues #1367 (switched to branch on classroom)

Logs out user if token has been invalidated, had to rerecord most of the cassettes because it was an Application Controller level change.

Added check for test environment when choosing token to use for organization so that we didn't pick invalid tokens from random users created during tests.

Changed rosters controller spec to expect not found return when user does not belong to org, as this is what the code does right now, I think we didn't catch this before because the result was cached in the cassettes.

Added token filtering to VCR, we already had the method for this but just weren't calling it


app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -5,7 +5,7 @@ class ApplicationController
 
   def current_scopes
     return [] unless logged_in?
-    session[:current_scopes] ||= current_user.github_client_scopes
+    session[:current_scopes] = current_user.github_client_scopes
```
srinjoym: Changed to just use `=` so that we don't just keep adding to scopes

app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -20,7 +20,10 @@ def adequate_scopes?
   end
 
   def authenticate_user!
-    return become_active if logged_in? && adequate_scopes?
+    if logged_in?
+      return log_out unless current_user.authorized_access_token?
```
srinjoym: Logs out current user if their access token has expired

app/controllers/application_controller/authentication_dependency.rb
```diff
@@ -48,6 +51,11 @@ def logged_in?
     !current_user.nil?
   end
 
+  def log_out
```
srinjoym: Moved log_out method to authentication_dependency so it can be used here and in sessions controller

app/models/organization.rb
```diff
@@ -36,7 +36,11 @@ def all_assignments(with_invitations: false)
   end
 
   def github_client
-    token = users.limit(1).order("RANDOM()").pluck(:token)[0]
+    if Rails.env.test?
+      token = users.first.token unless users.first.nil?
```
srinjoym: If we're test mode then we pick the true user token (which is the first user) instead of randomly picking

spec/controllers/orgs/rosters_controller_spec.rb
```diff
@@ -30,11 +30,11 @@
         expect(response).to render_template("rosters/new")
       end
 
-      it "redirects if the user doesn't belong to the organization" do
+      it "sends not found if the user doesn't belong to the organization" do
```
srinjoym: Need feedback on this change @d12 
Rosters Controller sends a not found if the user can't be found so updated the test to match that.

.gitignore
```diff
@@ -43,3 +43,6 @@ spec/examples.txt
 
 # yarn
 /node_modules
+
+# mac
+.DS_Store
```
BenEmdon: 🙌 💻 💯

spec/models/github_user_spec.rb
```diff
@@ -33,15 +33,15 @@
 
   describe "#github_avatar_url", :vcr do
     it "returns the correct url with a default size of 40" do
-      expected_url = "https://avatars2.githubusercontent.com/u/#{github_user.id}?v=4&size=40"
-      expect(github_user.github_avatar_url).to eql(expected_url)
+      expected_url = %r{https:\/\/avatars[0-9].githubusercontent.com\/u\/#{github_user.id}\?v=4&size=40\z}
```
srinjoym: Added regex check since there can be multiple numbers in the avatar url (and the test started failing because of that)

spec/support/vcr.rb
```diff
@@ -38,11 +38,21 @@
     classroom_owner_organization_github_login
   end
 
+  # Owner token
+  c.filter_sensitive_data("<TEST_CLASSROOM_OWNER_GITHUB_TOKEN>") do
```
srinjoym: We weren't calling the actual methods for filtering tokens in vcr, might want to refactor this file in the future so there doesn't have to be multiple calls.

app/controllers/application_controller/authentication_dependency.rb
d12: > so that we don't just keep adding to scopes

What do you mean by this?

`||=` is usually used for simple caching, `session[:current_scopes] ||= current_user.github_client_scopes` returns `session[:current_scopes]` if it's not nil, and otherwise, it sets `session[:current_scopes]` and returns `current_user.github_client_scopes`.

app/models/organization.rb
d12: :+1:

spec/controllers/orgs/rosters_controller_spec.rb
d12: :+1:

app/controllers/application_controller/authentication_dependency.rb
srinjoym: With the `||=` operator the current_scopes did not return the right value if the user ever reduced their scopes (for example if the token was invalidated and valid scopes went to empty). In that case we just kept using the cached value of the scopes since it wasn't actually nil.

d12: Ok, sounds good :+1:

PR #1379, opened 2018-07-01
Allowing downloaded roster to contain group names

ashleytqy:
This PR addresses https://github.com/education/classroom/issues/1350

The `Download` button on the roster management page will open a modal with buttons to download CSV with / without groupings.

screen shot 2018-07-02 at 7 12 35 pm

PR #1382, opened 2018-07-04
Add scrolling

d12:
Adds scrolling to @ashleytqy 's work in #1379.


d12:
http://recordit.co/KFr78yeEwa

Issue #1381, opened 2018-07-04
Improve Sidekiq queue efficiency 


BenEmdon:
### Feature request :sparkles:
Currently classroom uses sidekiq queues inefficiently. 
This is our current queue setup:
https://github.com/education/classroom/blob/4f9db7e412c57ce8aeb328a355b223e0a7254546/config/sidekiq.yml#L7-L14

This is what the Sidekiq docs say about having lots of queues:
> I don't recommend having more than a handful of queues. Lots of queues makes for a more complex system and [Sidekiq Pro cannot reliably handle multiple queues](https://github.com/antirez/redis/issues/1785) without polling. 
  https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues

One way we could improve this would be to use the regular 3 queues (`critical`, `default`, and `low`).

v4vasanth:
Hi! I can help with this ✋

gustavosobral:
I think I can do this one. Have you done any work on this @v4vasanth?

--- 
@BenEmdon So, your idea here is to just create these three queues that you've mentioned and re-assign the workers? I thought about doing something like this:

```
:queues: # https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
  - [critical, 3]
  - [default, 2]
  - [low, 1]
```

And assigning the Workers that have a queue with priority 3 to the `critical` queue and so on.

Other option would be this setup:

```
:queues: # https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
  - critical
  - default
  - low
```

But from the Sidekiq documentation we have that:
> This means that any job in the default queue will be processed only when the critical queue is empty.

I don't like the idea of only processing from `default` when `critical` is totally empty. So I will go for the option one.

PR #1385, opened 2018-07-10
Import resiliency (3/3)

BenEmdon:
## What
This PR creates a new user flow that utilizes the new import resiliency changes made in #1368 and #1378 for the purpose of solving #1253. 

This particular PR proposes a couple of significant additions:
* Registers `CreateGitHubRepositoryJob`'s and `PorterStatusJob`'s queues with sidekiq
* Configures ActionCable with our Redis setup
* Added a `status` column to `AssignmentInvitation`
* Introduces a new `import_resiliency` feature 🚩 
* The following routes on `AssignmentInvitationController`:
  * `GET #progress` - JSON api that returns the status of an invitation
  * `POST #create_repo` - if the invitation is `accepted`, kicks off a `CreateGitHubRepositoryJob`
  * `GET #setupv2` - renders a live updating view of the repo creation progress

All of the new repo creation flow is hiding behind the `import_resiliency` feature 🚩 

## ActionCable
All the server work for ActionCable is done. The client side work will require some extra thought so I have left that work for a follow up PR since this is already an 🐘. Instead of ActionCable we are using short polling and hitting a `GET #progress` for now.

## What if it errors?
Right now if it errors the UI will reflect the error state but I am yet to add retry button. There is a route ready for the retry button but I have a question about unique/locking background jobs.

## What does it look like?
jul-10-2018 11-18-07
jul-10-2018 11-18-41


app/controllers/assignment_invitations_controller.rb
```diff
@@ -8,19 +9,79 @@ class AssignmentInvitationsController < ApplicationController
...
+  def setupv2
+    render status: 404 unless import_resiliency_enabled?
+  end
+
+  # rubocop:disable MethodLength
+  def create
+    if import_resiliency_enabled?
+      job_started = false
+      if current_invitation.accepted? || current_invitation.errored?
+        AssignmentRepo::CreateGitHubRepositoryJob.perform_later(current_assignment, current_user)
```
BenEmdon: Should we ensure that the job is a unique job before we kick it off? Since jobs retry with exponential backoff we could end up in a situation where another job is kicked off while another is about to be retired.

app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -56,12 +49,29 @@ def perform(assignment, user)
...
+          status: assignment.invitation&.status
+        )
+        PorterStatusJob.perform_later(assignment_repo, user)
+      else
+        assignment.invitation&.completed!
+        ActionCable.server.broadcast(
+          RepositoryCreationStatusChannel.channel(user_id: user.id),
+          text: Creator::REPOSITORY_CREATION_COMPLETE,
+          status: assignment.invitation&.status
+        )
```
BenEmdon: Should we be recording stats about successful repo creation on this vs the old way? if so using  `GitHubClassroom.statsd`?

config/sidekiq.yml
```diff
@@ -5,10 +5,12 @@
...
-  - [boom,         3]
+  - [assignment,        1]
+  - [chewy,             1]
+  - [last_active,       1]
+  - [github_event,      3]
+  - [trash_can,         2]
+  - [deadline,          3]
+  - [boom,              3]
+  - [create_repository, 3]
+  - [porter_status,     3]
```
BenEmdon: Worth mentioning that this many custom queues could be inefficient (#1381 )

d12:
Also one other thing I'd suggest, this PR is pretty big and it's hard to review and be sure it's 100% correct. If it's possible in the future, I'd recommend breaking PRs into more atomic, smaller changes 😅

app/controllers/assignment_invitations_controller.rb
```diff
@@ -8,19 +9,79 @@ class AssignmentInvitationsController < ApplicationController
...
+  # rubocop:enable CyclomaticComplexity
 
   def setup; end
 
+  def setupv2
+    render status: 404 unless import_resiliency_enabled?
+  end
+
+  # rubocop:disable MethodLength
+  def create
```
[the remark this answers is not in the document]
BenEmdon: Good point. `create` doesn't make sense. What are your thoughts on the actions instead following:
```ruby
patch :create_assignment
patch :create_repo
patch :create_assignment_repo
```

app/models/assignment_invitation.rb
```diff
@@ -20,12 +22,18 @@ class AssignmentInvitation < ApplicationRecord
...
   delegate :title, to: :assignment
 
   # Public: Redeem an AssignmentInvtiation for a User invitee.
   #
   # Returns a AssignmentRepo::Creator::Result.
-  def redeem_for(invitee)
+  #
+  # rubocop:disable MethodLength
+  # rubocop:disable AbcSize
+  def redeem_for(invitee, import_resiliency: false)
```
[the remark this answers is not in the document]
BenEmdon: I can't use the feature dependency here because we are outside of a controller. (feature 🚩s are defined on `ApplicationController`).

app/controllers/assignment_invitations_controller.rb
BenEmdon: By default Sidekiq doesn't have this feature. It does in enterprise, but we aren't on that plan.

Another option would be to use [mhenrixon/sidekiq-unique-jobs](https://github.com/mhenrixon/sidekiq-unique-jobs). This would require us to move away from ActiveJob and to using straight Sidekiq.

Possibly worth investigating in a follow up PR.

BenEmdon:
## Update
* Changed `#create` -> `#create_repo`
* Moved poll interval out into a constant
* Merged ActionCable fix in #1389 
* Merged migration in #1386 

## Things to do in followup PRs
* Fix concurrency issue with `CreateGitHubRepositoryJob`
* Add support for group assignments
* Support ActionCable instead of short polling

Can I get some final 👀 on this? @d12

PR #1396, opened 2018-07-13
Fix concurrency issue with CreateGitHubRepositoryJob

### LINKED ISSUE #1392: Fix concurrency issue with CreateGitHubRepositoryJob
### Enhancement :sparkles:
`CreateGitHubRepositoryJob` retries with exponential backoff and if we add a manual retry button to the `setupv2.html.erb` we could run into a problem where we enqueue 2 of the same jobs.

It would be optimal if we guard on the `invitation_status` as the jop starts performing to prevent creating a GitHub repo twice.

In response to the post PR tasks in #1385

BenEmdon:
## What
The concurrency problem with `CreateGitHubRepositoryJob` is that you could kick off two jobs that could create two repos. We want `CreateGitHubRepositoryJob` to only create one repo always.

## How

We do this by short circuiting the job and returning early if the invitation status is:
* `unaccepted` (if the invitation wasn't accepted)
* `creating_repo` (if a job is already creating a repo)
* `importing_starter_code`(if starter code is being imported already)
* `completed` (if a repo already exists)

Closes #1392

### DIFF
--- a/app/jobs/assignment_repo/create_github_repository_job.rb
+++ b/app/jobs/assignment_repo/create_github_repository_job.rb
@@ -19,6 +19,9 @@ class AssignmentRepo
     def perform(assignment, user)
       start = Time.zone.now
 
+      invitation_status = assignment.invitation&.status
+      return if %w[unaccepted creating_repo importing_starter_code completed].include?(invitation_status)
+
       assignment.invitation&.creating_repo!
 
       creator = Creator.new(assignment: assignment, user: user)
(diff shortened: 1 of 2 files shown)


app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -19,6 +19,9 @@ class CreateGitHubRepositoryJob < ApplicationJob
     def perform(assignment, user)
       start = Time.zone.now
 
+      invitation_status = assignment.invitation&.status
+      return if %w[unaccepted creating_repo importing_starter_code completed].include?(invitation_status)
```
BenEmdon:
```ruby
invitation_status = assignment.invitation&.status
return if %w[unaccepted creating_repo importing_starter_code completed].include?(invitation_status)
assignment.invitation&.creating_repo!
```

Should this happen inside an ActiveRecord transaction?

d12: No I think we're good, I don't see any reason to have a transaction here

d12:
Looks great! 🚢

PR #1400, opened 2018-07-16
Poll harder better longer stronger

### LINKED ISSUE #1399: Long imports cause importing to hang
Really long imports causes importing to hang on the `setupv2` screen.

For example importing `rails/rails` takes a couple of minutes but before the import completes the `PorterStatusJob` has already timed out a max number of times.

BenEmdon:
## What
This PR introduces some enhancements to the Octopoller API and to how we poll from the porter status job. The enhancements introduced aim to fix the problem where long imports would time out (fixes #1399).

## Octopoller enhancements
The Octopoller API now looks as follows:
```diff
result = Octopoller.poll(wait: 1, timeout: 10) do 
  if we_are_there_yet() # waits 1 second between asking for a max of 10 seconds 
    "We are here, finally!"
  else
-   raise "No, we aren't there yet!"
+   :re_poll
  end
end

# result => "We are here, finally!"
```
Before this change we would have to `raise` an error to re poll, which didn't feel right.
The new way of re polling is to return the symbol `:re_poll` for Octopoller to re poll.

I also did away with the `error_handler` proc argument which was redundant:
```diff
-  def poll(wait: 1, timeout: 15, error_handler: nil)
+  def poll(wait: 1, timeout: 15)
```

## `PorterStatusJob` enhancements
`PorterStatusJob` now uses the new Octopoller API and has different logic for retrying.

Instead of retrying the job on a `Octopoller:TimeoutError`, the `TimeoutError` is instead rescued and kicks off a cascading job. This is the solution to long running imports since jobs will continuously be enqueued until the import has completed. My only worry is that this will flood our Sidekiq servers.


app/jobs/assignment_repo/porter_status_job.rb
```diff
@@ -2,50 +2,52 @@
...
-
-      result = Octopoller.poll(error_handler: error_handler) do
-        begin
-          progress = github_repository.import_progress[:status]
-          case progress
-          when GitHubRepository::IMPORT_COMPLETE
-            Creator::REPOSITORY_CREATION_COMPLETE
-          when *GitHubRepository::IMPORT_ERRORS
+      begin
+        result = Octopoller.poll(timeout: 30) do
```
d12: I like this tidier polling DSL a lot more 🙌 

The only thing I'd suggest to make this snippet even more concise is to use `timeout: 30.seconds` instead of `timeout: 30`. It's pretty obvious to us that this can't be minutes or ms, but being explicit is usually nice. Looking at the octopoller logic, I think passing `30.seconds` will work just fine.

lib/octopoller.rb
```diff
@@ -11,24 +11,18 @@ class TimeoutError < StandardError; end
   # error_handler - A proc that will run with each instance of an error
   # yield - A block that will execute, and if it raises an error it will re-run until success or the timeout is reached
   # raise - Raises an Octopoller::TimeoutError if the timeout is reached
-  #
-  # rubocop:disable MethodLength
-  def poll(wait: 1, timeout: 15, error_handler: nil)
+  def poll(wait: 1, timeout: 15)
```
d12: This can happen in another PR or really it doesn't have to happen at all, but I think some polling use cases would rather use a `max_retries` option instead of a `timeout` in seconds argument. E.g. "I want to poll a maximum of 5 times". In our case, seconds works fine, but I could see octokit wanting to use a maximum retries system. That's how some of their polling logic is currently implemented.

BenEmdon: I like that idea. Also a polling with exponential backoff would be nice. I'll definitely make that in a follow up PR.

PR #1403, opened 2018-07-17
Separate invitation status error into 2 statuses

### LINKED ISSUE #1402: Add a separate `invitation.status` for porter status error
Currently if the repo import step of the new resiliency beta fails, both the create repo and import status appear to failed:

screen shot 2018-07-16 at 2 55 21 pm

It would be clearer to show where exactly it failed by adding another status to the `invitation.status`

BenEmdon:
### New invitation status
This PR separates the invitation status `errorred` into `errored_creating_repo` and `errored_importing_starter_code`.

The reason for this is so that the user can have a transparent view into what went wrong in the import process:
jul-16-2018 17-34-11
Previously it would show to big red `X`s.

### Deletes existing AssignmentRepos on `#create_repo`
Modified the `create_repo` route to delete a repo if one was made already. This means that a failed repo will be replaced with a new one. Previously we were creating new repos that had the attempt appended to the end.
No more: `BenEmdon-assignment-1` -> `BenEmdon-assignment-1-1`:
Here's a gif that starts out with a failed repo and retries:
jul-16-2018 17-44-41

## Enhancements
Refactored the `setupv2.js` to have fewer custom actions per state.  

---
Closes #1402


app/controllers/assignment_invitations_controller.rb
```diff
@@ -52,22 +52,24 @@ def accept
   def setup; end
 
   def setupv2
-    render status: 404 unless import_resiliency_enabled?
+    return not_found unless import_resiliency_enabled?
```
d12: Is the `return` needed here?

app/controllers/assignment_invitations_controller.rb
```diff
@@ -52,22 +52,24 @@ def accept
...
   end
 
   # rubocop:disable MethodLength
   def create_repo
     if import_resiliency_enabled?
       job_started = false
       if current_invitation.accepted? || current_invitation.errored?
+        assignment_repo = AssignmentRepo.find_by(assignment: current_assignment, user: current_user)
+        assignment_repo&.destroy if assignment_repo
```
d12: This worries me just a little bit, if some bug in the current implementation or a bug introduced at a later date allows `create_repo` to be fired after a student has started working on their assignment, there's a chance we might destroy their in-progress assignment and all the work they pushed up to GitHub.

Do you have any ideas for how we might be able to be more confident here? One idea I came up with was querying for the number of commits on the repo if it exists, and only destroying if it's 0. it's a bit of overhead, but the case where there's an existing assignment_repo isn't the common path, and I'd rather pay a little bit for the confidence it'll give us. Nobody likes seeing their work destroyed :/

BenEmdon: I like the idea of querying the repo for the number of commits 👌 That extra layer of check is worth it.

srinjoym: Any chance there could be more than 0 commits if creating the repo fails?

BenEmdon:
## Update
Now the repo only deletes when the repo has 0 commits
```diff
   def create_repo
      if import_resiliency_enabled?
        job_started = false
        if current_invitation.accepted? || current_invitation.errored?
          assignment_repo = AssignmentRepo.find_by(assignment: current_assignment, user: current_user)
-         assignment_repo&.destroy if assignment_repo
+         assignment_repo&.destroy if assignment_repo&.github_repository&.empty?
          AssignmentRepo::CreateGitHubRepositoryJob.perform_later(current_assignment, current_user)
          job_started = true
        end
       render json: {
         job_started: job_started
       }
     else
       render status: 404, json: { error: "Not found" }
      end
    end
```

Can I get a final 👀 @d12

PR #1404, opened 2018-07-17
Classroom API for Classroom-Desktop

srinjoym:
Adds API for use with Classroom Desktop to fetch assignment and assignment repo information from Classroom. Also adds Download Repositories button to assignment and group assignment pages. API will likely be modified in the future as we develop Classroom-Desktop, but I wanted to get a preliminary PR in so that we can make smaller PRs in the future.

API Documentation: https://documenter.getpostman.com/view/2898223/RWMEP91P

screen shot 2018-07-17 at 3 13 04 pm


app/controllers/api/application_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+# rubocop:disable ClassAndModuleChildren
+class API::ApplicationController < ApplicationController
+  include Rails::Pagination
+
+  before_action :add_security_headers
+
+  def authenticate_user!
+    return become_active if logged_in? && adequate_scopes? && current_user.authorized_access_token?
+    render_forbidden
```
srinjoym: If user isn't logged in, returns 403 with unauthorized message. Might modify in future to redirect to OAuth URL.

app/controllers/api/application_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  private
+
+  def render_forbidden
+    render json: {
+      message: "Unauthorized"
+    }, status: :forbidden
+  end
+
+  def add_security_headers
```
srinjoym: Send security headers so we can access Pagination headers

app/serializers/assignment_repo_serializer.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AssignmentRepoSerializer < ActiveModel::Serializer
```
srinjoym: Attributes returned when Assignment Repo is rendered in json

app/serializers/group_assignment_repo_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+  attributes :id
+  attributes :username
+  attributes :repoUrl
+  attributes :displayName
+
+  def username
+    object.group.title
+  end
+
+  # rubocop:disable MethodName
```
srinjoym: Returning in camel case to keep with JSON standards

app/controllers/api/application_controller.rb
d12: :+1:

app/controllers/api/application_controller.rb
d12: I'll admit I'm not super well versed in Access-Control headers in Rails. This seems fine, but @mastahyeti would you mind :eyes: this?

app/controllers/api/assignment_repos_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def index
+      repos = AssignmentRepo.where(assignment: @assignment)
+      paginate json: repos
+    end
+
+    private
+
+    def set_assignment
+      @assignment = @organization.assignments.includes(:assignment_invitation).find_by!(slug: params[:assignment_id])
```
d12: Are we using `assignment_invitation` anywhere?

app/serializers/assignment_repo_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  # rubocop:disable MethodName
+  def repoUrl
+    object.github_repository.html_url
+  end
+
+  def displayName
+    object.user.github_user.name || ""
+  end
+  # rubocop:enable MethodName
```
d12: 3 API requests per repo 😱 

There's not much that can be done though, unless there's any key here we can remove.

app/serializers/group_assignment_repo_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+  def repoUrl
+    object.github_repository.html_url
+  end
+
+  def displayName
+    object.group.repo_accesses.map(&:user).map(&:github_user).map(&:login).join(", ")
```
d12: Whats `displayName` for? This is going to be extremely expensive to calculate.

app/views/assignments/show.html.erb
```diff
@@ -76,6 +84,7 @@
       </div>
     <% else %>
       <% if @assignment_repos.present? %>
+
```
d12: Could you remove this line from the diff?

spec/controllers/api/application_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe API::ApplicationController, type: :controller do
+  controller do
+    def index
+      render json: {}, status: :ok
```
d12: Any reason we're stubbing the controller actions in these tests?

If the serializers were changed or broken, we'd want the suite to fail for sure, and I don't think that's the case right now.

app/controllers/api/application_controller.rb
mastahyeti: What is going to be making XHR requests to this? This allows any origin to make a GET request to any URL, sending any headers. It seems like that is _probably_ more open than necessary, but I'm not sure how this API is being used.

app/serializers/group_assignment_repo_serializer.rb
BenEmdon: Nitpick: It's convention in Rails to use snake case in JSON. Take a look at GitHub's REST API for example https://developer.github.com/v3/

spec/controllers/api/assignment_repos_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe API::AssignmentReposController, type: :controller do
+  let(:organization)      { classroom_org                                                           }
+  let(:user)              { classroom_teacher                                                       }
```
BenEmdon: This is some interesting style 🤔 
Nitpick: How about this?
```ruby
let(:user)         { classroom_teacher }
let(:assignment)   { create(:assignment, organization: organization, title: "Learn Clojure") }
```
I feel this reads slightly more effortlessly.

spec/controllers/api/assignments_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    it "returns success" do
+      expect(response).to have_http_status(:success)
+    end
+
+    it "returns only one assignment" do
+      expect(json.length).to eql(1)
+    end
+
+    it "returns serialized versions of indivdual assignment" do
+      expect(json.first.to_json).to eq(@assignment_json)
```
BenEmdon: Nitpick: If you compare hashes instead of JSON test failures would be slightly easier to read

spec/support/json_helpers.rb
```diff
@@ -0,0 +1,5 @@
+module JsonHelpers
+  def json
+    JSON.parse(response.body)
+  end
+end
```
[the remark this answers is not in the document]
BenEmdon: Nitpick: this file is still missing a newline

app/controllers/api/group_assignments_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module API
+  class GroupAssignmentsController < API::ApplicationController
+    include ActionController::Serialization
+    include OrganizationAuthorization
+
+    before_action :set_assignment, except: :index
```
BenEmdon: Nitpick: instead of `except: :index` you could use `only: :show` to avoid any unwanted side effects of adding a new route later

app/controllers/api/application_controller.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# rubocop:disable ClassAndModuleChildren
+class API::ApplicationController < ApplicationController
+  include Rails::Pagination
+
+  prepend_before_action :ensure_download_repositories_flipper_is_enabled
```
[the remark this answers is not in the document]
srinjoym: Needed to use `prepend_before_action` here because I wanted to run this action before we run `authenticate_user!`. If I use a regular `before_action` then if the user is not logged in they get a forbidden message from `authenticate_user!` even if the flipper is disabled. 

I presume we want to show the 404 message if the flipper is not enabled as if the api doesn't exist right?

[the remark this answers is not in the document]
d12: Makes sense :+1:

app/controllers/api/application_controller.rb
srinjoym: Classroom-Desktop is making requests to this API to get student's assignment repos. I figured the pagination headers were ok to expose to everyone because they didn't have any sensitive data? What do you think?

app/serializers/group_assignment_repo_serializer.rb
srinjoym: I also used camel case here to make life on the Classroom-Desktop side (because everything there has to be camel case 🤦‍♂️ I think it makes sense to switch to snake case, but I'll do that as a separate PR in the future after finalizing the Classroom-Desktop code.

spec/controllers/api/application_controller_spec.rb
srinjoym: This controller is acting as an anonymous controller in the tests which essentially just means that it extends `API::ApplicationController`. I had to use the anonymous controller here because we don't actually have any actions in `API::ApplicationController` and I wanted to test the before actions of application controller.

spec/controllers/api/assignment_repos_controller_spec.rb
srinjoym: Good point, fixed 👍

app/serializers/assignment_repo_serializer.rb
srinjoym: We can probably return fewer things but I'll update that in a future PR after we finalized Classroom-Desktop!

app/controllers/api/application_controller.rb
mastahyeti: A desktop app isn't going to be bound by CORS and won't have any restrictions on what headers it can read. CORS defines hows one website is allowed to access resources from another website and is enforced by web browsers. If this API is only being used by a native app you probably don't need to make any CORS changes.

srinjoym: Was running into problems with CORS since Electron technically makes a request from a different origin which is why I added these headers originally, but I think I found a way around that while investigating the problem again.

I can turn off the same origin policy for the browser window by setting the `webSecurity` in the browser window options to false (https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions).

Is that a safer approach or should I keep `webSecurity` on and expose the response headers?

mastahyeti: What JavaScript networking APIs are you using in the Electron app? You should be able to use the [`net`](https://electronjs.org/docs/api/net) API instead of trying to make XHR requests.

srinjoym: @mastahyeti and I just discussed different options for authentication here, and agree that this will likely need to be converted to a token based API in a future PR. I can't use the `net` api since I'm currently using the browser session for authentication on the Classroom-Desktop app.

For this PR, I'm going to keep the CORS headers in so we can access the API from Classroom-Desktop without turning off `webSecurity` in Electron (as that poses other security risks).  I've created an issue on the appsec review to get more 👀 on this: https://github.com/github/appsec-reviews/issues/230

PR #1410, opened 2018-07-18
Remove old setup flow

BenEmdon:
## What 
Looks like it was using the wrong flow when there was starter code. 
This removes that old flow.


app/controllers/assignment_invitations_controller.rb
```diff
@@ -35,11 +35,7 @@ def accept
     else
       create_submission do
         GitHubClassroom.statsd.increment("exercise_invitation.accept")
-        if current_submission.starter_code_repo_id
-          redirect_to setup_assignment_invitation_path
-        else
-          redirect_to success_assignment_invitation_path
-        end
+        redirect_to success_assignment_invitation_path
```
d12: @benemdon Could you test this locally on a repo with starter code and without import resiliency enabled?

BenEmdon: ## With Starter code
jul-18-2018 11-22-36

## Without starter code
jul-18-2018 11-23-41

PR #1412, opened 2018-07-18
PorterStatusJob stat reporting

BenEmdon:
## What
This PR adds two new stats to report on the porter status job:
* `v2_exercise_repo.import.success`
* `v2_exercise_repo.import.fail`


app/jobs/assignment_repo/porter_status_job.rb
```diff
@@ -38,13 +38,15 @@ def perform(assignment_repo, user)
...
           assignment_repo.assignment.invitation&.completed!
           ActionCable.server.broadcast(
             RepositoryCreationStatusChannel.channel(user_id: user.id),
             text: result,
             status: assignment_repo.assignment.invitation&.status
           )
+          GitHubClassroom.statsd.increment("v2_exercise_repo.import.success")
         end
       rescue Octopoller::TimeoutError
         PorterStatusJob.perform_later(assignment_repo, user)
```
d12: Do you think it might be useful to report here as well?

BenEmdon: Yes! This also raises the question, should we be reporting retries on `CreateGitHubRepositoryJob`? I feel we should because technically the success rate of the new import resiliency pattern is `net success = success - (failure  - retries)`

BenEmdon:
> Do you think it might be useful to report here as well?
  @d12 

Yes! This also raises the question, should we be reporting retries on `CreateGitHubRepositoryJob`? I feel we should because technically the success rate of the new import resiliency pattern is `net success = success - (failure  - retries)`

d12:
It couldn't hurt!

PR #1425, opened 2018-07-23
New InviteStatus model 

BenEmdon:
## What
I found a raise condition with the new import repo flow. Currently the import flow relies on the `status` state on the `AssignmentInvitation` model. The problem is that there is a 1:N relationship between an `AssignmentInvitation` and a `User`.

This means two users would be modifying a shared `Invitation.status` attribute. This is bad since we use an `AssignmentInvitation`s `status` to display the progress of the repo create/import.

This PR introduces a new model `InviteStatus` which has a N:1 relationship with `AssignmentInvitation` and `User`. 

To clarify:
- `AssignmentInvitation has_many InviteStatuses`
- `User has_many InviteStatuses`
- `InviteStatus belongs_to AssignmentInvite`
- `InviteStatus belongs_to User`

This new model replaces the `status` state on `AssignmentInvitation` with the new `InviteStatus` model. This resolves the raise condition found.


BenEmdon:
Migration for this PR is in #1426

app/controllers/assignment_invitations_controller.rb
```diff
@@ -144,6 +140,10 @@ def current_invitation
     @current_invitation ||= AssignmentInvitation.find_by!(key: params[:id])
   end
 
+  def current_invitation_status
+    current_invitation.status(current_user)
```
d12: Can we memoize this? In it's current state, we're invoking `current_invitation#status` every time this method gets called.

```ruby
@current_invitation_status ||= current_invitation.status(current_user)
```

Or, if a current invitation status can be nil:

```ruby
return @current_invitation_status if defined?(@current_invitation_status)

@current_invitation_status = current_invitation.status(current_user)
```

app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -16,14 +16,18 @@ class CreateGitHubRepositoryJob < ApplicationJob
...
     # rubocop:disable AbcSize
+    # rubocop:disable CyclomaticComplexity
+    # rubocop:disable PerceivedComplexity
     def perform(assignment, user)
       start = Time.zone.now
 
-      return unless assignment.invitation&.waiting? \
-        || assignment.invitation&.errored_creating_repo? \
-        || assignment.invitation.status.nil?
+      invite_status = assignment.invitation&.status(user)
```
d12: Looking at the code below, we're returning if this is nil, but we're still guarding against nil with `&.`. What if right below this line, we added `return unless invite_status`, and removed all the nil guards?

app/jobs/assignment_repo/porter_status_job.rb
```diff
@@ -10,6 +10,8 @@ class PorterStatusJob < ApplicationJob
     def perform(assignment_repo, user)
       github_repository = assignment_repo.github_repository
 
+      invite_status = assignment_repo.assignment.invitation&.status(user)
```
d12: Same comment here as above. If there is no invitation, can we just abort the job?

app/models/assignment_invitation.rb
```diff
@@ -70,17 +59,14 @@ def enabled?
     assignment.invitations_enabled?
   end
 
-  def errored?
-    errored_creating_repo? || errored_importing_starter_code?
+  def status(user)
+    invite_statuses.find_by(user_id: user.id) \
+      || InviteStatus.create(user_id: user.id, assignment_invitation_id: id)
```
d12: Normally we put the `||` on the first line, then we don't need a `\`

spec/models/invite_status_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe InviteStatus, type: :model do
+  subject { InviteStatus }
+
```
d12: In addition to these tests, could you also add tests ensuring the ActiveModel relationships are working correctly?

We might also want tests in the user model spec asserting we can do `user.assignment_invitations` and that works correctly.

Also a test showing that if a user gets destroyed, it's invitestatuses get destroyed, but the assignment invitation itself does _not_ get destroyed would be great.

app/controllers/assignment_invitations_controller.rb
BenEmdon: Thanks for catching this.
`current_invitation` shouldn't be `nil` since its a 1:1 dependency on the `Assignment` model. Will go ahead with the former approach ☝️

BenEmdon:
## Update 

Added:
- Memoization for `current_invitation_status`
- Validation for the presence of an `AssignmentInvitation` on `Assignment`
- Validation for the presence of a `User` on `InviteStatus`

app/jobs/assignment_repo/create_github_repository_job.rb
BenEmdon: I instead validate the presence of an `AssignmentInvitation` on an `Assignment`. `Assignment`s are always created with an invitation:
https://github.com/education/classroom/blob/ebd4d33fc9eaeaba052f8b5716cc1acf92c64fc4/app/controllers/assignments_controller.rb#L14-L17

By validating the presence of an invitation we won't need to guard on the invitation anymore.

BenEmdon:
## Update 

Added:
- Memoization for `current_invitation_status` on `AssignmentInvitationController`
- Validation for the presence of an `AssignmentInvitation` on `Assignment` (and tests 305d36b)
- Validation for the presence of a `User` on `InviteStatus`
- Validation for the presence of a `AssignmentInvitation` on `InviteStatus`
- Validation for the uniqueness of `<User, AssignmentInvitation>` on `InviteStatus`
- Tests to ensure ActiveModel relationships are working correctly dbbe7c5

Can I get another review @d12

app/controllers/assignment_invitations_controller.rb
```diff
@@ -142,6 +142,10 @@ def current_invitation
     @current_invitation ||= AssignmentInvitation.find_by!(key: params[:id])
   end
 
+  def current_invitation_status
+    @current_invitation_status ||= current_invitation.status(current_user)
```
d12: :+1:

d12:
Migration is done:

```
-- create_table(:invite_statuses)
D, [2018-07-24T13:45:50.895762 #22] DEBUG -- :    (6.9ms)  CREATE TABLE "invite_statuses" ("id" bigserial primary key, "status" integer DEFAULT 0, "assignment_invitation_id" bigint, "user_id" bigint, "created_at" timestamp NOT NULL, "updated_
at" timestamp NOT NULL, CONSTRAINT "fk_rails_ea2e794dff"
FOREIGN KEY ("assignment_invitation_id")
  REFERENCES "assignment_invitations" ("id")
, CONSTRAINT "fk_rails_ecab83cd90"
FOREIGN KEY ("user_id")
  REFERENCES "users" ("id")
)
D, [2018-07-24T13:45:50.902777 #22] DEBUG -- :    (2.8ms)  CREATE  INDEX  "index_invite_statuses_on_assignment_invitation_id" ON "invite_statuses"  ("assignment_invitation_id")
D, [2018-07-24T13:45:50.909902 #22] DEBUG -- :    (2.8ms)  CREATE  INDEX  "index_invite_statuses_on_user_id" ON "invite_statuses"  ("user_id")
   -> 0.0241s
== 20180720232458 CreateInviteStatuses: migrated (0.0243s) ====================

D, [2018-07-24T13:45:50.917036 #22] DEBUG -- :   SQL (1.4ms)  INSERT INTO "schema_migrations" ("version") VALUES ($1) RETURNING "version"  [["version", "20180720232458"]]
D, [2018-07-24T13:45:50.920912 #22] DEBUG -- :    (3.5ms)  COMMIT
D, [2018-07-24T13:45:50.928873 #22] DEBUG -- :   ActiveRecord::InternalMetadata Load (1.4ms)  SELECT  "ar_internal_metadata".* FROM "ar_internal_metadata" WHERE "ar_internal_metadata"."key" = $1 LIMIT $2  [["key", "environment"], ["LIMIT", 1]
]
D, [2018-07-24T13:45:50.943937 #22] DEBUG -- :    (1.3ms)  BEGIN
D, [2018-07-24T13:45:50.946045 #22] DEBUG -- :    (1.3ms)  COMMIT
D, [2018-07-24T13:45:50.947576 #22] DEBUG -- :    (1.3ms)  SELECT pg_advisory_unlock(2076930691114369125)
```

PR #1428, opened 2018-07-23
Assignmentinvitation drop status column

BenEmdon:
## What
In #1425 the `status` attribute on `AssignmentInvitation` was moved to its own child model `InviteStatus`. Because of this there is no need for a `status` column on `assignment_invitations`. 
This PR proposes a migration to remove the `status` column.

This is failing until #1425 merges.


BenEmdon:
#1425 merged
Can I get a review :eyes: @d12

PR #1435, opened 2018-07-25
Don't hit the cache on polling progress

BenEmdon:
## What
Possibly why there is a discrepancy between the time of import completion and when the UI presents completion is that the `client.source_import_progress` is hitting the cache and thus is not necessarily up to date.

This PR proposes that we include the header in our import progress request: 
```ruby
{ "Cache-Control" => "no-cache, no-store" }
```


johndbritton:
This seems likely, good catch.

PR #1441, opened 2018-07-27
Added new metric for when adding a collaborator fails

BenEmdon:
## What
Added `v2_exercise_repo.create.adding_collaborator.fail` stat reporting for when `CreateGitHubRepositoryJob` fails to add collaborator.


app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -67,15 +67,20 @@ def perform(assignment, user)
...
         status: invite_status.status
       )
       logger.warn(err.message)
-      GitHubClassroom.statsd.increment("v2_exercise_repo.create.fail")
+      case err.message
+      when Creator::REPOSITORY_COLLABORATOR_ADDITION_FAILED
+        GitHubClassroom.statsd.increment("v2_exercise_repo.create.adding_collaborator.fail")
+      else
+        GitHubClassroom.statsd.increment("v2_exercise_repo.create.fail")
+      end
```
BenEmdon: Are there other cases that we should consider handling differently? Like when starting to import starter code fails?

app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -67,15 +67,20 @@ def perform(assignment, user)
         )
       end
     rescue Creator::Result::Error => err
-      creator.delete_github_repository(assignment_repo.try(:github_repo_id))
+      creator.delete_github_repository(assignment_repo&.github_repo_id)
```
d12: `.try` and `&.` do different things, could we remove this from the diff?

app/jobs/assignment_repo/create_github_repository_job.rb
d12: Generally, as much metric granularity as possible is good, because we can aggregate in datadog. I think this is fine for this PR.

d12:
@BenEmdon I left one comment, once that's sorted this is :+1:

PR #1442, opened 2018-07-28
Implement ActionCable

### LINKED ISSUE #1395: Use ActionCable instead of short polling in setupv2
### Enhancement :sparkles:
ActionCable is already mostly set up. 
Just need to implement it such that it reacts to messages from the `RepoCreationStatusChannel` instead of short polling on the `/progress` route. 

In response to the post PR tasks in #1385

BenEmdon:
## What
This PR implements ActionCable on client side in place of short polling. The backend ActionCable message broadcasting is already complete.

---

This closes #1395


app/assets/javascripts/setupv2.js
```diff
@@ -158,14 +158,6 @@
     location.href = success_path();
   };
 
-  check_progress = function() {
```
d12: Why don't we need this anymore?

BenEmdon: `check_progress` would make a `GET` request to get the progress of the of the ongoing create and import. It would make this request continuously. 
Instead of polling for the progress we are receiving the progress with `ActionCable`.

d12: Ah right, :+1:

PR #1446, opened 2018-07-30
Fix 404 bug

### LINKED ISSUE #1444: 404 after repo is deleted
**Describe the bug**
If an AssignmentRepos associated GitHub Repo is deleted the invitation 404s 

**To Reproduce**
1. Create an AssignmentRepo and GitHub repo by accepting an invitation
2. Delete the repo on GitHub
3. Go to the invitation page
4. See 404

**Expected behavior**
To see a invitation in setup try again

BenEmdon:
## What
If someone deleted their GitHub repository on .com, then they would get a 404 on classroom when requesting their invitation again.

## Wait when does this happen?
To clarify this happens in the following case:
screen shot 2018-07-30 at 11 28 58 am
> A user/admin deleted the repo on GitHub.com

## What exactly happens to people in this instance?
If they request the invitation link then they get a 404

## Why was this happing?
This was happening because we introduced `route_based_on_status` in #1438 
The `InviteStatus` would be `completed` but the repo wouldn't exist so the before action `ensure_submission_repository_exists` would throw a `not_found`.

## How can we fix this?
We can fix this by changing the `InviteStatus` from `completed` to `accepted` in the `ensure_submission_repository_exists`.

Closes #1444


BenEmdon:
## Update
In 9dbea25 I made a change so that if the `current_submission` doesn't exist it redirects to `setupv2` instead of `404`ing.

PR #1449, opened 2018-07-31
Automatic retries for creating a GitHub repo

BenEmdon:
## What
This PR proposes that we perform automatic retries when trying to create a GitHub repo. 

## Why
If we perform 3 retries before the user even sees the failure we can eliminate most repo creation problems. 

## How
This is done by adding a 3rd argument to `CreateGitHubRepositoryJob`.

---
Closes #1449


BenEmdon:
This is ready for review 👀 what do you think @d12

app/controllers/assignment_invitations_controller.rb
```diff
@@ -54,7 +54,7 @@ def create_repo
         GitHubClassroom.statsd.increment("v2_exercise_repo.import.retry")
       end
       current_invitation_status.waiting!
-      AssignmentRepo::CreateGitHubRepositoryJob.perform_later(current_assignment, current_user)
+      AssignmentRepo::CreateGitHubRepositoryJob.perform_later(current_assignment, current_user, 3)
```
d12: Reading this method invocation, clear that `current_assignment` is providing the assignment to the method, and `current_user` is providing the user, but it's difficult or impossible to infer the purpose of `3` without looking at the method definition. Could you make `retries` a kwarg in this method?

PR #1450, opened 2018-08-01
Classroom Desktop Token Authenticated API

srinjoym:
Updates to the Classroom Desktop API PR to switch to token authentication! Removed CORS headers which were exposing pagination headers because we can now use the `http` module on Classroom-Desktop to make the request.

#### User clicks download repositories button on Classroom
1. Redirects to `/desktop` route of assignment which generates a Temporary JWT(expiration of 5 minutes) with the id of the current user and and sends it to Classroom-Desktop with a deep-link
2. Classroom Desktop uses the temporary JWT and hits the `/login/oauth/access_token` endpoint which validates the temporary JWT and returns a JWT with a longer expiration period (24 hours) to use for cloning repositories

#### User logs in on Classroom Desktop
1. Opens auth window on Classroom-Desktop with node-integration off and hits `/login/oauth/authorize` which requires the user to login and then returns the temporary JWT over deep link (as if the button was clicked like Step 1 above)
2. Same as above, Classroom-Desktop exchanges temporary JWT for permanent JWT for cloning.


app/controllers/api/application_controller.rb
```diff
@@ -5,26 +5,32 @@ class API::ApplicationController < ApplicationController
...
   def authenticate_user!
     return become_active if logged_in? && adequate_scopes? && current_user.authorized_access_token?
     render_forbidden
   end
 
+  def verify_api_token
+    if params[:access_token].present?
+      data = JsonWebToken.decode(params[:access_token])
+      unless data.nil? || data[:user_id].nil?
+        return session[:user_id] = data[:user_id]
```
srinjoym: Check that JWT is valid and then set current user to the user in the token

lib/json_web_token.rb
```diff
@@ -0,0 +1,26 @@
...
+class JsonWebToken
+  class << self
+    def encode(payload, exp = 5.minutes.from_now)
+      payload[:exp] = exp.to_i
+      JWT.encode(payload, jwt_secret)
+    end
+
+    def decode(token)
+      body = JWT.decode(token, jwt_secret)[0]
+      HashWithIndifferentAccess.new body
```
d12: Is there a reason we need `HashWithIndifferentAccess`?

lib/json_web_token.rb
```diff
@@ -0,0 +1,26 @@
...
+      nil
+    end
+
+    private
+
+    def jwt_secret
+      @jwt_secret ||= Rails.application.secrets.jwt_secret
+      return @jwt_secret if @jwt_secret.present?
+
+      raise "JWT_SECRET is not set, please check you .env file"
```
d12: typo on `you`!

lib/json_web_token.rb
```diff
@@ -0,0 +1,26 @@
...
+  class << self
+    def encode(payload, exp = 5.minutes.from_now)
+      payload[:exp] = exp.to_i
+      JWT.encode(payload, jwt_secret)
+    end
+
+    def decode(token)
+      body = JWT.decode(token, jwt_secret)[0]
+      HashWithIndifferentAccess.new body
+    rescue StandardError
```
rzhade3: Should we be rescuing a specific kind of error here? This error handling would not give us any information about the specific JWT error.

lib/json_web_token.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class JsonWebToken
+  class << self
+    def encode(payload, exp = 5.minutes.from_now)
+      payload[:exp] = exp.to_i
+      JWT.encode(payload, jwt_secret)
```
rzhade3: Might want to be explicit about the hashing algorithm that is being used to encode the JWT

lib/json_web_token.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def decode(token)
+      body = JWT.decode(token, jwt_secret)[0]
```
rzhade3: Likewise here, might want to specify the hashing algorithm that is used: [Verifying the cryptographic signature](https://github.com/jwt/ruby-jwt#algorithms-and-usage)

app/controllers/oauth_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  prepend_before_action :ensure_download_repositories_flipper_is_enabled_globally, only: :access_token
+
+  skip_before_action :authenticate_user!, only: :access_token
+  skip_before_action :verify_authenticity_token, only: :access_token
+
+  def authorize
+    not_found if params[:redirect_uri].blank?
+    code = current_user.api_token
+    code_param = { code: code }.to_param
+    redirect_to "#{params[:redirect_uri]}?#{code_param}"
```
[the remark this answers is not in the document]
rzhade3: Is there a specific use case which requires the use of a dynamic redirect uri? If not, could we hard code this, or, as mentioned by @d12, perform some sort of validation on the uri?

app/controllers/assignments_controller.rb
```diff
@@ -64,6 +64,11 @@ def destroy
     end
   end
 
+  def desktop
+    code = current_user.api_token
```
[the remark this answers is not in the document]
srinjoym: Using `code` here to refer to the temporary token that's passed before the app sends the actual `access_token` over. Following the terminology that we use for auth on [dotcom](https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/#2-users-are-redirected-back-to-your-site-by-github).

Not sure if it could be confusing, what do you think?

lib/json_web_token.rb
srinjoym: Just allows me to use symbols to access the data after it's decoded. Is it fine to use here?

app/controllers/oauth_controller.rb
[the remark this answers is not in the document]
srinjoym: No use case right now for this to be dynamic, was following the dotcom structure of having a `redirect_uri` for future support. I can add in checks to make sure it's in the format of a URL or just hardcode it.

What do you guys think?

lib/json_web_token.rb
srinjoym: Didn't check the error type here because I don't return any special response if the token is expired vs. it just not being valid, both return a 403 response code. Is that fine?

lib/json_web_token.rb
srinjoym: Any suggestions for which one to use? Not sure what the differences are so I just used the default one

app/controllers/oauth_controller.rb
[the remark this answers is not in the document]
rzhade3: If there's no functionality to be gained from using a dynamic redirect URL being used here, I would rather if it was hardcoded. I can't think of many attack vectors that could exploit this, but better safe than sorry 😛

lib/json_web_token.rb
d12: It may be worth distinguishing between the two for logging purposes. We could keep metrics on the different types of decode failure types.

app/controllers/group_assignments_controller.rb
```diff
@@ -65,6 +65,11 @@ def destroy
     end
   end
 
+  def desktop
+    code = current_user.api_token
+    redirect_to "x-github-classroom://?assignment_url=#{organization_group_assignment_url}&code=#{code}"
```
d12: Is it possible for a group assignment URL or a code to contain a `&` or a `+`? These characters might break the URL.

lib/message_verifier.rb
```diff
@@ -0,0 +1,32 @@
...
+  class << self
+    def encode(payload, exp = 5.minutes.from_now)
+      payload[:exp] ||= exp
+      verifier.generate payload
+    end
+
+    def decode(token)
+      body = verifier.verify(token)
+
+      raise ActiveSupport::MessageVerifier::InvalidSignature if Time.current > body[:exp]
```
d12: Whats the point of raising here, then rescuing and throwing away the error?

rzhade3: `body = verifier.verify(token)` returns the same error if the signature is invalid, and that line is to make sure that the signing has not expired, so that we can cleanly return nil if the token is either tampered with or expired. 

Although, it might be better practice to put everything into a `begin... rescue` block.

srinjoym: Yea was raising the same error instead of adding a separate ‘return nil’. Is raising an error and then catching it bad practice?

d12: Ah I see. An expired token doesn't seem like an invalid signature error, could we make a new error class `TokenExpired` or something and then raise that, then in the rescue, catch both? We can just return nil in the rescue, but this lets us do something different depending on how we failed later on.

PR #1455, opened 2018-08-02
Group.users relation

BenEmdon:
## What
This PR adds a way to get the users in a group.

## How
The way we were doing it was through a `Group`'s `RepoAccess`. This PR just makes that a formal `has_many` relationship.

Part of #1456


BenEmdon:
Friendly bump @d12 👋🙂

PR #1466, opened 2018-08-08
Fix grouping scope issue w/ the Download Roster button

d12:
@BenEmdon @srinjoym for review

### DIFF
--- a/app/controllers/orgs/rosters_controller.rb
+++ b/app/controllers/orgs/rosters_controller.rb
@@ -12,7 +12,6 @@ module Orgs
 
     helper_method :current_roster, :unlinked_users
 
-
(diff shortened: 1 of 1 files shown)


app/controllers/orgs/rosters_controller.rb
```diff
@@ -169,6 +171,12 @@ def ensure_enough_members_in_roster
       redirect_to roster_url(current_organization)
     end
 
+    def ensure_allowed_to_access_grouping
+      return if params[:grouping].nil?
+
+      not_found unless Grouping.find(params[:grouping]).organization_id == current_organization.id
+    end
```
BenEmdon: 👍

PR #1468, opened 2018-08-12
Fix error flash on assignment acceptance

BenEmdon:
# what 

Looks like assignment acceptance failures weren't flashing (or recording statsd). The problem was that we were using the wrong symbol in a switch case:
```diff
+ when :failed
- when :error
```

The symbol comes from `AssignmentRepo::Creator::Result`:
https://github.com/education/classroom/blob/f25b598a22e578b0ef16af57bfef29ec4b371b4d/app/models/assignment_repo/creator.rb#L41-L43


spec/controllers/assignment_invitations_controller_spec.rb
```diff
@@ -276,6 +276,29 @@
       patch :accept, params: { id: invitation.key }
     end
 
+    context "redeem returns an fail" do
```
d12: :heart: Great tests :)

PR #1470, opened 2018-08-14
Add support for private cloning with tokens and User endpoint for authentication

srinjoym:
Added separate `clone_url` method for assignment repositories so we can query that (and get the temporary token) right before downloading on Classroom-Desktop.

Also added `UsersController` to the API side which just returns the username of the current authenticated user, this is used by Desktop on start-up to check if the current user's token is still valid.


.vscode/launch.json
```diff
@@ -0,0 +1,64 @@
+{
```
d12: Did you mean to commit this file?

spec/controllers/api/assignment_repos_controller_spec.rb
```diff
@@ -35,13 +35,34 @@
...
-        expect(json.first["repoUrl"]).to eq(@assignment_repo.github_repository.html_url)
-      end
-
       it "returns user display name" do
         expect(json.first["displayName"]).to eq(user.github_user.name)
       end
     end
   end
+
+  describe "GET #clone_url", :vcr do
```
d12: Can you add tests in here for cases where the user is not authorized to fetch the clone_url?

spec/controllers/api/group_assignment_repos_controller_spec.rb
```diff
@@ -33,13 +33,35 @@
...
-        expect(json.first["repoUrl"]).to eq(@group_assignment_repo.github_repository.html_url)
-      end
-
       it "returns empty string as display name" do
         expect(json.first["displayName"]).to eq("")
       end
     end
   end
+
+  describe "GET #clone_url", :vcr do
```
d12: Same comment as above here

spec/controllers/api/assignment_repos_controller_spec.rb
srinjoym: Are you talking about the case where they have an invalid access token for the API? I test that functionality in the API Application Controller tests, but I could add it here too

app/models/github_repository.rb
```diff
@@ -210,6 +210,20 @@ def self.find_by_name_with_owner!(client, full_name)
     end
   end
 
+  def temp_clone_url
+    GitHub::Errors.with_error_handling do
+      repository = client.repository(
+        @id,
+        accept: "application/vnd.github.daredevil-preview+json"
```
BenEmdon: 👹

spec/controllers/api/assignment_repos_controller_spec.rb
BenEmdon: I think an classroom API level test would be good. That way we can explicitly check how the API will behave when the user isn't authorized.

spec/controllers/api/assignment_repos_controller_spec.rb
```diff
@@ -35,13 +35,34 @@
...
+    after do
+      AssignmentRepo.destroy_all
+    end
+
+    it "returns success" do
+      expect(response).to have_http_status(:success)
+    end
+
+    it "returns json with temp clone url" do
+      expect(json["temp_clone_url"]).to_not be_nil
```
BenEmdon: It would be nice to compare the entire json blob here if possible. That way we can easily look at the tests to see what the API endpoint returns.

spec/controllers/api/users_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(response).to have_http_status(:success)
+    end
+
+    context "user serializer returns correct attributes" do
+      it "returns user id" do
+        expect(json["id"]).to eq(user.id)
+      end
+
+      it "returns username" do
+        expect(json["username"]).to eq(user.github_user.login)
```
BenEmdon: These tests don't need to separated as they are just checking the same json hash. Would be easier to read and write a test that checks the entire json blob. Something like this:
```ruby
describe "success" do
  it "has a status of 200" do
    expect(response).to have_http_status(:success)
  end

  it "returns a structured body" do
    expect(json).to eq(
      "id": user.id,
      "username": user.github_user.login
    )
  end
end
```

spec/controllers/api/users_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:user) { classroom_teacher }
+
+  describe "GET #authenticated_user", :vcr do
+    before do
+      get :authenticated_user, params: {
+        access_token: user.api_token
+      }
+    end
+
+    it "returns success" do
```
BenEmdon: I think its worth adding a failure case too 🛑

spec/controllers/api/assignment_repos_controller_spec.rb
srinjoym: When I made the initial PR for the API, I added authorization tests for `API::ApplicationController` which all of the API controllers extend.

Here are the tests: https://github.com/education/classroom/blob/e350989f35b8e47565f5d240c7e28b346cd54f99/spec/controllers/api/application_controller_spec.rb#L40-L63

I think these are enough to test behavior when the User is not authorized. What do you guys think?

spec/controllers/api/assignment_repos_controller_spec.rb
srinjoym: Great idea!! Chatted with @d12 about this too, and ended up doing a similar thing to what you suggested. Combined all of JSON checks into one test but still kept them as individual `expect` calls so we know which attribute is failing from the tests!

spec/controllers/api/users_controller_spec.rb
srinjoym: Couldn't really think of any failure cases here other than the user not being authorized which is already tested as I mentioned above

PR #1471, opened 2018-08-14
Add launch modal and detect if Classroom-Desktop is installed 

srinjoym:
Detect if Classroom-Desktop is installed through some JS magic, works across all browsers I've tested so far (unlike dotcom's approach which doesn't work on Safari, might want to switch over to a method like this?)

Automatically redirects to desktop landing page if app is not installed, and links to page just in case the deep link does not open. 

Had to upgrade primer to latest version for the dropdown to work, also had to import the dropdown css directly from primer because it hasn't been added to the core repo yet.

#### Initial State
screen shot 2018-08-14 at 11 26 45 am

#### After `Open in Classroom-Desktop` is clicked
screen shot 2018-08-14 at 11 26 45 am


app/assets/javascripts/desktop-protocol-handler.js
```diff
@@ -0,0 +1,44 @@
...
+
+    var el = $(this);
+    var iFrame = $('#js-hidden-iframe')[0];
+
+    timeIndex = setTimeout(function() {
+      window.location = "http://www.classroom.github.com/desktop";
+    }, 3000);
+
+    // attempt to open deep link in iframe to avoid exposing link to user
+    iFrame.contentWindow.location.href = el.attr("href");
```
srinjoym: Attempting to open the deep link in a hidden iframe so that Safari does not throw an error and display its own page if it does not exist.

app/assets/javascripts/desktop-protocol-handler.js
```diff
@@ -0,0 +1,44 @@
+(function() {
+
+  var timeIndex
+
+  window.addEventListener("blur", function(e)
+  {
+    clearTimeout(timeIndex);
```
srinjoym: If we get the `blur` event on the window, we assume that the browser successfully opened the app and clear the timeout for redirecting to the desktop landing page. This is similar to the approach that Slack uses and I learned about it from here: https://github.com/electron/fiddle/issues/43#issuecomment-409032237

PR #1474, opened 2018-08-15
Add Srinjoy and Ben to codeowners and README

d12:
cc @srinjoym @BenEmdon


d12:
Removed Vanessa's ping as this isn't a copy change

PR #1478, opened 2018-08-15
Has many group invite statuses

### LINKED ISSUE #1456: GroupInviteStatus model needed
## What
We need a way to track progress of repo setup. This should be done with a 1:N:1 relationship of `Group:GroupInviteStatus:GroupInvitation`.

Part of #1452

BenEmdon:
## What
This PR proposes changes to add `has_many :group_invite_statuses` to both the `Group` and `GroupAssignmentInvitation` models.

## What does this look like?
frame 4
_The changes correlate to the relationships labeled with a blue `N`_

Closes #1456


app/models/group_assignment_invitation.rb
```diff
@@ -42,6 +43,11 @@ def enabled?
     group_assignment.invitations_enabled?
   end
 
+  def status(group)
+    group_invite_statuses.find_by(group: group) ||
```
srinjoym: 🎨 just a stylistic thing: might want to refactor this method to follow the structure of other methods in this file, something like this:

```
group_invite_status = group_invite_statuses.find_by(group: group)
return group_invite_status if group_invite_status.present?

GroupInviteStatus.create(group: group, group_assignment_invitation: self)
```

spec/models/group_spec.rb
```diff
@@ -75,4 +75,23 @@
...
+    let(:grouping)     { create(:grouping, organization: organization) }
+    let(:group)        { Group.create(grouping: grouping, title: "#{Faker::Company.name} Team") }
+    let(:invitation)   { create(:group_assignment_invitation) }
+
+    it "returns a list of invite statuses" do
+      group_invite_status = GroupInviteStatus.create(group: group, group_assignment_invitation: invitation)
+      expect(group.group_invite_statuses).to eq([group_invite_status])
+    end
+
+    it "on #destroy destroys invite status and not the invitation" do
```
srinjoym: Does it make sense to add a test here that makes sure that all the `GroupInviteStatuses` of a group get destroyed when the group is destroyed?

app/models/group_assignment_invitation.rb
BenEmdon: This is styled like status method on `assignment_invitation.rb`
https://github.com/education/classroom/blob/9d38c3649d0d9c370f63ba1c464d23195a1044ea/app/models/assignment_invitation.rb#L62-L65

Should I re-style that too?

srinjoym: ooh I see 🤔 imo, the second style is a bit easier to read even though it's more code, what do you think?

If you do refactor this method, it makes sense to refactor the one in `assignment_invitation.rb` too 👍

spec/models/group_spec.rb
srinjoym: Yea agreed we would just be testing native functionality, I was thinking it might be nice to document the `dependent: :destroy` relationship just in case of future breaking changes?

Issue #1487, opened 2018-08-17
Create new routes for GroupAssignmentInvitationsController


BenEmdon:
## Routes we want:
- [x] `setup`
  - [x] implemented controller action
  - [x] `setup.html.erb`
  - Implement javascript in different PR
- [x] `create_repo`
  - implemented controller action
  (implementation requires a new background job for repo creation)
- [x] `progress`
  - [x] implemented controller action

https://github.com/education/classroom/issues/1452#issuecomment-410400388

PR #1486, opened 2018-08-17
Remove repository setup dead code from GroupAssignmentInvitationsController

BenEmdon:
## What
This PR proposes removing some old `repo_setup` code left by a previous intern. 
This PR doesn't remove the core logic of repo setup but rather removes it's feature flag hidden code path in `GroupAssignmentInvitationsController`.

It also removes:
* `app/assets/javascripts/setup-repo.js`
* `app/views/group_assignment_invitations/setup.html.erb`
* Unused cassettes
* `repo_setup` feature flag

PR #1488, opened 2018-08-17
GroupAssignmentInvitationController #setupv2

BenEmdon:
## What
This PR proposes a new route `setupv2` on `GroupAssignmentInvitationController`.

The route is implemented to render the same `setupv2.html.erb` as used in `AssignmentInvitationsController`.

The javascript for this page will come in a later PR once ActionCable is set up.

## Route Progress
- [x] `setup`
  - [x] implemented controller action
  - [x] `setup.html.erb`
  - Implement javascript in different PR

Part of #1487


BenEmdon:
Since #1486 merged, this is now ready for review
CC: @d12 @srinjoym

config/routes.rb
```diff
@@ -48,6 +48,9 @@
       get   :accept
       patch :accept_assignment
       patch :accept_invitation
+      get   :setupv2
+      post  :create_repo
+      get   :progress
```
d12: Why are there 3 new routes here, but only one new view/test file?

d12: Ah, I just reviewed the other PR and saw tests and controller actions defined there. I'd either clean up the diff or merge them together, merging this without the controller actions / views from the other PR would result in broken paths (that probably wouldn't be hit, but best to be tidy and keep things together.

PR #1489, opened 2018-08-17
GroupAssignmentInvitationController #create_repo and #progress routes

BenEmdon:
# Please merge #1488  first

## What
This PR proposes the following new routes on `GroupAssignmentInvitationController`: 
* `#create_repo`
* `#progress`

The `#create_repo` is not yet implemented and raises a `NotImplementedError`. I will implement the `#create_repo` route in a follow up PR after I have the background job needed to fulfill the routes purpose.

The `#progress` route is fully implemented.

## Route Progress
- [X] `create_repo`
  - implemented controller action
  (implementation requires a new background job for repo creation)
- [X] `progress`
  - [X] implemented controller action

Closes of #1487


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -93,6 +106,10 @@ def ensure_github_repo_exists
     create_group_assignment_repo(selected_group: group)
   end
 
+  def ensure_group_import_resiliency_enabled
+    render status: 404, json: { error: "Not found" } unless group_import_resiliency_enabled?
```
srinjoym: I think it might make more sense for this method to be in [`feature_flags_dependency.rb`](https://github.com/education/classroom/blob/7243fa161ce40927b4e1b76eab790e39c2ab52c3/app/controllers/application_controller/feature_flags_dependency.rb) since there are similar methods in that file already.

Also not sure if there's a reason you're rendering JSON for the error, should just be able to use the `not_found` helper?

```
def ensure_group_import_resiliency_enabled
     render not_found unless group_import_resiliency_enabled?
end
```

PR #1491, opened 2018-08-19
Route based on invite status

### LINKED ISSUE #1490: In GroupAssignmentsInvitationsController route based on invite status
## What
Apply the same idea of routing based on invite status in `AssignmentsInvitationsController` to `GroupAssignmentsInvitationsController`.

BenEmdon:
## What
This PR proposes the before action `route_base_on_status` which routes the user to the right page given the group's invite status.

## What does this look like?
This diagram depicts what was happening before:
> 

This diagram depicts what should be happening after this change:
> 

Resolves #1490


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -2,15 +2,18 @@
...
-  before_action :check_user_not_group_member,            only: [:show]
-  before_action :check_should_redirect_to_roster_page,   only: [:show]
-  before_action :authorize_group_access,                 only: [:accept_invitation]
-  before_action :ensure_github_repo_exists,              only: [:successful_invitation]
+  before_action :route_based_on_status,                  only: %i[setupv2 successful_invitation]
+  before_action :check_group_not_previous_acceptee,      only: :show
+  before_action :check_user_not_group_member,            only: :show
+  before_action :check_should_redirect_to_roster_page,   only: :show
+  before_action :authorize_group_access,                 only: :accept_invitation
+  before_action :ensure_github_repo_exists,              only: :successful_invitation
```
BenEmdon: Only real change here is L11
```diff
+ before_action :route_based_on_status,                  only: %i[setupv2 successful_invitation]
```
The rest of the changes in this diff are removing square brackets when redundant.

PR #1499, opened 2018-08-24
Redeem with group import resiliency

### LINKED ISSUE #1492: Make GroupAssignmentInvitations #redeem change the invitation.status(group)

BenEmdon:
## What ✨ 
This PR proposes a new flow for group assignment redemption. 

Steps that happen happen in `GroupAssignmentInvitation#redeem_for`:
```diff
+ fail early if invitation is no longer enabled
create a RepoAccess for the invitee
if Group does not exist -> create Group
if GroupAssignmentRepo exists -> return GroupAssignmentRepo
- if GroupAssignmentRepo does not exist -> create GroupAssignmentRepo
+ if GroupInviteStatus is unaccepted? -> set GroupInviteStatus to accepted
+ return pending state
```

Instead of creating a `GroupAssignmentRepo` in `GroupAssignmentInvitation#redeem_for` we defer the repo creation so that it can be done in a background job (coming in a later PR).

I introduced the class `GroupAssignmentInvitation::Result` so that the `redeem_for` could return either of the following states:
```ruby
.success(repo) # redeem_for succeeded and an AssignmentRepo already exists
.failed(error) # redeem_for failed with an error
.pending       # redeem_for succeeded and a repo is not yet created
```

This new flow is behind the `group_import_resiliency_enabled?` feature flag.

--- 
Closes #1492


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -145,20 +145,25 @@ def validate_max_members_not_exceeded!(group)
...
+    result = invitation.redeem_for(
+      current_user,
+      selected_group,
+      new_group_title,
+      group_import_resiliency: group_import_resiliency_enabled?
+    )
+
+    case result.status
+    when :failed
+      GitHubClassroom.statsd.increment("group_exercise_invitation.fail")
```
srinjoym: Small note here: We're now counting invitations not being enabled as a failure, whereas before we weren't. Not sure that case should count as a failure, what do you think?

srinjoym: Small note here: We're now counting invitations not being enabled as a failure, whereas before we weren't. Not sure that case should count as a failure, what do you think?

spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -13,9 +13,10 @@
     }
     create(:group_assignment, options)
   end
-  let(:grouping) { group_assignment.grouping }
+  let(:grouping)   { group_assignment.grouping }
+  let(:group_name) { "#{Faker::Company.name} Team" }
   let(:group) do
-    group = Group.create(grouping: grouping, title: "#{Faker::Company.name} Team")
+    group = Group.create(grouping: grouping, title: group_name)
```
srinjoym: You could possibly get rid of the the `group_name` variable here by defining a new `group` factory in [`factories.rb`](https://github.com/education/classroom/blob/master/spec/factories.rb), something like this:

```
factory :group do
  title { Faker::Company.name }
  slug  { title.parameterize  }
  github_id  { rand(1..1_000_000) }
end
```

app/controllers/group_assignment_invitations_controller.rb
BenEmdon: You're totally right 👍 We don't want to count disabled invitations as a failure. Good catch 👌

BenEmdon:
Addressed @srinjoym's comments. Could I get another 👀

PR #1504, opened 2018-08-27
Dont report octokit 404s

BenEmdon:
## What
This PR proposes that we don't record 404s to haystack, since they aren't very helpful exceptions. 
The `AssignmentsController` and `GroupAssignmentsController` generate an overwhelming number of these exceptions, which just creates a lot of noise in our exceptions firehose.


lib/github/errors.rb
```diff
@@ -8,7 +8,7 @@ class << self
       def with_error_handling
         yield
       rescue Octokit::Error => err
-        Failbot.report!($ERROR_INFO)
+        Failbot.report!($ERROR_INFO) unless err.class == Octokit::NotFound
```
srinjoym: LGTM 👍 why are we getting 404s in the first place?

BenEmdon: We get TONS of 404s. They happen mostly when a resource on our end no longer exists on GitHub.com

Issue #1505, opened 2018-08-27
ActionView::Template::Error: undefined method `github_team' for nil:NilClass


BenEmdon:
**Describe the bug**

```
ActionView::Template::Error
undefined method `github_team' for nil:NilClass
```

<table>
  <thead>
    <tr>
      <th>attribute</th>
      <th>value</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>action</td>
      <td>show</td>
    </tr>
    <tr>
      <td>cause</td>
      <td><pre>NoMethodError: undefined method `github_team&#39; for nil:NilClass<br>	/app/app/models/group_assignment_repo.rb:52:in `github_team&#39;<br>	/app/app/models/group_assignment_repo.rb:44:in `disabled?&#39;<br>	/app/app/view_models/shared_assignment_repo_view.rb:11:in `disabled?&#39;<br>	/app/app/view_models/shared_assignment_repo_view.rb:39:in `disabled_class&#39;<br></pre></td>
    </tr>
    <tr>
      <td>class_name</td>
      <td>ActionView::Template::Error</td>
    </tr>
    <tr>
      <td>controller</td>
      <td>orgs/group_assignment_repos</td>
    </tr>
    <tr>
      <td>method</td>
      <td>GET</td>
    </tr>
    <tr>
      <td>params</td>
      <td>{}</td>
    </tr>
    <tr>
      <td>referrer</td>
      <td>https://classroom.github.com/classrooms/40213532-classroom-test-org-edon/group-assignments/ga-1</td>
    </tr>
    <tr>
      <td>ruby</td>
      <td>ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]</td>
    </tr>
    <tr>
      <td>session</td>
      <td>{}</td>
    </tr>
    <tr>
      <td>time</td>
      <td>0.209490492</td>
    </tr>
    <tr>
      <td>url</td>
      <td>https://classroom.github.com/classrooms/40213532-classroom-test-org-edon/group-assignments/ga-1/group_assignment_repos/83679</td>
    </tr>
    <tr>
      <td>user</td>
      <td>false</td>
    </tr>
    <tr>
      <td>user_agent</td>
      <td>Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.1.2 Safari/605.1.15</td>
    </tr>

  </tbody>
</table>

<details>
[...]

PR #1507, opened 2018-08-27
Revert "Delete team when it's deleted in GitHub"

BenEmdon:
Reverts education/classroom#1346

Would prevent future instances of #1505


BenEmdon:
The result of having #1346 in prod for a while is that we a a number of `GroupAssignmentRepo`s that are no longer valid.

Issue #1509, opened 2018-08-27
Creation job similar to `CreateGitHubRepositoryJob`


BenEmdon:
A job that creates an GroupAssignmentRepo and starts importing starter code

PR #1514, opened 2018-08-28
Delete Group Assignment Repo if anything goes wrong during setup

BenEmdon:
## What
This PR proposes we add a `delete_github_repository_on_failure` block around the before callbacks in [`app/models/group_assignment_repo.rb`](https://github.com/education/classroom/blob/478483d1cfc861ae91dac5c402ae5a5e71f707a1/app/models/group_assignment_repo.rb#L31-L32). 
This change makes it so that if _pushing starter code_ or _adding a team to the repository_ fails, the repo gets deleted as soon as possible.  

This is the first part of #1509


app/models/group_assignment_repo.rb
```diff
@@ -28,8 +28,10 @@ class GroupAssignmentRepo < ApplicationRecord
   before_validation(on: :create) do
     if organization
       create_github_repository
-      push_starter_code
-      add_team_to_github_repository
+      delete_github_repository_on_failure do
+        push_starter_code
+        add_team_to_github_repository
+      end
```
BenEmdon: This diff is the major change. To understand what this does look at the definition of  [`#delete_github_repository_on_failure`](https://github.com/education/classroom/blob/c358de13a4cc526daee9a5ec58d845c6d4e8564a/app/models/concerns/github_repoable.rb#L33-L40).
https://github.com/education/classroom/blob/c358de13a4cc526daee9a5ec58d845c6d4e8564a/app/models/concerns/github_repoable.rb#L33-L40

app/models/concerns/github_repoable.rb
```diff
@@ -56,9 +56,7 @@ def push_starter_code
     assignment_repository   = GitHubRepository.new(client, github_repo_id)
     starter_code_repository = GitHubRepository.new(client, starter_code_repo_id)
 
-    delete_github_repository_on_failure do
-      assignment_repository.get_starter_code_from(starter_code_repository)
-    end
+    assignment_repository.get_starter_code_from(starter_code_repository)
```
BenEmdon: This block ditches the use`#delete_github_repository_on_failure` so that it can instead be used in `app/models/group_assignment_repo.rb`

lib/github/search.rb
```diff
@@ -2,9 +2,6 @@
 
 module GitHub
   class Search
-    REPOSITORY_REGEX = /[a-zA-Z0-9\._-]+/
-    USERNAME_REGEX = /[a-zA-Z0-9_-]+/
```
BenEmdon: I just moved these regexes from the `GitHub::Search`module to the `GitHub` module.

spec/models/group_assignment_repo_spec.rb
```diff
@@ -35,43 +34,82 @@
...
+                github_repo = GitHubRepository.new(organization.github_client, @group_assignment_repo.github_repo_id)
+                add_github_team_url = github_url("/teams/#{group.github_team_id}/repos/#{github_repo.full_name}")
+                permission_param = { permission: "admin" }
+                expect(WebMock).to have_requested(:put, add_github_team_url)
+                  .with(body: hash_including(permission_param))
+              end
+            end
+          end
+        end
+
```
BenEmdon: The tests above this comment didn't change at all. They were just moved up into the context `success`.

BenEmdon:
> 👍 would be nice to switch individual assignment repos to use github_repoable in the future

Totally agree 🙌 `GitHubRepoable` is better documented and cleanly implemented in comparison to `AsignmentRepo::Creator`.

PR #1519, opened 2018-08-29
Add GroupAssignmentRepo::CreateGitHubRepositoryJob

### LINKED ISSUE #1509: Creation job similar to `CreateGitHubRepositoryJob`

BenEmdon:
## What
This PR Proposes a new class `GroupAssignmentRepo::CreateGitHubRepositoryJob` which is a background job responsible for creating a `GroupAssignmentRepo` with it's associated GitHub repo.

This new job class is modeled to behave very similarly (as similarly as possible). What the job does is does is as follows:
```ruby
# Creates a GroupAssignmentRepo and an associated GitHub repo. 
# Starts a source import if specified.
#
# group_assignment - The assignment that will be used to create the GroupAssignmentRepo
# group            - The group that the GroupAssignmentRepo will belong to
# retries          - The number of times the job will automatically retry
#
def perform(group_assignment, group, retries: 0)
```

Closes #1509


BenEmdon:
## Update 
#1514 and #1501 merged so this is ready for some 👀 
CC: @srinjoym @d12

PR #1520, opened 2018-08-29
#create_repo on GroupAssignmentInvitationsController 

BenEmdon:
## What
This PR implements the `#create_repo`on `GroupAssignmentInvitationsController`.
The purpose of this route is the same as `#create_repo` in [`group_assignment_invitations_controller.rb`](https://github.com/education/classroom/blob/master/app/controllers/assignment_invitations_controller.rb#L47-L66).

### It's purpose
When you `POST #create_repo` the following happens:
- if invite_status is `accepted?` or `errored?`
  - delete the `GroupAssignmentRepo` `if present?`
  - report retry Datadog stat `if errored?`
  - kick off a `GroupAssignmentRepo::CreateGitHubRepositoryJob`
  - set the invite_status to `waiting`
- otherwise
  - do nothing and return status


BenEmdon:
## Update 
#1519 merged so this is ready for review 👀

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -41,9 +41,30 @@ def setupv2
...
+        end
+        report_retry
+        group_invite_status.waiting!
+        GroupAssignmentRepo::CreateGitHubRepositoryJob.perform_later(group_assignment, group, retries: 3)
+        true
+      else
+        false
+      end
+    render json: {
+      will_create_repo: will_create_repo,
```
srinjoym: Is the field here different from the `job_started` field in `create_repo` in `AssignmentRepoInvitationsController`? Might be easier to follow in the future if they both use the same field name?

BenEmdon: TL;DR : I will change it back to `job_started`

The field is the exact same idea. At the time of writing this I thought `will_create_repo` sounded more descriptive and less confusing but now I feel that `job_started` is better because it's context is from within the `PUT /create_repo` route. I will change it back to `job_started`.

Issue #1543, opened 2018-09-05
Students suddenly cannot import repo into their assignments after many students completed their assignments successfully


bvelez:
Please choose either _issue_ or _feature request_

##
#### What you were trying to do ?
Offer a coding lab section using a GitHub Classroom assignment

#### What happened (include screenshot if you can) ?
After many students were able to create their assignment, Classroom begin giving this error:

![githubclassroomerror](https://user-images.githubusercontent.com/5605641/45124109-1ea13880-b137-11e8-8660-999685807a0e.png)

And did not allow any more students to create their assignments.

#### Step-by-step reproduction instructions

The link to my classroom is: https://classroom.github.com/classrooms/42574113-uprm-ciic4010-advanced-programming-fall-2018

Try to create assignment using this invitation link: https://classroom.github.com/a/CMps_6d3

##
### Feature request :sparkles:
_Please provide a description of the feature you would like Classroom to have_

BenEmdon:
Hey @bvelez 👋  thanks for writing in.
It appears that your classroom accepted all the assignments quickly leading to your token getting rate limited.

If you get the students affected to accept their assignments an hour after an hour the problem may go away. Keep me updated if this problem persists. We are looking into a solution now.

BenEmdon:
We have identified a possible work around. @bvelez you will have to try the following steps:
* log out of GitHub.com
* go to https://import.github.com/login and login again

This should regenerate your import token and allow your students to continue importing.
Please let me know if this works.

gkapfham:
Hello @BenEmdon, thanks for posting this workaround. Someone from GitHub Education also sent me the same point by Twitter. Again, thanks. I start my laboratory session in a few minutes and thus I have to be brief. I have some questions which I will write out separately:

- To clarify, is it enough for me to go to this mentioned site and login or do I actually need to do an import for each student in the class who is having a problem? (If needed, asking introductory students to do an import and to use the correct name is likely to be time consuming and error-prone).

- How many imports at a single time will rate limit my account and cause the problem? Today, I am always running laboratory sessions of fewer than 40 students. Will that cause a problem?

- Finally, any guess as to when you will merge the PR that seems to have a fix for this problem?

I'd like to know my options now so that if things don't work I don't set my classes too much further back to the point where we cannot cover the required amount of material during the laboratory sessions. So. thanks for any insights and for all of your work on GitHub Classroom. We've been using it in the past two years and would like to see all faculty in my department use it if we can iron out these issues.

[...]

Issue #1542, opened 2018-09-05
Github classroom importing repo fails


mikiehan:
I am using cbu-egr222-fa18 org account and created an assignment. 
However the import of "cbu-egr222-fa18/test-assignment" is failing when a student accepts the invitation. 

I have just one simple test.txt in the starter code. I also created test2-assignment but it also fails. Below are what my students sent me what they see from their view.

![image](https://user-images.githubusercontent.com/13935460/45119173-79269e80-b10f-11e8-843c-950a2600d0ce.png)

![image](https://user-images.githubusercontent.com/13935460/45119196-8b084180-b10f-11e8-90ba-ef6f6b1e9b72.png)

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

BenEmdon:
Hey @mikiehan 👋  thanks for writing in!
What happens when your students press the `Retry repository setup` button on GitHub classroom?

BenEmdon:
Would you mind linking me to the source repo? (the template repo that assignments are made out of)
Would you mind linking me to the assignment? (as if I were one of your students)

mikiehan:
https://classroom.github.com/a/lB7gk-eC
This is the invitation link

mikiehan:
The same thing happens when they press "Retry repository setup" button. I've just added you as the collaborator for the template repo, which has just one file test.txt

mikiehan:
This is another screenshot that my students set me
![image](https://user-images.githubusercontent.com/13935460/45120184-4500ad00-b112-11e8-9df3-114b775464bb.png)

BenEmdon:
The problem appears to be happening on GitHub.com, I have notified the responsible teams and we are investigating.

BenEmdon:
@mikiehan would you mind adding @dpmex4527 as a collaborator to your source repo?

mikiehan:
Just did

mikiehan:
I am sorry but what does #1544 mean?

mikiehan:
I am sorry but what does #1544 mean?

mikiehan:
Still my students are having issues doing import. The starter file contains a very simple test.txt. I hope to use it by next week for my assignment.

BenEmdon:
We have identified a possible work around. @mikiehan you will have to try the following steps:
* log out of GitHub.com
* go to https://import.github.com/login and login again

This should regenerate your import token and allow your students to continue importing.
Please let me know if this works.

[...]

Issue #1537, opened 2018-09-04
Students can't clone a repository from an acceptance link


sbailey6:
**Describe the bug**
I made an individual assignment, using starter code from a private repository I created. Last week, the acceptance link worked fine, but today, when students went to use the acceptance link in a later lab section, the following error occurred.

![image](https://user-images.githubusercontent.com/11618866/45055310-6f446300-b05d-11e8-9f24-f3f93dcf7994.png)

I had to keep retrying multiple times by refreshing the page, but eventually the error went away on its own. 

Any advice on how to avoid this in the future?

BenEmdon:
Hey @sbailey6 

We have identified the problem, it appears part of GitHub's API is down. The team responsible for the outage is currently responding. Sorry for the inconvenience.

I'll post back with updates.

BenEmdon:
The problem appears to be that that there was a influx of code imports, which caused the service to throttle GitHub. It will take a while before the queues of jobs die down.

We can't really say at this point. I'd estimate 1-2 hours for the situation to improve. I'll continue with updates here.

DbCrWk:
Having the same issues—thanks for responding so fast!

BenEmdon:
The problem appears was that starter code imports were being throttled by GitHub.com
The situation has improved slightly in the last 30 min. As we have reduced the import traffic.
I have a longer term solution in mind about how we will avoid throttling 🙂
I'll continue to  keep this thread posted with updates.

umasshokie:
Does this mean that we can tell students to try to create their repositories again? This caused us mass havok in our intro courses this afternoon, which of course isn't great to convincing students that GitHub is worth their time (but we'll still try). Don't want to tell them it works too soon though.

BenEmdon:
It looks like everything has calmed down for now. Please write in if the problem persists for you.

danwallach:
We ended up with half of the class in and half of the class out. The painful workaround is to create new repos manually with the right names, manually push to the new repo from your PC, and then manually give the student write-access to their repo. Takes about two minutes per student, and you've got about a 10% chance of some sort of typo making you have to rewind and do everything from scratch.

BenEmdon:
The issue has been resolved. If you experience it again, please reopen this issue.

Issue #1536, opened 2018-09-04
Source import (Starter Code Import) API is down


wz337:
Please choose either _issue_ or _feature request_

##
#### What you were trying to do ?
_I tried to create a new assignment ..._

#### What happened (include screenshot if you can) ?
_The assignment was not created ..._

#### Step-by-step reproduction instructions
Please provide as much explanations as you can, reproduce a bug is the first step to write a fix :raised_hands:

##
### Feature request :sparkles:
_Please provide a description of the feature you would like Classroom to have_

gchallen:
I'm seeing dozens of failed repository creations. Is this a general GitHub Classroom problem or something about our organization? (We have ~4500 private repositories already, but I did request many more than that.)

BenEmdon:
Hey @wz337 @gchallen 👋 

We have identified the problem, it appears part of GitHub's API is down. The team responsible for the outage is currently responding. Sorry for the inconvenience.

I'll post back with updates.

gchallen:
👍

gchallen:
@BenEmdon : any idea on an ETA for this, even in terms of minutes v. hours? I have the better part of 900 students banging against GC and I want to make a call about what to do...

BenEmdon:
Hey @gchallen,

The problem appears to be that that there was a influx of code imports, which caused the service to throttle GitHub. It will take a while before the queues of jobs die down.

We can't really say at this point. I'd estimate 1-2 hours for the situation to improve. I'll continue with updates here.

gchallen:
Fair enough. Should I shut down the invitation link in the mean time to avoid generating more traffic?

BenEmdon:
Up to you 🙂 it would definitely send your students the right message if you do so.

gchallen:
Hmm... OK. Well, right as I said that things seem to have improved.

gchallen:
OK, shut off. Can you ping us when this is fully resolved? I guess my students will be able to thank GitHub for a small extension 😉.

BenEmdon:
@gchallen we investigated and it looks like it was your token was the reason for the large influx of users 😅 The situation has since improved (not fully). 
We have an idea on how to quickly support classes the size of yours (we don't often have 900 students in a single classroom accept a assignment). 
I apologize for the inconvenience, I will investigate a more stable solution right away.

gchallen:
[...]

PR #1544, opened 2018-09-06
Repository import event job

BenEmdon:
## What
This PR proposes a new approach for reporting a `AssignmentRepo`'s starter code import as opposed to our current offering (`PorterStatusJob`).
Instead of polling and for a source imports status, instead we will kick off a job on the `repository_import` event.

The pros and cons of this approach were discussed in https://github.com/education/classroom/issues/1498

## Why is this now a priority?
This should reduce the load of requests that have caused throttling incidents (#1543 #1542 #1537 #1536)

## How will this affect UX 
We will no longer get intermediate steps form the import process and will instead look like:

sep-06-2018 16-32-09


BenEmdon:
This is ready for some 👀 
@d12 @srinjoym

app/jobs/repository_import_event_job.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    case status
+    when "success"
+      invite_status.completed!
+      broadcast_assignment_repo_success(user, invite_status, html_url)
+      GitHubClassroom.statsd.increment("v3_exercise_repo.import.success")
+    when "failure"
+      invite_status.errored_importing_starter_code!
+      broadcast_assignment_repo_failure(user, invite_status, html_url)
+      GitHubClassroom.statsd.increment("v3_exercise_repo.import.failure")
```
BenEmdon: In the case of failure here I am not destroying the Assignment repo (It's destroyed on manual retry). Should we destroy it here too?

app/controllers/application_controller/feature_flags_dependency.rb
```diff
@@ -37,4 +37,9 @@ def download_repositories_enabled?
     logged_in? && current_user.feature_enabled?(:download_repositories)
   end
   helper_method :download_repositories_enabled?
+
+  def repository_import_webhook_enabled?
+    logged_in? && current_user.feature_enabled?(:repository_import_webhook)
+  end
```
BenEmdon: I've added a feature flag so we can roll this out to some % of actors to try it out.

app/jobs/repository_import_event_job.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def perform(payload_body)
+    github_repo_id = payload_body.dig("repository", "id")
+    status = payload_body.dig("status")
+    html_url = payload_body.dig("repository", "html_url")
+
+    repo = AssignmentRepo.find_by(github_repo_id: github_repo_id)
+    return if repo.blank?
+    # Group assignments coming in a follow up PR
+    # repo ||= GroupAssignmentRepo.find_by(github_repo_id: github_repo_id)
```
BenEmdon: If we go with this approach I will close #1548 and make this job work for `GroupAssignmentRepo`s too.

app/jobs/repository_import_event_job.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    handle_assignment_repo(repo, status, html_url)
+  end
+
+  private
+
+  # rubocop:disable MethodLength
+  def handle_assignment_repo(assignment_repo, status, html_url)
+    user = assignment_repo.user
```
BenEmdon: I suppose we should be checking if the user has the feature flag enabled before going ahead with any actions. 
(Will add the feature flag check)

app/jobs/repository_import_event_job.rb
d12: It's probably ok :+1:

BenEmdon:
I'll wait until Monday to deploy this ⚓️

Issue #1498, opened 2018-08-23
Incorporate the new `repository_import` event webhook


BenEmdon:
### Feature request :sparkles:
A new webhook for the source importer API just released: [`#repositoryimportevent`](https://developer.github.com/v3/activity/events/types/#repositoryimportevent). 

## The difference between the webhook and `PorterStatusJob`
The webhook only fires after a source import completes, while [`PorterStatusJob`](https://github.com/education/classroom/blob/master/app/jobs/assignment_repo/porter_status_job.rb) gets all the intermediate steps between start and finish.

The **benefit** of intermediate data is we have a user experience that shows the progress of the import:

![aug-23-2018 11-17-50](https://user-images.githubusercontent.com/11095731/44534636-4370c700-a6c6-11e8-9955-42d4b7fc78d3.gif)
 
## How we can utilize the webhook
We could do choose to do one of the following:
* Replace the `PorterStatusJob` with a webhook job. 
  **CON:** Since the webhook only fires after an import completes we would lose all intermediate data which means we would have to ditch the progress bar ☝️ This might make the user feel like they are left hanging.
  **PRO:** This would mean significantly less background job infrastructure.
* Supplement the `PorterStatusJob`with a webhook job.
  **PRO:** This way we can double check that an import succeeds/fails. We could also check less frequently with `PorterStatusJob` because the second that the import finishes we would receive a webhook.
  **CON:** Another background job ¯\\_(ツ)_/¯

## Alternative approach 
Possibly we could add a `repository_import_status` event to the GitHub API that fires off a webhook whenever the status changes. This might be met with resistance since there might not be use cases outside classroom for it.

Would love to hear some thoughts on these approaches. This feature isn't urgent or necessary, but rather an enhancement.

CC: @d12 @srinjoym @johndbritton

johndbritton:
From the documentation:

> Triggered when a successful or unsuccessful repository import finishes for a GitHub organization or a personal repository. To receive this event for a personal repository, you must create an empty repository prior to the import.

This would mean that if the import succeeds or fails, a webhook will be fired and we can update the status immediately.

> Supplement the PorterStatusJobwith a webhook job.

I don't think you need a webhook job. You just need an endpoint to receive the incoming webhook. When you receive it you'd just need to update the status of the import.

I think the approach that makes sense here is to tune our polling to be less frequent and update to handle the webhook job immediately when delivered. At some point in the future we could drop the polling.

[...]

PR #1548, opened 2018-09-07
Group Porter Status Job

BenEmdon:
## What
This PR proposes a `PorterStatusJob` for group assignments. The purpose of the `PorterStatusJob` is to poll GitHub's source imports API for a starter code import's status to then broadcast the import's progress via ActionCable.

## Follow up PR work
* Add enqueue `PorterStatusJob` from `GroupAssignmentRepo::CreateGitHubRepositoryJob`

## What about the `repository_import` webhook event?
If we take the approach in #1544 then we should think about taking the same approach for `GroupAssignmentRepo`s

Closes https://github.com/education/classroom/issues/1521


BenEmdon:
Closing this in favor of the webhook mentioned in #1498

PR #1565, opened 2018-09-14
Homepage Ticker Visualization

srinjoym:
[WIP] Adds ticker to new landing page! Depends on #1564 

screen shot 2018-09-13 at 9 04 53 pm


BenEmdon:
@srinjoym is this a launch item?

srinjoym:
Decided to use ID of last Record instead of a full count, tabling this for now

PR #1567, opened 2018-09-14
Added 'cancelled' status handling to repository_import_event_job.rb 

BenEmdon:
## What ✨ 
This PR proposes we handle the upcoming `cancelled` status from the `repository_import` event webhook separately by doing the following:
* set the `invite_status` to `errored_importing_starter_code`
* broadcasting `"Youre starter code import was manually cancelled, please try again."`
* reporting a cancelled stat to datadog (`v3_exercise_repo.import.cancelled`)

---
This change is subject to change based on what happens upstream


app/jobs/repository_import_event_job.rb
```diff
@@ -6,6 +6,7 @@ class RepositoryImportEventJob < ApplicationJob
 
   CREATE_COMPLETE = "Your GitHub repository was created."
   IMPORT_FAILED = "We were not able to import starter code to your assignment, please try again."
+  IMPORT_CANCELLED = "Youre starter code import was manually cancelled, please try again."
```
ashleytqy: Should "Youre" be "Your"?

BenEmdon: Yes! Great catch @ashleytqy!

PR #1573, opened 2018-09-17
Require chewy

BenEmdon:
## What
This PR requires the `config/chewy.rb` which was previously not being required.

This PR should also make it so that indexing happens in the background.


config/application.rb
```diff
@@ -15,6 +15,9 @@
 require "failbot_rails"
 FailbotRails.setup("classroom#{'-staging' if Rails.env.staging?}")
 
+# configure chewy
+require_relative "chewy"
```
johndbritton: I'm not an expert on autoloading, would it be appropriate to move the file somewhere else so that it is autoloaded?

BenEmdon: Thats a good question. @d12 @tarebyte would either of you know?

tarebyte: @BenEmdon could we just move this to `config/initializers`?

app/jobs/application_job.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
+  around_perform do |_, block|
```
johndbritton: It's interesting that you needed to add this. Is this documented by Chewy as something you need to do?

BenEmdon: Ah, this was a workaround for when Chewy was version 0.10.0. Since #1582 merged I can remove this.

BenEmdon:
Could I get a final review on this 👀 
CC: @d12 @johndbritton @tarebyte

config/initializers/chewy.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# configure chewy
+require "./config/chewy"
```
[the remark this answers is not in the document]
BenEmdon: > This requires a file that is in effect one line Chewy.request_strategy = :active_job. Could we just put that one line here and remove the separate config file?

We could totally move the chewy config into this file 👍 

> ...so maybe you can remove the yml file as well and just configure everything in one place.

I prefer we keep the `.yml` because its easy to scan for important details like secrets.

PR #1574, opened 2018-09-17
Bring back Elasticsearch indexing

### LINKED ISSUE #1556: Gracefully handle Elasticsearch availability issues
We need to remove the hard dependency on Elasticsearch, if the service goes down Classroom should remain useable.

See this PR for details of the hard dependency: https://github.com/education/classroom/pull/1555

My recommended solution is to move the `update_index` calls into background jobs.

BenEmdon:
Since #1573 has merged we now perform all indexing in background jobs. This means that when ES goes down we will have graceful degradation (preventing the server from going down). When ES is down background jobs will fail and automatically be retried.

Reverts education/classroom#1555
Closes #1556


BenEmdon:
closing this in favor of #1588

Issue #1585, opened 2018-09-19
Chewy::UndefinedUpdateStrategy


BenEmdon:
Chewy::UndefinedUpdateStrategy

**[Chewy::UndefinedUpdateStrategy](https://haystack.githubapp.com/classroom/rollups/634135b9ba95dbeafe83bd076d637845)** &mdash; First occurred on 2018-09-19 10:05:27 -0700

```
  Index update strategy is undefined for current context.
  Please wrap your code with `Chewy.strategy(:strategy_name) block.`

```

*Brought to you by Haystack's `!` hotkey.  Try it on a needle or rollup today!*

PR #1588, opened 2018-09-21
Fix ES Headaches 

### LINKED ISSUE #1584: Elasticsearch::Transport::Transport::Errors::BadRequest
Elasticsearch::Transport::Transport::Errors::BadRequest

**[Elasticsearch::Transport::Transport::Errors::BadRequest](https://haystack.githubapp.com/classroom/rollups/b8d755b0dcb88a985d017fb14a8ba665)** &mdash; First occurred on 2018-09-19 10:07:30 -0700

```
[400] 
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "Rejecting mapping update to [stafftools] as the final mapping would have more than 1 type: [group_assignment_repo, assignment, organization, assignment_invitation, group_assignment_invitation, group_assignment, repo_access, deadline, grouping, assignment_repo, user, group]"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "Rejecting mapping update to [stafftools] as the final mapping would have more than 1 type: [group_assignment_repo, assignment, organization, assignment_invitation, group_assignment_invitation, group_assignment, repo_access, deadline, grouping, assignment_repo, user, group]"
  },
  "status": 400
}
```

*Brought to you by Haystack's `!` hotkey.  Try it on a needle or rollup today!*

BenEmdon:
This PR proposes a number of changes in response to our ES version change.

## Separating Indices
Firstly in response to #1584, this PR proposes we split apart the `StafftoolsIndex`'s types into their own indices. The problem is that [ElasticSearch 6 no longer supports multiple mapping types on a single index](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking-changes-6.0.html).  With this we also need to perform a multi index search query. 

The new indices are:
- AssignmentIndex
- AssignmentInvitationIndex
- AssignmentRepoIndex
- DeadlineIndex
- GroupAssignmentIndex
- GroupAssignmentInvitationIndex
- GroupAssignmentRepoIndex
- GroupIndex
- GroupingIndex
- OrganizationIndex
- RepoAccessIndex
- UserIndex

The multi index search can be found being used in:
[`app/controllers/stafftools/resources_controller.rb`](https://github.com/education/classroom/pull/1588/files#diff-7fa286c63638b15fdae36056d72f09f0)
And its source:
[`lib/multi_index_search_request.rb`](https://github.com/education/classroom/pull/1588/files#diff-d8b87212c0d97ca661e3c1d889be43e2)

## Background jobs kicking off background jobs
In response to #1585, this PR proposes we wrap our `ApplicationJob`s in a `Chewy.strategy(:active_job)` block. This is safe to do since Chewy workers don't inherit from `ApplicationJob` (they are `Chewy::Strategy::ActiveJob::Worker`).

This `ApplicationJob` change can be found in:
[`app/jobs/application_job.rb`](https://github.com/education/classroom/pull/1588/files#diff-ca33478ffbe0ef63784b3a32a155a83e)

Closes #1584 #1585


app/jobs/application_job.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
+  around_perform do |_, block|
```
johndbritton: I still think it would be better to set the default strategy for Chewy globally than to set it here.

app/jobs/last_active_job.rb
```diff
@@ -15,6 +15,6 @@ def perform(user_id, time_last_active)
 
     time_last_active = Time.zone.at(time_last_active)
     user.update_columns(last_active_at: time_last_active) # rubocop:disable Rails/SkipsModelValidations
-    User.update_index("stafftools#user") { user }
+    User.update_index("user#user") { user }
```
johndbritton: Can you explain why the argument here is `user#user`

app/controllers/stafftools/resources_controller.rb
```diff
@@ -2,6 +2,25 @@
 
 module Stafftools
   class ResourcesController < StafftoolsController
+    class StafftoolsRequest < Chewy::Search::Request
```
johndbritton: The naming for `StafftoolsRequest` is unclear to me. Is there a better name we could use?

app/jobs/last_active_job.rb
BenEmdon: Since we deleted the `StafftoolsIndex` we no longer update that index. The user type is now defined on the `UserIndex`. We do the same thing here: 
https://github.com/education/classroom/pull/1588/files#diff-4676c008b11a5480d73d4a6de01e45b9

app/jobs/application_job.rb
johndbritton: Add a comment linking to this discussion. It’s good to have this documented.

app/jobs/last_active_job.rb
johndbritton: So is this syntax a Chewy thing to
Maintain compatibility? If there’s only one type per index why would we need to specifu the index and the type?

app/controllers/stafftools/resources_controller.rb
johndbritton: `StafftoolsSearchRequest` Would be more clear or you could use `MultiIndexSearchRequest`.

app/jobs/last_active_job.rb
BenEmdon: > So is this syntax a Chewy thing to maintain compatibility?
It must be.

BenEmdon:
@johndbritton I addressed your comments. Mind if I get a re-review? 👀

BenEmdon:
@johndbritton I addressed your comments. Mind if I get a re-review? 👀

PR #1597, opened 2018-09-24
Remove d12 from CODEOWNERS

d12:
As discussed in Slack, @BenEmdon and @srinjoym are mostly reviewing eachother's PRs now, so I'm going to remove myself from the codeowners list. In cases where either of you would like my review, feel free to ping me! That way I know what you explicitly want my review on.


johndbritton:
Thanks @d12 for your service. Thanks for passing the torch 🔥.

PR #1610, opened 2018-09-25
Report handled octokit errors to datadog

BenEmdon:
## What

The problem I discovered was that our haystack (error reporting system) was being flooded with handled octokit errors (see internal image via slack). This noise makes it slightly harder to scan haystack for exceptions and impossible to scan haystack for unhandled octokit exceptions.

By switching the reporting from haystack to datadog we could easily improve our haystack visibility as well as improve our graphs about API exceptions (by moving them into datadog)


BenEmdon:
> @johndbritton Code looks correct, but can you write more about why you want to change this?

Updated description and posted internal supplementary image in slack!

Issue #1518, opened 2018-08-29
GraphQL - Implementation Checklist


d12:
This is a rough list of steps that need to happen as we work towards [GraphQL in Classroom](https://github.com/education/classroom/issues/1496). This work isn't a priority, it'll mostly be happening over my weekends and evenings.

## Steps

1. Store GitHub Global Relay IDs in Classroom
  - [x] Migrations to add appropriate `github_global_relay_id` columns
  - [x] When creating Users, Organizations, or Assignment Repos, record the Global Relay ID.
  - [x] Backfill Global Relay IDs for old Users, Organizations and Assignment Repos.

2. Local GraphQL MVP
  - [x] Add required gems
  - [x] Add autogenerated graphql-ruby boilerplate
  - [x] Define local query object with a test field
  - [x] Setup local GraphQL Client
  - [x] Add rake task to allow local GraphQL testing
  - [ ] Expose a /graphql route in development only to allow developers to use cURL/graphiql for debugging
  - [x] Add a single test for the local query object test field

3. Schema Stitching API
  - [x] Setup GitHub GraphQL Client
  - [x] Setup the remote loader from [graphql-remote_loader](https://github.com/d12/graphql-remote_loader)
  - [x] Add `#global_relay_id` method to our models (this is the ID representing our local models, not GitHub models.)
  - [x] Add some field an object in our API that fetches data from the GitHub API
  - [x] Test behavior of loaded field in GraphQL via VCR (`classroom_teacher` and `classroom_org` will need GitHub Global Relay IDs, may need to go in the .env, or we can make a REST request to record it.)

4. Building out our API
  - [ ] Model our data and relationships with a GraphQL schema.
  - [ ] Develop conventions around nullability, naming, etc.
  - [ ] Develop a consistent way to test authorization, fields, connections, etc.
  - [ ] Handle upstream errors in whatever way is most appropriate.

5. Enhancing our API
  - [ ] Assert we aren't hitting the REST API when resolving a GraphQL query
  - [ ] Collect robust timing metrics. Ideally with the specificity to identify slow resolvers.
  - [ ] Lint against GraphQL anti-patterns

6. Powering views with GraphQL
  - [ ] Create a second version of views that are backed w/ GraphQL instead of REST.
  - [ ] Put second version of these views behind a feature flag
  - [ ] More tests!
  - [ ] Ensure metrics comparing old vs new version of views are available
[...]

PR #1614, opened 2018-09-28
[GraphQL] Set node_id on new records

d12:

This PR adds `github_global_relay_id`'s to the db when creating Users, Orgs, or Assignments/GroupAssignments.

Thankfully none of them required new API calls except organization, and I'm not too worried about 1 extra API call when someone creates a classroom.


app/models/null_github_organization.rb
```diff
@@ -20,4 +20,8 @@ def login
   def name
     "Deleted organization"
   end
+
+  def node_id
```
BenEmdon: What are these `NullGitHubResource`s? Are these how we mock GitHub resources? 😮

PR #1618, opened 2018-10-01
Actually turn off autocomplete

BenEmdon:
## What
Does what #1601 was supposed to do


BenEmdon:
> screen shot 2018-10-01 at 1 55 36 pm

Mistakes were made!

PR #1619, opened 2018-10-01
Remove porter status job

BenEmdon:
## What
Previously we used `PorterStatusJob` to poll GitHub's API for the status of a _starter code import_. We have since changed our approach to wait for the `repository_import` webhook instead (#1498).

This PR just removes the old approach (and it's feature flag).


app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -38,10 +38,9 @@ def perform(assignment, user, retries: 0)
           message: IMPORT_STARTER_CODE,
           user: user,
           invite_status: invite_status,
-          percent: user.feature_enabled?(:repository_import_webhook) ? 50 : 0,
+          percent: 50,
```
johndbritton: Why is this hard coded as `50`?

spec/jobs/assignment_repo/create_github_repository_job_spec.rb
```diff
@@ -146,7 +123,7 @@
         .with(
           text: subject::IMPORT_STARTER_CODE,
           status: "importing_starter_code",
-          percent: 0,
+          percent: 50,
```
johndbritton: same question

app/jobs/assignment_repo/create_github_repository_job.rb
BenEmdon: We should actually get rid of the `percent` field entirely. These percentages are no longer meaningful.

BenEmdon:
# Update
Removed the `percent` field from all actionable broadcasts (as the field isn't being used anymore)

PR #1629, opened 2018-10-04
Elastic search rake task

BenEmdon:
Recently we switched to a new ElasticSearch provider, and no longer have historical data in our ElasticSearch.

This PR proposes an **ElasticSearch Reindex rake task**. 
Also changed in this PR:
* Centralized the map of all ES indexed models to their indices into `lib/es_indices.rb`
* added "ES" as an Rails acronym

## How it works
This task iterates through each ElasticSearch indexed model and makes executes an ActiveRecord `#touch`. Each `#touch` triggers an ElasticSearch reindex with the `:active_job` strategy.

## Output
```ruby
➜  classroom git:(elastic_search_rake_task) rake elasticsearch_reindex
=============================
------- Reindex Stats -------
=> number of models touched:
=> {"Assignment"=>1,
 "AssignmentInvitation"=>1,
 "AssignmentRepo"=>1,
 "GroupAssignment"=>1,
 "GroupAssignmentInvitation"=>1,
 "GroupAssignmentRepo"=>1,
 "Group"=>1,
 "Grouping"=>1,
 "Organization"=>1,
 "RepoAccess"=>1,
 "User"=>4}
=============================
```


lib/tasks/elasticsearch_reindex.rake
```diff
@@ -0,0 +1,25 @@
...
+# Executes an ActiveRecord `#touch` call on all ElasticSearch indexed models.
+# Each `#touch` triggers an ElasticSearch reindex with the `:active_job` strategy.
+task elasticsearch_reindex: :environment do
+  task_stats = {}
+  Chewy.strategy(:active_job) do
+    ESIndices.models.each do |model|
+      task_stats[model.to_s] = 0
+      model.find_in_batches(batch_size: 100) do |model_instances|
+        model_instances.each do |model_instance|
+          model_instance.touch # rubocop:disable Rails/SkipsModelValidations
```
johndbritton: Will this call to `touch` modify the `updated_at` time on our models? If so, I don't think this is the right approach.

lib/es_indices.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# This module contains all ElasticSearch indices and their assocated models.
+module ESIndices
```
johndbritton: This is cleaner, thanks for moving this to a separate file.

lib/tasks/elasticsearch_reindex.rake
BenEmdon: It would. 

Another approach would be to read from the db and manually invoke `update_index("model_name#model_name") { model_instance }`. I'll try that approach.

BenEmdon:
## Update 
The new approach is to invoke `update_index` on each model instead of using ActiveRecord `#touch`.  This new approach does not cause a write to the DB.

## Output
```ruby
➜  classroom git:(elastic_search_rake_task) rake elasticsearch_reindex
=====================================
----------- Reindex Stats -----------
Number of reindexing jobs enqueued:
=> {"Assignment"=>1,
 "AssignmentInvitation"=>1,
 "AssignmentRepo"=>1,
 "GroupAssignment"=>1,
 "GroupAssignmentInvitation"=>1,
 "GroupAssignmentRepo"=>1,
 "Group"=>1,
 "Grouping"=>1,
 "Organization"=>1,
 "RepoAccess"=>1,
 "User"=>4}
=====================================
```

lib/tasks/elasticsearch_reindex.rake
```diff
@@ -0,0 +1,25 @@
...
+# Executes an `update_index` call on all ElasticSearch indexed models.
+task elasticsearch_reindex: :environment do
+  task_stats = {}
+  Chewy.strategy(:active_job) do
+    ESIndices.models.each do |model|
+      model_name = model.to_s.underscore
+      task_stats[model.to_s] = 0
+      model.find_in_batches(batch_size: 100) do |model_instances|
+        model_instances.each do |model_instance|
+          model.update_index("#{model_name}##{model_name}") { model_instance }
```
johndbritton: Will this run inline or in a background job?

If it runs inline and there's an exception this will cause a break. If it's in the background, you'll just end up with some failed jobs when exceptions are raised.

BenEmdon: The indexing will happen in a background job. 
If indexing fails, it will be retried with exponential backoff (default in sidekick).

johndbritton:
Closed in favor of https://github.com/education/classroom/pull/1642

PR #1630, opened 2018-10-04
Alert if the GitHub API goes down

johndbritton:
This adds a simple check to application health that depends on the GitHub API.

In the event that the API is not responding or is responding with errors a Classroom alert will be triggered.


BenEmdon:
@johndbritton This looks good to me. Why did you un-deploy this?

johndbritton:
> @johndbritton This looks good to me. Why did you un-deploy this?

It was always failing due to checking for `status == green` instead of `status == good`

Issue #1627, opened 2018-10-03
Flaky test


johndbritton:
There's a flaky test related to an unexpected HTTP request for creating a team on GitHub.

Details: https://github.com/education/classroom/pull/1625#issuecomment-426479545

PR #1631, opened 2018-10-05
Unlock invite statuses cron job

BenEmdon:
Sometimes the repository setup process gets stuck from porter being slow, or the API going down. 

This PR proposes a new cron job that could run every X interval of time to check if the repository setup status has been stuck for over an hour. If the setup has been stuck for over an hour then it manually resets the setup to the `"unaccepted"` state.

Other notable changes in this PR:
* centralized the status enum from `InviteStatus` and `GroupInviteStatus`  into the `SetupStatus` concern
* created a `UnlockInviteStatusesService` to hold all the logic for for the cron job

## Output
```ruby
➜ $ rake unlock_invite_statuses
=====================================
---- Unlock InviteStatuses Stats ----
Number of invite statuses unlocked:
=> {"invite_status"=>
  {"waiting"=>1, "creating_repo"=>0, "importing_starter_code"=>0},
 "total_invite_statuses"=>1,
 "group_invite_status"=>
  {"waiting"=>0, "creating_repo"=>0, "importing_starter_code"=>0},
 "total_group_invite_statuses"=>0}
=====================================
```

## Questions:
* Should the threshold to reset an `invite_status` be 1 hour or less?
* How often should we run this cron job?

## Notes
* I have added the Heroku Scheduler (free plugin) 
  Docs: https://devcenter.heroku.com/articles/scheduler


lib/tasks/unlock_invite_statuses.rake
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# This task is run inside Heroku Scheduler at regular intervals.
+# See Heroku Scheduler docs at: https://devcenter.heroku.com/articles/scheduler
+
+# Unlocks `InviteStatus` and `GroupInviteStatus` records that have been locked (stuck) for more than an hour.
+# This task sets locked statuses to the "unaccepted" status.
+task unlock_invite_statuses: :environment do
```
johndbritton: I'd suggest moving this logic out to either the model or a service object so you can more easily write tests for it then call that code from your rake task.

app/models/group_invite_status.rb
```diff
@@ -3,6 +3,7 @@
 class GroupInviteStatus < ApplicationRecord
   ERRORED_STATUSES = %w[errored_creating_repo errored_importing_starter_code].freeze
   SETUP_STATUSES = %w[accepted waiting creating_repo importing_starter_code].freeze
+  LOCKED_STATUSES = %w[waiting creating_repo importing_starter_code].freeze
```
johndbritton: Do you need this on individual invites as well? Can it be shared across both?

lib/tasks/unlock_invite_statuses.rake
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# This task is run inside Heroku Scheduler at regular intervals.
+# See Heroku Scheduler docs at: https://devcenter.heroku.com/articles/scheduler
+
+# Unlocks `InviteStatus` and `GroupInviteStatus` records that have been locked (stuck) for more than an hour.
```
johndbritton: > more than an hour

change this to `TIME` and set time in one place.

lib/tasks/unlock_invite_statuses.rake
```diff
@@ -0,0 +1,41 @@
...
+    invite_status_model::LOCKED_STATUSES.each do |status, _|
+      task_stats[model_name][status] = 0
+    end
+
+    invite_status_model
+      .where(status: invite_status_model::LOCKED_STATUSES)
+      .find_in_batches(batch_size: 100) do |invite_statuses|
+        invite_statuses.each do |invite_status|
+          time_difference = Time.now.utc - invite_status.updated_at.utc
+          next unless time_difference > 1.hour
```
johndbritton: > 1.hour

also reference `TIME` here

lib/tasks/unlock_invite_statuses.rake
```diff
@@ -0,0 +1,41 @@
...
+          task_stats[model_name][invite_status.status] += 1
+          invite_status.unaccepted!
```
johndbritton: Suggest reporting an exception to haystack here. This should not happen, and any time that we do have to do this cleanup is a bad user experience and should be reported as an exception.

app/models/group_invite_status.rb
BenEmdon: They already exist on the individual `InviteStatus` model:
https://github.com/education/classroom/blob/master/app/models/invite_status.rb#L6

I will look into a clean way for the two models to share these constants.

lib/tasks/unlock_invite_statuses.rake
BenEmdon: I suppose this would make more sense as a "service", although I am not sure what that entails yet. I'll have to read up.

BenEmdon:
## Update
* Moved logic out of rake task and into a Service
* Centralized enum status logic 
* Made time configurable
* Report locked setups to Failbot (haystack)

This is ready for a re-review 👀

spec/models/group_invite_status_spec.rb
```diff
@@ -2,18 +2,21 @@
...
+  let(:group)        { Group.create(grouping: grouping, title: "#{Faker::Company.name} Team") }
   let(:invitation)   { create(:group_assignment_invitation) }
 
   describe "valid", :vcr do
     let(:invite_status) do
       subject.create(group: group, group_assignment_invitation: invitation)
     end
 
+    # TODO: make Group factory so we can make GroupInviteStatus factory to test SetupStatus behavior with:
+    # it_behaves_like 'setup_status'
```
BenEmdon: We have some test duplication here because we are unable to use the shared `it_behaves_like` tests since we are don't have a Group FactoryBot. If we can make a Group FactoryBot then we could solve #1627

PR #1634, opened 2018-10-06
Backfill GitHub global relay IDs

d12:
Do not merge before https://github.com/education/classroom/pull/1633

This is a backfill rake task for Users, Orgs, AssignmentRepos and GroupAssignmentRepos.

Unfortunately with current GitHub tooling and the number of records we're storing, backfilling via the REST API isn't feasible at all. Since we know exactly how global relay IDs are formed, we're cheating a little bit here and just calculating it from the type and database ID, both of which we have.

I tested this locally, all the produced gid's worked :+1:

```
➜  classroom git:(global_relay_id_backfill) ✗ rake global_relay_id_backfill
Backfilling Global Relay ID's to Users, Orgs, AssignmentRepos, and GroupAssignmentRepos.
Updating users...
Updating orgs...
Updating AssignmentRepos...
Updating GroupAssignmentRepos...
Done! We backfilled global relay IDs to:
1 users
1 organizations
1 assignment repos
1 group assignment repos
```


lib/tasks/global_relay_id_backfill.rake
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+# rubocop:disable Rails/SkipsModelValidations
+# rubocop:disable Metrics/BlockLength
+
```
BenEmdon: Nit: disable metrics the line before the block

PR #1635, opened 2018-10-06
First version of GraphQL in Classroom

d12:

This PR adds a simple GraphQL schema to GitHub Classroom!

Most of the PR is boilerplate required to have a GraphQL API. I've kept this PR to only the local GraphQL stuff, none of the schema stitching magic is present in this PR.

I included a single test showing the test field I added works, which acts as an integration test by exercising the full GraphQL stack.

I also added a rake task to make debugging GraphQL queries a little easier. In a future PR, I'll add a development route so we can cURL or point Graphiql to the API.

```
➜  classroom git:(first_version_of_graphql_in_classroom) ✗ rake query "query { test }"

Response:
{
  "test": "I am working!"
}

Parse time: 0.003005
Query time: 0.001484
```

@BenEmdon @srinjoym this is a pretty big one, I'm glad to set up a half hour next week to review this one as a group if you think it might be easier.


BenEmdon:
Looks like integrations are working!

Gemfile.lock
```diff
@@ -1,3 +1,12 @@
+GIT
+  remote: https://github.com/d12/graphql-client.git
```
BenEmdon: Why are we using your fork instead of `GitHub/graphql-client`?

d12: I touched on that in the Gemfile :)

https://github.com/education/classroom/pull/1635/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R31

d12: > Using fork of graphql-client which fixes a bug in graphql-client on Rails 5

PR #1637, opened 2018-10-09
Hide Download Repositories button when there are no assignment repos

### LINKED ISSUE #1589: Don't show download repo button unless there assignment repos
See https://github.com/education/classroom-assistant/issues/71#issuecomment-423628486 for details

rmacklin:
Fixes #1589

### DIFF
--- a/app/views/assignments/show.html.erb
+++ b/app/views/assignments/show.html.erb
@@ -27,7 +27,7 @@
         </div>
       </div>
 
-      <% if download_repositories_enabled? %>
+      <% if download_repositories_enabled? && @assignment_repos.present? %>
         <%= render partial: 'shared/open_on_assistant_modal', locals: {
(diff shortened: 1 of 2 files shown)


johndbritton:
Thinking about this a bit more and remembering https://github.com/education/classroom-assistant/issues/71, maybe we should set the button to disabled instead of removing it when there are no repositories to download.

BenEmdon:
I agree, I think it would be preferable to disable the button instead of removing it. Sorry for changing the constraints @rmacklin

rmacklin:
Sorry I wasn't able to get back to this. I was aiming to do it this weekend, but you beat me to it! Thanks for fixing it :)

BenEmdon:
Thanks for contributions, I made sure to merge them 😄

PR #1640, opened 2018-10-10
[WIP] Ensure organization webhook is active

johndbritton:
Make it possible to check for an active organization webhook, reactivate inactive organization webhooks, and create new organization webhooks where none exist.


app/models/organization.rb
```diff
@@ -56,6 +56,35 @@ def one_owner_remains?
...
+    return false if webhook_id.nil?
+
+    begin
+      github_client.org_hook(github_id, webhook_id).active
+    rescue Octokit::NotFound
+      false
+    end
+  end
+
+  def create_organization_webhook
```
johndbritton: @tarebyte this code is duplicated from the `Organization::Creator` class. Do you have any suggestions on how to move it into one place? It's locked up in the creator and not callable from anywhere else.

app/models/organization.rb
```diff
@@ -56,13 +56,39 @@ def one_owner_remains?
     users.count == 1
   end
 
+  def organization_webhook_active?
+    return false if webhook_id.nil?
+
+    begin
+      github_client.org_hook(github_id, webhook_id).active
+    rescue Octokit::NotFound
```
johndbritton: this could also raise `Octokit::Forbidden` but we should be checking that the token has enough scopes before making the call.

app/models/organization.rb
```diff
@@ -56,13 +56,39 @@ def one_owner_remains?
...
+
+  def ensure_organization_webhook
+    return if organization_webhook_active?
+
+    silently_remove_organization_webhook if webhook_id.present?
+    update!(webhook_id: nil, is_webhook_active: false)
+    create_organization_webhook
+  end
+
   def silently_remove_organization_webhook
```
johndbritton: This method should probably set `webhook_id` and `is_webhook_active` to `false, but doing that within this method breaks some tests.

johndbritton:
This works, but needs to be cleaned up and have a few more tests added. I did a branch deploy to install webhooks organizations that had `webhook_id` set to `nil`.

johndbritton:
I'm not going to take this PR any further. @BenEmdon feel free to crib from this as you work on webhooks resiliency.

PR #1642, opened 2018-10-11
Centralized ElasticSearch model and index lookup

johndbritton:
Kept some of the changes from https://github.com/education/classroom/pull/1629 but dropped the rake task.

### DIFF
--- a/app/controllers/stafftools/resources_controller.rb
+++ b/app/controllers/stafftools/resources_controller.rb
@@ -2,20 +2,6 @@
 
 module Stafftools
   class ResourcesController < StafftoolsController
-    ALL_INDICES = [
-      AssignmentIndex,
-      AssignmentInvitationIndex,
-      AssignmentRepoIndex,
-      GroupAssignmentIndex,
-      GroupAssignmentInvitationIndex,
-      GroupAssignmentRepoIndex,
-      GroupIndex,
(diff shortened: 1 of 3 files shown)


johndbritton:
@BenEmdon It turns out that the rake tasks included with Chewy seem to work fine. 

See: https://github.com/toptal/chewy#chewysync for more info.

My only concern is around rate limiting during the sync, but so far it doesn't seem to be a problem.

johndbritton:
We should also investigate this:

https://github.com/education/classroom/blob/e5f356aef05d1d581b94aa7e618c1673fa4b144d/app/jobs/last_active_job.rb#L18

I don't think it actually reindexes the user.

BenEmdon:
> It turns out that the rake tasks included with Chewy seem to work fine.

Were you able to run the chewy sync task without being rate limited?

BenEmdon:
> I don't think it actually reindexes the user.

I guess we should remove this then:

https://github.com/education/classroom/blob/e5f356aef05d1d581b94aa7e618c1673fa4b144d/app/jobs/last_active_job.rb#L18

johndbritton:
> Were you able to run the chewy sync task without being rate limited?

I was able to do successfully run `rake chewy:reset[organization]` which created a new index from scratch for that model.

PR #1644, opened 2018-10-11
[Idea] Sleep until rate limit updates and try again

johndbritton:
In working on updating ES indexes, I'm slightly worried about exhausting some rate limits.

This is just an idea for how we could rescue from a hitting our rate limits, hold off for a cool down period, and then resume work.

If something like this would work, I think it might be worth defaulting to not behaving this way and implementing a way to choose the behavior, similar to how we can choose to skip the API cache.

cc/ @d12 @tarebyte for your opinions on the approach.

### DIFF
--- a/app/models/github_model.rb
+++ b/app/models/github_model.rb
@@ -118,6 +118,14 @@ class GitHubModel
   # Returns a Sawyer::Resource or raises and error.
   def github_client_request(client, id, **options)
     GitHub::Errors.with_error_handling { client.send(github_type, id, options) }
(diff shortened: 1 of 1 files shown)


johndbritton:
Maybe we could move the sleep / retry logic into the `with_error_handling` definition and enable it by passing an argument?

tarebyte:
> Maybe we could move the sleep / retry logic into the `with_error_handling` definition and enable it by passing an argument?

Yeah that seems reasonable to me 👍

PR #1649, opened 2018-10-11
Alert when rate limit gets low

johndbritton:
If our app level rate limit is close to exhausted, trigger an alert.

This should almost never happen, so I feel ok with it being in the app health status.


BenEmdon:
@johndbritton I see you deployed this so I'll go ahead and merge for you ✅

PR #1650, opened 2018-10-12
[WIP] Check for invalid admin tokens

srinjoym:
WIP

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -35,13 +35,21 @@ class Organization < ApplicationRecord
       group_assignments.includes(:group_assignment_invitation)
   end
 
-  def github_client
-    if Rails.env.test?
-      token = users.first.token unless users.first.nil?
+  def github_client(random_token = !Rails.env.test?)
+    if random_token
+      org_tokens = users.order("RANDOM()").pluck(:token)
+      
+      org_tokens.each do |token|
+        client = Octokit::Client.new(access_token: token)
+        org = GitHubOrganization.new(client. github_id)
+        return client if org.admin?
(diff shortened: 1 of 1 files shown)


srinjoym:
@BenEmdon Could you take a look at this when you get a chance?

app/models/organization.rb
```diff
@@ -35,13 +36,16 @@ def all_assignments(with_invitations: false)
       group_assignments.includes(:group_assignment_invitation)
   end
 
-  def github_client
-    if Rails.env.test?
-      token = users.first.token unless users.first.nil?
+  def github_client(random_token: !Rails.env.test?)
```
BenEmdon: A similar change was proposed in https://github.com/education/classroom/pull/1309
@d12 @tarebyte didn't seem in favor of having a test conditional in a model. I'll tag @d12 as a reviewer to weigh in on this.

srinjoym: We started checking for the test environment variable in #1380 (
https://github.com/education/classroom/pull/1380/files/1a09eed308e2af679f24ef06bb6ebf307aaf8fff#r199937174) because of a bunch of test failures when the wrong token was used for auth. If we try to stub the method, were going to have make changes on a lot of files because the token validation is a before action on the base `ApplicationController` 

@d12 @BenEmdon  I can take a look into switching to stubs in a future PR if we want to move this out of the model?

d12: I think a better solution would be redoing the factory stuff to make it so that factory Classrooms aren't auto-populated with a user with a bad token. For now though, I'm :+1: with this.

PR #1657, opened 2018-10-16
Disable assignment download button

### LINKED ISSUE #1589: Don't show download repo button unless there assignment repos

BenEmdon:
## What
This PR proposes we disable the "Download Repositories" button when there are no `assignment_repos`.

## Screenshots
screen shot 2018-10-16 at 16 04 02

Closes #1637, Closes  #1589


app/views/shared/_open_on_assistant_modal.html.erb
```diff
@@ -1,12 +1,17 @@
-<iframe id="js-hidden-iframe" src="about:blank" style="display:none"></iframe>
+<% unless disabled %>
+  <iframe id="js-hidden-iframe" src="about:blank" style="display:none"></iframe>
+<% end %>
 <div class="mt-3 pr-2">
   <details class="dropdown details-reset details-overlay d-inline-block">
-    <summary id="js-download-repos-btn" class="btn" aria-haspopup="true">
+    <summary <%=  disabled ? "" : "id=\"js-download-repos-btn\"" %> class="btn <%= disabled ? "disabled" : "" %>" aria-haspopup="true">
```
srinjoym: The id of the `summary` has extra quotes around it with this (`"js-download-repos-btn"` instead of `js-download-repos-btn`)

Should be able to just drop the disabled check on the id since the user can't click if it's disabled anyways :

```suggestion
    <summary id="js-download-repos-btn" class="btn <%= disabled ? "disabled" : "" %>" aria-haspopup="true">
```

BenEmdon: The `id` of the summary does not have extra quotes around it:
```ruby
"id=\"js-download-repos-btn\""
``` 
☝️ is a string type return value. It would produce the string: `id="js-download-repos-btn"`

---
The reason for removing the `id` was because it causes a javascript `TypeError`:

oct-17-2018 12-32-51
This gif is what happens when you apply the suggested change you gave ☝️

srinjoym: Ah good catch! The quotes around the `id` are showing up as redundant on my inspector though and it's causing the modal to not reset after the app is launched when there are assignments:

What's happening right now:
image
What should happen:
image

Here's another fix, haven't seen the `undefined` error with this one:
```suggestion
    <summary id=<%=  disabled ? "" : "js-download-repos-btn" %> class="btn <%= disabled ? "disabled" : "" %>" aria-haspopup="true">
```

BenEmdon: Applied your suggestion ✅

app/views/shared/_open_on_assistant_modal.html.erb
```diff
@@ -1,12 +1,17 @@
-<iframe id="js-hidden-iframe" src="about:blank" style="display:none"></iframe>
+<% unless disabled %>
+  <iframe id="js-hidden-iframe" src="about:blank" style="display:none"></iframe>
+<% end %>
 <div class="mt-3 pr-2">
   <details class="dropdown details-reset details-overlay d-inline-block">
-    <summary id="js-download-repos-btn" class="btn" aria-haspopup="true">
+    <summary id=<%=  disabled ? "" : "js-download-repos-btn" %> class="btn <%= disabled ? "disabled" : "" %>" aria-haspopup="true">
```
rmacklin: Curious if you think there should be an aria-label and a tooltip when this is disabled to mention why it's disabled?

BenEmdon: That's a great idea, feel free to branch off my branch in #1665

PR #1667, opened 2018-10-18
Order paginated requests

### LINKED ISSUE #1659: Buggy listing of classroom repos
**Describe the bug**
In more than one classrooms (and assignments), I found that the listing of private repos of enrolled students is seemingly incorrect. Some repos are listed twice (although on different pages) while others are missing.

**To Reproduce**
1. Log in to https://classroom.github.com
2. Choose a classroom, and then an assignment
3. Investigate the listing of students' repos
4. Click page 2 on the bottom
5. Repos listed in page 2 are those that were previously listed on page 1, instead of those that should be listed in addition to page 1

**Expected behavior**
All the assignment repos should be listed (paged if there are more than should be displayed in a single page).

**Screenshots**
screen shot 2018-10-12 at 15 00 07
Page 1 of one of the assignments where the above mentioned incorrect listing happened.

screen shot 2018-10-12 at 15 00 15
Page 2 of the same assignment as above. Note that the listed repos are those included in page 1. There should have been other repos that are not listed in page 1.

**Additional context**
This seems to have happened to all the classrooms and assignments currently used in our university courses.

### LINKED ISSUE #1661: Repositories missing
**Describe the bug**
After students created their repositories and committed. Some students' repositories in the classroom are missing and some are repeating.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

BenEmdon:
This PR proposes a fix for #1661 and #1659 
There appears to have been weird behavior when querying without `order` on Assignments#show
https://github.com/education/classroom/blob/master/app/controllers/assignments_controller.rb#L33-L39

## How I determined the source of the bug
1. I discovered that our paginated assignment repo query was producing duplicates and missing records: https://github.com/education/classroom/issues/1659#issuecomment-431140486
1. I uncovered that there wasn't anything wrong with our pagination gem since I was able to reproduce the bug using `limit` and `offset`: https://github.com/education/classroom/issues/1659#issuecomment-431140486
1. Then I finally discovered that adding `order(:id)` would prevent duplicates and missing records in pagination: https://github.com/education/classroom/issues/1659#issuecomment-431157685

## Why would query yield a response like this? (updated)
Without `order` in the presence of a `limit`, the result will be an unpredictable subset of the query rows.

> When using LIMIT, it is important to use an ORDER BY clause that constrains the result rows into a unique order. Otherwise you will get an unpredictable subset of the query's rows. You might be asking for the tenth through twentieth rows, but tenth through twentieth in what ordering? The ordering is unknown, unless you specified ORDER BY.

From Postgres documentation: https://www.postgresql.org/docs/9.3/static/queries-limit.html

## No tests?
Unfortunately I was unable to confirm the problem by creating a local failing test. I did however test the queries on production in https://github.com/education/classroom/issues/1659#issuecomment-431157685
I hope that sufficient evidence of the fix ☝️ 

Closes #1661 Closes #1659

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -27,18 +27,27 @@ class AssignmentsController < ApplicationController
     end
   end
 
+  # rubocop:disable MethodLength
   # rubocop:disable Metrics/AbcSize
   def show
     if @organization.roster
-      @roster_entries = @organization.roster.roster_entries.page(params[:students_page]).order_for_view(@assignment)
+      @roster_entries = @organization.roster.roster_entries
+        .order(:id)
+        .page(params[:students_page])
+        .order_for_view(@assignment)
 
       @unlinked_user_repos = AssignmentRepo
+        .order(:id)
         .where(assignment: @assignment, user: @unlinked_users)
         .page(params[:unlinked_accounts_page])
     else
-      @assignment_repos = AssignmentRepo.where(assignment: @assignment).page(params[:page])
+      @assignment_repos = AssignmentRepo
+        .where(assignment: @assignment)
+        .order(:id)
+        .page(params[:page])
     end
   end
+  # rubocop:enable MethodLength
   # rubocop:enable Metrics/AbcSize
 
   def edit; end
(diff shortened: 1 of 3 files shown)


d12:
@BenEmdon Do you know what PR caused this? It seems like the issues started happening very abruptly, one of the recent PR's must have changed something.

BenEmdon:
@d12 unfortunately I have no idea what might have caused this. This particular code hasn’t been touched in a very long time. The fix is related to the bahvior of Postgres and I’m pretty sure that hasn’t changed :confused:

I can confirm that this resolves #1661 and #1659 ✅

PR #1673, opened 2018-10-22
Fix CGI Escape on API Tokens

### LINKED ISSUE #1632: Classroom Assistant - Could not find assignment
**Describe the bug**
Classroom assitant couldn't find my assigment and I don't know why.

**Screenshots**
image

srinjoym:
Fixes #1632 

After lots of local testing, it looks like we're hitting an issue when we URL encode the API tokens using `CGI.escape` the user token (to replace special characters when passing it over the API). 

For some users (has to be the right user id range) the token cannot be decoded properly after it comes back for authorization. The solution is to `CGI.unescape` the token before parsing it so we retain the special characters in the token.


app/controllers/assignments_controller.rb
```diff
@@ -77,7 +77,7 @@ def destroy
   end
 
   def assistant
-    code_param = CGI.escape(current_user.api_token)
+    code_param = current_user.api_token
```
srinjoym: Moved URL encoding to base `MessageVerifier` class so that we don't have to do this everytime we generate a token

PR #1680, opened 2018-10-26
Fix flakey tests + clean up Group model

### LINKED ISSUE #1627: Flaky test

BenEmdon:
This PR proposes we remove network requests from the `Group` model validations, and into a `Group::Creator`. This will allow us to create `Group` instances without creating also requesting GitHub to create a team.

## Breakdown of all the changes
* Removed the `on: create` `Group` validation that requests to create a GitHub team
* Made `github_teamable.rb` safer (with error handling)
  * Wrote tests for `github_teamable.rb` since it didn't have tests
* Wrapped `delete_team` in `GitHub::Errors.with_error_handling do ...`
* Added a `Group::Creator` class to manage creation of `Group` instances
  * The creator ensures that a GitHub team is created on `Group` creation
  * Replaced all usage of `Group.create` in `/app` with `Group::Creator.perform(..)`
* Added a `Group` and `GroupInviteStatus` factorybot
* Fixed tests that created `Group` by requesting GitHub to create a group (fixes #1627)


spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -256,23 +247,13 @@
...
 
-      it "redeems the users invitation" do
-        patch :accept_invitation, params: { id: invitation.key, group: { title: "Code Squad" } }
-
-        expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/teams"))
-        expect(WebMock).to have_requested(:post, github_url("/organizations/#{organization.github_id}/repos"))
-
-        expect(group_assignment.group_assignment_repos.count).to eql(1)
-        expect(student.repo_accesses.count).to eql(1)
-      end
```
BenEmdon: Removed this test since it's actually testing some dead code (I need to get around to removing a bunch of dead code).

spec/controllers/group_assignment_invitations_controller_spec.rb
```diff
@@ -650,10 +620,6 @@
       it "deletes the old group assignment repo" do
         expect { @group_assignment_repo.reload }.to raise_error(ActiveRecord::RecordNotFound)
       end
-
-      it "creates a new group assignment repo for the group" do
-        expect(GroupAssignmentRepo.last.id).not_to eq(@group_assignment_repo.id)
-      end
```
BenEmdon: Also dead code ☝️

spec/models/group_invite_status_spec.rb
```diff
@@ -1,59 +1,19 @@
...
-        invite_status.creating_repo!
-        expect(invite_status.setting_up?).to be_truthy
-      end
-
-      it "is setting_up? when errored_importing_starter_code?" do
-        invite_status.importing_starter_code!
-        expect(invite_status.setting_up?).to be_truthy
-      end
-    end
+    it_behaves_like "setup_status"
```
BenEmdon: All the tests removed in this file are already in `setup_status_spec.rb`. See `it_behaves_like "setup_status"`

spec/models/concerns/github_teamable_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+class MockGitHubTeam
+  attr_reader :id
+
+  def initialize(id)
+    @id = id
+  end
+end
+
+shared_examples_for "github_teamable" do
```
BenEmdon: Added these tests because this `ActiveSupport::Concern` didn't have any test coverage.

app/models/group.rb
```diff
@@ -27,10 +27,6 @@ class Group < ApplicationRecord
 
   validates :slug, uniqueness: { scope: :grouping }
 
-  before_validation(on: :create) do
-    create_github_team if organization
-  end
```
BenEmdon: This right here is what was causing flakey tests (and preventing us from creating a factorybot) I have since moved this behavior into the `Group::Creator`.

app/models/group_assignment_invitation.rb
```diff
@@ -29,9 +29,11 @@ class GroupAssignmentInvitation < ApplicationRecord
   def redeem_for(invitee, selected_group = nil, new_group_title = nil)
     return Result.failed("Invitations for this assignment have been disabled.") unless enabled?
 
-    repo_access    = RepoAccess.find_or_create_by!(user: invitee, organization: organization)
-    invitees_group = group(repo_access, selected_group, new_group_title)
+    repo_access = RepoAccess.find_or_create_by!(user: invitee, organization: organization)
+    group_creator_result = group(repo_access, selected_group, new_group_title)
+    return Result.failed(group_creator_result.error) if group_creator_result.failed?
```
srinjoym: What happens in this case, is the error caught further up the stack? Just checking because we didn't return a `Result.failed` in any case before

spec/controllers/orgs/group_assignment_repos_controller_spec.rb
```diff
@@ -6,26 +6,19 @@
...
     create(:group_assignment, title: "Learn Ruby", organization: organization, public_repo: false)
   end
-
-  let(:group) { Group.create(title: Time.zone.now, grouping: group_assignment.grouping) }
-
-  let(:group_assignment_repo) do
-    GroupAssignmentRepo.create(group_assignment: group_assignment, group: group)
-  end
+  let(:group) { create(:group, grouping: group_assignment.grouping, github_team_id: 2_976_561) }
+  let(:group_assignment_repo) { GroupAssignmentRepo.create(group_assignment: group_assignment, group: group) }
```
srinjoym: Will using an actual `GroupAssignmentRepo` instead of a factory cause any flaky-ness? That might be the next thing to tackle if it also causes problems

spec/models/concerns/github_teamable_spec.rb
srinjoym: 🙌

spec/controllers/orgs/group_assignment_repos_controller_spec.rb
BenEmdon: A factory would be better. Unfortunately the `GroupAssignmentRepo` model suffers from the same problem that the `Group` model suffers from: networks requests in the validations.

I would love to clean that up and move `GroupAssignmentRepo` to a `Creator` pattern and clean up its job!

app/models/group_assignment_invitation.rb
BenEmdon: Previously if it failed to create a group it would throw a `GitHub::Error` here (and it wouldn't be handled 😬).
With this change we handle handle group creation errors in the `Group::Creator`:
https://github.com/education/classroom/blob/383df517636d446c8a25c6f6e8cebc61ccbbffd4/app/models/group/creator.rb#L60-L61

PR #1683, opened 2018-10-29
Paginate groups on group_assignment_invitations_controller#show

### LINKED ISSUE #1682: 500 Internal server errors in very large group assignments
**Describe the bug**
When students attempt to visit the assignment invitation link for a group assignment, the application often gives a 500 Internal server error. Occasionally, after refreshing several times, the page will eventually load which suggests that the application is sometimes timing out on the request.

**To Reproduce**
Visit <https://classroom.github.com/g/WrZUiXxO>. It often (but not always) results in a 500 Internal server error.

**Expected behavior**
The signup listing of existing groups for the assignment with options to join an existing team or create a new team.

**Additional context**
We are using GitHub Classroom to serve a very large classroom of about 900 students working in teams of up to 2 students. There may be on the order of 500 team repositories. I suspect that increasing the server timeout or improving the performance of the request will resolve the problem.

BenEmdon:
## What
This PR proposes we paginate groups on `GroupAssignmentInvitationsController#show`.

Here is what it looks like:
oct-29-2018 15-25-44

_Ignore the empty groups, they are just there to visualize_

Fixes #1682


BenEmdon:
Our deployment pipeline cant read the branch name because of the `#` character so I'll make the branch `paginate-groups-on-group_assignment_invitations` and deploy that.

BenEmdon:
Can confirm this fixes the slow page load 👍 
I'll post the link in slack 🔗

PR #1686, opened 2018-10-31
Validate that no emojis are in a Group's Title

### LINKED ISSUE #1465: Team with emoji-only name
**Describe the bug**
When using a group assignment in GH Classroom, one team decided that they would enter the team name "🐈". This was accepted by the team creation process, but generated a team with no URL so no way to access the team's discussion page etc. I did manage to somehow find another way to get to the team page and change the name to real characters.

**To Reproduce**
1. Make a group assignment in GH Classroom
2. Distribute the link to students
3. A student follows the instructions to create a new team
4. The student enters just an emoji as the name of the new team
5. The team is created with the name 🐈
6. Go to the Org page for the classroom and click the teams tab
7. The teams tab shows all the teams, including 🐈, although that team is only named in the description and has no link
8. Try to visit the team's page somehow

**Expected behavior**
A team name can only contain valid characters. Invalid names are rejected and no team is created.

**Screenshots**
screen shot 2018-08-08 at 3 17 52 pm
This is after I added the name "cat". Before that, the secret tag was the only thing on that row.

BenEmdon:
This PR proposes we add a `no_emoji` validator so that we can prevent emoji input in for certain attributes.

The syntax of this looks like so:
```ruby
class SomeModel < ApplicationRecord

  validates :some_attribute, no_emoji: true

end

some_object = SomeModel.new(some_attribute: "just a string")
some_object.valid? # => true
some_object.save   # => true

# When we make the 
some_object.some_attribute = "🎃"
some_object.valid? # => false
some_object.save   # => false
some_object.save!  # => some attribute cannot contain emojis (ActiveRecord::RecordInvalid)
```

We use the `no_emoji` validator to prevent the `Group` from having emojis.

Fixes #1465


BenEmdon:
## Update
Only validate on create in 1ccbea7

BenEmdon:
## What it looks like

oct-31-2018 16-42-23

Issue #1647, opened 2018-10-11
Organizations Are Missing Webhooks


BenEmdon:
From running the following query on production, it is evident that there are still a significant amount of `Organizations` that are missing webhooks:
```ruby
pry(main)> Organization.where(webhook_id: nil).count
DEBUG -- :    (7.0ms)  SELECT COUNT(*) FROM "organizations" WHERE "organizations"."deleted_at" IS NULL AND "organizations"."webhook_id" IS NULL
=> 3838
```
This number excludes `Organizations` that have a `webhook_id` but who's webhooks are _deleted_ or _inactive_.

## The source of this problem
The problem occurred when we introduced webhooks back in https://github.com/education/classroom/pull/797, we forgot to backfill organizations created before the change with web hooks.

## Why this is important

All `Organizations` require webhooks to perform starter code imports as we rely on the `repository_import` event.

## Steps to ensure all Organizations have `webhook_id`s and those webhooks are active
- [ ] method around ensuring a webhook is present and valid 
  - make sure we don't hit API cache
- [ ] model `OrganizationWebbhook` as a `GitHubResource`
- [ ] change `is_webhook_active` column on `Organization` to reflect the last webhook delivered as a date (or move this to a delegated model)

## Steps to ensure an org is in good health ✅
- [ ] define a method to check if an `Organization` meets all minimum requirements
  - if webhooks are present and enabled
  - if a user in the org has all the scopes needed

srinjoym:
@BenEmdon What do you think about renaming this issue to include "Importing starter code taking a long time" or something like that so people know that it's related to this problem? That could help redirect people here as they run into issues

BenEmdon:
I'd like to keep the development track and incident reports separate so we can filter out the noise 😕 Maybe I should start a master issue for all the incidents?

srinjoym:
That totally makes sense! I think we can open up another pinned issue just for the incidents and include some instructions on what to do if people hit this issue. Hopefully that makes it a bit easier to sort through

BenEmdon:
Pinned #1778

BenEmdon:
> ![Screen Shot 2019-05-28 at 5 51 30 PM](https://user-images.githubusercontent.com/11095731/58514916-3f9c3480-8171-11e9-8acc-7798a97ada38.png)

Interesting, hooks created by the OAuth app cannot be seen by users. And hooks created by users cannot be seen by the Oauth app.

BenEmdon:
##Update 
[...]

PR #1687, opened 2018-11-01
Organization webhook migration

BenEmdon:
This PR proposes we add a new model named `OrganizationWebhook` which has 2 purposes:
1. keep a record of when the last webhook was received using the `last_webhook_recieved` column 
1. persist and manage an organization's webhook via its `github_webhook_id` column

## Why not add extra columns to the Organization model?
It might be too write heavy on the Organization model if we commit to writing to it every time we get a webhook event.

## What will this look like for multiple classrooms?
org webhook relationship

Part of #1647
Models for this coming in a follow up PR

### DIFF
--- a/db/migrate/20181101185901_create_organization_webhooks.rb
+++ b/db/migrate/20181101185901_create_organization_webhooks.rb
@@ -1,0 +1,12 @@
+class CreateOrganizationWebhooks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :organization_webhooks do |t|
+      t.integer :github_webhook_id
+      t.integer :github_organization_id, null: false
+      t.datetime :last_webhook_recieved
+
+      t.timestamps
+    end
+    add_index :organization_webhooks, :github_webhook_id, unique: true
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20181101185901_create_organization_webhooks.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOrganizationWebhooks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :organization_webhooks do |t|
+      t.integer :github_webhook_id
+      t.datetime :last_webhook_recieved
+
```
BenEmdon: Should this table also keep a record of the `github_organization_id`? Or are we planning to build a model for that?

db/migrate/20181101185901_create_organization_webhooks.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOrganizationWebhooks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :organization_webhooks do |t|
+      t.integer :github_webhook_id
```
BenEmdon: Should this column name be `github_webhook_id` or just `github_id`?

BenEmdon:
This could be better designed if we had an Organization model that lives separately from the multi org Classroom model. Making this WIP for now as I take a jab at #1688

PR #1248, opened 2018-01-31
Allow a GitHub Organization to be used by multiple Classrooms

tarebyte:
Closes https://github.com/education/classroom/issues/595

After I opened https://github.com/education/classroom/pull/1247 I wondered what it would look like if we just removed all of the safe guards and just let an org be used for multiple classrooms.

Turns out it's not nearly as bad as I thought 😄 

## ToDO
- [x] Ship #1247
- [x] Add tests
- [x] Do some pen testing to make sure I didn't break existing features

## Follow Up
- [ ] Migration `Organization` model to `Classroom` model (that's gonna be a pain in the butt)

PR #1688, opened 2018-11-01
Rename Organization to Classroom: Part 1 The Great Migration

BenEmdon:
This PR proposes we rename the `Organizations` table to `Classrooms` since we added multi classroom functionality in https://github.com/education/classroom/pull/1248

This PR is not safe to deploy on its own and will require the `Organization` model in `app/models/organization.rb` to be renamed to `Classroom` (coming in part 2)

This migration must run in total isolation. To run this in total isolation we will have to turn on Heroku's maintenance mode.

Tests will fail on this PR as it doesn't have the model rename mentioned above ☝️


BenEmdon:
If you want to follow along with Part 2 Organization => Classroom (rough title) you can do so at:
https://github.com/education/classroom/compare/rename_organization_to_classrooms

BenEmdon:
Closing this for now while I work on other priorities

PR #1693, opened 2018-11-08
Shoulda Matchers Gem

BenEmdon:
This PR proposes we add the shoulda matchers gem https://github.com/thoughtbot/shoulda-matchers
This gem is useful for testing `ActiveRecord` relationships and validations. Here's an example of some shoulda-matchers specs:

https://github.com/education/classroom/blob/fb0aad38ccde06e82e0d3947cb3d72412dca697d/spec/models/invite_status_spec.rb#L10-L14

I am adding this as I remodel webhooks as part of https://github.com/education/classroom/issues/1647


spec/models/invite_status_spec.rb
```diff
@@ -7,38 +7,16 @@
...
-  end
-
-  it_behaves_like "setup_status"
+  it { should belong_to(:assignment_invitation) }
+  it { should belong_to(:user) }
 
-  describe "valid" do
-    let(:invite_status) { create(:invite_status, user_id: user.id, assignment_invitation_id: invitation.id) }
+  it { should validate_presence_of(:assignment_invitation) }
+  it { should validate_presence_of(:user) }
```
BenEmdon: Rewrote these tests to test that it was working as intended

PR #1692, opened 2018-11-08
Organization webhook migration

BenEmdon:
This PR proposes we add a new model named `OrganizationWebhook` which has 2 purposes:
1. keep a record of when the last webhook was received using the `last_webhook_recieved` column 
1. persist and manage an organization's webhook via its `github_webhook_id` column

## Why not add extra columns to the Organization model?
It might be too write heavy on the Organization model if we commit to writing to it every time we get a webhook event.

## What will this look like for multiple classrooms?
org webhook relationship

Part of #1647
Models for this coming in a follow up PR

PR #1694, opened 2018-11-12
Organization webhook

BenEmdon:
Migration for this is separate in #1692 

This PR proposes we add a new model named `OrganizationWebhook` which has 2 purposes:
1. keep a record of when the last webhook was received using the `last_webhook_recieved` column 
1. persist and manage an organization's webhook via its `github_webhook_id` column

## Changes summary
* Added `app/models/organization_webhook.rb`
* `Organization belongs_to :organization_webhook`
* create organization webhook model if one doesn't exist in `app/models/organization/creator.rb`

## What will this look like for multiple classrooms?
org webhook relationship

Part of #1647
Backfill migration coming in a follow up PR


app/models/organization/creator.rb
```diff
@@ -66,12 +66,18 @@ def perform
       begin
         github_organization = GitHubOrganization.new(users.first.github_client, github_id)
 
+        organization_webhook = OrganizationWebhook.find_or_initialize_by(github_organization_id: github_id)
+        webhook_id = create_organization_webhook!
```
srinjoym: I think it makes sense to move the logic for creating a webhook from the `Organization::Creator` to this new `OrganizationWebhook` model so that we don't have the code in multiple places.

@BenEmdon What do you think? I can move these methods over in a future PR if you'd rather focus on what you're working on

BenEmdon: Totally agree 👍 That work aligns with the effort I am making in #1647 and I'll get to that in a follow up PR 😄

BenEmdon:
Merged https://github.com/education/classroom/pull/1692
Just waiting on the 👍 👎  on this cc @srinjoym

PR #1695, opened 2018-11-12
Organization webhook model backfill

BenEmdon:
## Before reviewing:
- The model migration was introduced in #1692 
- The model was introduced in #1694 

☝️ Review these first

## This PR
This PR proposes a task to backfill `Organizations` to belong to an `OrganizationWebhook`.
The task can be found in `lib/tasks/organization_webhook_backfill.rake`

## Diagram of the relation
org webhook relationship

Part of #1647

PR #1696, opened 2018-11-13
 Updates last_webhook_recieved for every webhook event

BenEmdon:
_Please review #1692 #1694 #1695 first_

## This PR
This PR proposes that we update the `last_webhook_recieved` column on `OrganizationWebhook` for every webhook event.

This change relies on other PR's being merged first ☝️ 

Part of https://github.com/education/classroom/issues/1647


app/controllers/hooks_controller.rb
```diff
@@ -21,6 +21,7 @@ def send_github_payload_to_job(payload_body)
     github_event = request.env["HTTP_X_GITHUB_EVENT"]
     return unless GitHub::WebHook::ACCEPTED_EVENTS.include?(github_event)
     "#{github_event}_event_job".classify.constantize.perform_later(payload_body)
+    update_last_webhook_recieved(payload_body)
```
BenEmdon: Maybe we should invoke this every time we revive a webhook from GitHub instead of if the webhook event type is one of the `GitHub::WebHook::ACCEPTED_EVENTS`?

app/controllers/hooks_controller.rb
```diff
@@ -35,4 +36,13 @@ def verify_payload
     end
     # rubocop:enable GuardClause
   end
+
+  def update_last_webhook_recieved(payload_body)
+    github_organization_id = payload_body.dig("organization", "id")
+    return false unless github_organization_id
+    OrganizationWebhook
+      .find_by!(github_organization_id: github_organization_id)
```
BenEmdon: Chose to use the unsafe `find_by!` here because it is _exceptional_ if we receive a webhook for an organization that isn't on record. 👍 👎 ❓

srinjoym: Sounds good to me 👍

BenEmdon:
Following this we should also make the `last_webhook_recieved` attribute visible from stafftools

PR #1707, opened 2018-11-26
Finds a User's token that has the `admin:org_hook` scope

BenEmdon:
This PR proposes methods for `OrganizationWebhook` to find User tokens with the `admin:org_hook` scope. These tokens are required to create webhooks for organizations.

In a follow up PR I will make the methods for creating the webhooks using these tokens.

Part of #1647 
> - [ ] move the logic for creating a webhook from the `Organization::Creator` to `GitHubOrganizationWebhook`

### DIFF
--- a/app/models/organization_webhook.rb
+++ b/app/models/organization_webhook.rb
@@ -27,7 +27,7 @@ class OrganizationWebhook < ApplicationRecord
     token = if Rails.env.test?
               users.first&.token
             else
-              user_with_admin_org_hook_scope.sample&.token
+              users_with_admin_org_hook_scope.sample&.token
             end
     raise NoValidTokenError, "No valid token with the `admin:org` hook scope." if token.nil?
     Octokit::Client.new(access_token: token)
--- a/app/models/organization_webhook.rb
+++ b/app/models/organization_webhook.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class OrganizationWebhook < ApplicationRecord
+  class NoValidTokenError < StandardError; end
+
   has_many :organizations
   has_many :users, through: :organizations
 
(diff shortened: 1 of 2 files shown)


app/models/organization_webhook.rb
```diff
@@ -9,7 +11,62 @@ class OrganizationWebhook < ApplicationRecord
...
+  #
+  # Returns a User's token or raises a NoValidTokenError if one could not be found.
+  #
+  # Warning: This could potentially take very long for organizations
+  # of a large size, so invoke cautiously.
+  def admin_org_hook_scoped_github_client
+    token = if Rails.env.test?
+              users.first&.token
+            else
+              user_with_admin_org_hook_scope.sample&.token
```
srinjoym: Should this be calling the method underneath? There's a small typo
```suggestion
              users_with_admin_org_hook_scope.sample&.token
```

spec/models/organization_webhook_spec.rb
```diff
@@ -18,18 +18,48 @@
...
+      end
     end
+  end
 
-    context "has no organizations" do
-      it "returns nil" do
-        expect(subject)
-          .to receive(:organizations)
+  describe "#users_with_admin_org_hook_scope" do
+    context "user with admin_org hook scope exists doesn't exist" do
```
srinjoym: Test name typo? :
```suggestion
    context "user with admin_org hook scope doesn't exist" do
```

spec/models/organization_webhook_spec.rb
```diff
@@ -18,18 +18,48 @@
...
-    it "requests a GitHubOrganization" do
-      expect_any_instance_of(Organization).to receive(:github_organization)
-      subject.github_organization
+  describe "#admin_org_hook_scoped_github_client" do
+    context "token is present" do
+      before do
+        allow(subject).to receive_message_chain(:users, :first, :token) { "token" }
+      end
+
+      it "raises a NoValidTokenError" do
```
srinjoym: We don't expect to raise an error here, just need to update test name?

BenEmdon: Good catch 👍 a case of some copy pasta 🍝

BenEmdon:
```suggestion
      it "returns a Octokit::Client" do
```

PR #1708, opened 2018-11-29
Admin org hook scope reduced search

BenEmdon:
This PR proposes we add an optional `organization` argument to `admin_org_hook_scoped_github_client` to shrink it's search scope to a single organization. 

Without the reduced search scope, `admin_org_hook_scoped_github_client` will search through all the organizations associated with the same GitHub org ID for users with the `admin:org_hook` scope. This could be very expensive for organizations with multiple classrooms.

Part of #1647 
> - [ ] move the logic for creating a webhook from the `Organization::Creator` to `GitHubOrganizationWebhook`


BenEmdon:
I have a different approach in mind. Closing this for now (and probably forever).

PR #1715, opened 2018-12-10
Added retrieve_org_hook_id! + Fixed create_org_hook!

BenEmdon:
This PR proposes we add the method `#retrieve_org_hook_id!` to `OrganizationWebhook`, and fixes the behaviour of `#create_org_hook!`

## Added 
`#retrieve_org_hook_id!` is an internal method that retrieves the classroom webhook id on GitHub if it exists and saves it. 

This method is useful for when a classroom record doesn't have a webhook ID but one exists on GitHub. We are able to get webhook from GitHub using the assumption that classroom only creates 1 webhook.

This method is added to the flow of ensuring a organization has a working webhook: `ensure_webhook_is_active`.

## Fixed
`create_org_hook!` now saves the returned org hook id, whereas before it would not save the id to the record.

--
Part of #1647 
> - [ ]  method around ensuring a webhook is present and valid


app/models/organization_webhook.rb
```diff
@@ -57,7 +57,7 @@ def admin_org_hook_scoped_github_client
   #
   # Returns true if successful, otherwise raises a GitHub::Error or ActiveRecord::RecordInvalid.
   def create_org_hook!(client)
-    github_organization(client).create_organization_webhook(config: { url: webhook_url }).id
+    self.github_id = github_organization(client).create_organization_webhook(config: { url: webhook_url }).id
```
BenEmdon: The `self.github_id` was previously missing meaning this method only created webhooks and didn't save their IDs.

app/models/organization_webhook.rb
```diff
@@ -90,6 +90,7 @@ def activate_org_hook(client)
   # of a large size. Invoke cautiously.
   def ensure_webhook_is_active!(client: nil)
     client ||= admin_org_hook_scoped_github_client
+    retrieve_org_hook_id!(client) if github_id.blank?
```
BenEmdon: To reiterate: if the `github_id` field on the record is blank, we double check that the webhook doesn't;t already exist.

Issue #1626, opened 2018-10-03
Classroom roster - can't see original id in assignments


palves-ulht:
### Feature request :sparkles:
I setup a classroom, using student IDs to fill the roster (nice feature, BTW!).

My assignment looks like this (you can see two student IDs: 21700000 and 21800000):
![image](https://user-images.githubusercontent.com/10359239/46408362-cf383300-c709-11e8-8a40-67c49110f7d4.png)

But when students accept the assignment and associate their github account with their student ID, I lose that information in the assignment:
For example, @palves-student is associated with 21700000 but I can't see that:
![image](https://user-images.githubusercontent.com/10359239/46408379-e414c680-c709-11e8-8bd3-099245458976.png)

This also happens in the Classroom Assistant:
![image](https://user-images.githubusercontent.com/10359239/46408393-f0008880-c709-11e8-9301-3723a9b521b9.png)

Would it be possible to show that information in both views? This is crucial to teachers, otherwise we'll have to ask our students to create a github account that includes their student ID (not very elegant...)

winterer:
@palves-ulht you can see all student IDs and their linked github account on the roster management page (On the classroom page, navigate to "Manage classroom" => "Roster management").
However, having the student IDs on the assignment pages too would really be an improvement, indeed.

jobartucz:
I was going to submit a request for exactly this! Thank you. While you are at it, on the same page, could you add the date of the last commit to the repository? That way, in one screen, I can see all of my students and whether they have submitted anything new to the assignment. It would make grading so much easier! (I can also add a separate feature request if you think it is warranted)

![screen shot](https://user-images.githubusercontent.com/26631024/52906165-0700cc80-320c-11e9-99e1-c2b6184c3974.png)

Issue #630, opened 2016-06-11
Repositories' names with a common pattern


drbeco:
Hi guys, this is a feature request.

I always tell my students to follow some standard names so I can keep track of what is what.

But that is always prone to error (take off the human factor whenever you can)

So it would be nice if the name was automatically generated.

I have a few ideas:

* The simplest: just a number. It would be like: `assignment-title-XXX`. The order would be the creation order. Very easy to implement. For example: `galaxy-001`, `galaxy-002`, and so on...
* More freedom: the teacher sets some `substring` to be used. The repository would be: `assignment-title-substring-XXX`. For example, consider the substring `ex3-grp`: `invaders-ex3-grp-001`, `invaders-ex3-grp-002`, and so on...
* Yet more freedom: a list of substrings that could be used. For example: `pacman-blinky`, `pacman-pinky`, `pacman-inky` and `pacman-clyde` for a maximum of 4 groups.

Thank you for considering the idea.

drbeco:
Bellow some real examples I currently use, but that I needed to emphasize very much in class to the students, so they would not forget to follow (that is very tiresome).

![image](https://cloud.githubusercontent.com/assets/5805111/15983238/3520aa80-2f77-11e6-813d-7ec423f5f047.png)

![image](https://cloud.githubusercontent.com/assets/5805111/15983240/3a5653a6-2f77-11e6-913f-460a2881e5d1.png)

![image](https://cloud.githubusercontent.com/assets/5805111/15983241/3d568a6c-2f77-11e6-9a36-6a4db9051478.png)

![image](https://cloud.githubusercontent.com/assets/5805111/15983242/432b0cba-2f77-11e6-840d-681070281d95.png)

![image](https://cloud.githubusercontent.com/assets/5805111/15983245/491fa996-2f77-11e6-8a72-50c77f98d0d9.png)

---

My rules for those above was:

* for some of them I determined the whole name
* for some I said what would be the `prefix` and told them to complete with a creative word for the team. This is really nice, because they feel more responsible for the creation.
* also I could say just the `sufix` or both `sufix`+`prefix` and let them complete.

aaossa:
I need a subset of this, I would like to customize prefix **and sufix** for each repository. I would imagine something like this in assignment settings:

![image](https://user-images.githubusercontent.com/10425834/30776786-1c2f73e8-a083-11e7-933f-b7f7088306ca.png)

Is it possible to do this? Can I work in a PR?

[...]

PR #1722, opened 2018-12-20
Add option to use Roster ID for Assignment-Repo suffix

ReshefElisha:
## What
_What does your pull request propose?_ :sparkles:
This PR implements a way to use the existing roster to organize student assignment-repos. A checkbox is added to assignment creation enabling the instructor to specify the use of the student's roster-id instead of their github-username as assignment-repo suffix.

This applies to #630 and partially to #1626

## Heads up
_app/views/assignment_invitations/show.html.erb_ still uses _"#{current_assignment.slug}-#{current_user.github_user.login}"_ for the displayed repo name. This has no bearing on functionality but may confuse users as the final repo name will not be the one displayed. I'm unfamiliar with RoR so I'm not sure what the visibility of the generate_github_repository_name function is to the view. If possible, it should be used anywhere the repo-name needs to be used or shown. Please advise how to proceed.


ReshefElisha:
re: the assignment_invitation issue: It's also inaccurate in case a repo needs a suffix. This is a larger scope so may warrant a different PR? Currently in that view the AssignmentRepo current_submission is nil, until the user chooses to accept the invitation. This may mean the generate_github_repo_name method needs to be reimplemented in the assignment_invitation controller to remove depending on having a current_submission.

PR #1727, opened 2019-01-02
Fetch GitHub Team name for Classroom Assistant

srinjoym:
Fetch the team name from GitHub everytime we serialize a group assignment repo. This fixes https://github.com/education/classroom-assistant/issues/117

### DIFF
--- a/app/serializers/group_assignment_repo_serializer.rb
+++ b/app/serializers/group_assignment_repo_serializer.rb
@@ -6,7 +6,7 @@ class GroupAssignmentRepoSerializer < ActiveModel::Serializer
   attributes :displayName
 
   def username
-    object.group.title
+    object.group.github_team.name
   end
 
   # rubocop:disable MethodName


andrewbredow:
@srinjoym Is this still WIP? Should we close it?

app/serializers/group_assignment_repo_serializer.rb
```diff
@@ -6,7 +6,7 @@ class GroupAssignmentRepoSerializer < ActiveModel::Serializer
   attributes :displayName
 
   def username
-    object.group.title
+    object.group.github_team.name
```
jeffrafter: I'm trying to understand the implications of fetching from the API here; If I am understanding correctly we fetch from the API every time we access the `github_*` properties. This means every time we serialize a `GroupAssignmentRepo` to JSON we will hit the github API. The main place this feels problematic is here: https://github.com/education/classroom/blob/8be229a98e7a56b808e977472c46d48b2d62c2ce/app/controllers/api/group_assignment_repos_controller.rb#L11-L12 

This introduces and github API N+1 if I am reading this right. I'm not clear if that is acceptable.

andrewbredow: Depending on our setup, I think OctoKit would [cache](https://github.com/plataformatec/faraday-http-cache#shared-vs-non-shared-caches) these at the client level?

d12: Octokit should cache them, correct

PR #1736, opened 2019-01-15
Multiple Classrooms per Org UI Tweaks

srinjoym:
Updated the flash banner message when an org is deleted and moved the background generation for each Classroom page onto the `Organization` model from `GitHub_Organization` (also allows us for more UI upgrades coming in the future 😄)


Magistern:
so, when can we try out the multiple classroom update? 👻

BenEmdon:
@Magistern Teachers who complete our [Campus Advisors](https://education.github.com/teachers/advisors) training have early access to new features, like this one, in private beta.

PR #1769, opened 2019-02-19
config/application: improve status checks.

MikeMcQuaid:
- Improve formatting by always adding a newline after each `raise` (as the current version of RuboCop recommends)
- Increase the timeout from 1 second to 5 seconds for all networked, GitHub API operations. As-is I could regularly reproduce timeouts by refreshing the `/_ping` endpoint ~10 times. This in turn results in flaky Pingdom checks which makes Classroom seem like it's down when it is not.
- Parse the new GitHub Status API and look specifically for the GitHub API as that's what we care about (e.g. not Issues and Pull Requests)
- Make the rate limit remaining considerable more conservative. Having 1,999 remaining requests should not make the service report a failing state.

CC @d12 and @johndbritton for review and thoughts.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -76,24 +76,37 @@ module GitHubClassroom
           # necessarily need replicas.
           # Docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
           raise "Elasticsearch status is #{status}" unless %w[green yellow].include?(status)
+
           "ok"
         end
 
-        ping.check :github do
-          uri = URI("https://status.github.com/api/status.json")
-          status = JSON.parse(Net::HTTP.get(uri))["status"]
-          raise "GitHub status is #{status}" unless status == "good"
+        ping.check :github, timeout: 5 do
+          uri = URI("https://www.githubstatus.com/api/v2/components.json")
+          status_components = JSON.parse(Net::HTTP.get(uri))["components"]
+          raise "GitHub status format is invalid!" if status_components.blank?
+
+          github_status_api_id = "brv1bkgrwx7q"
+          api_component = status_components.find do |component|
+            component["id"] == github_status_api_id
+          end
+          raise "GitHub API status ID (#{github_status_api_id}) cannot be found!" if api_component.blank?
+
+          api_status = api_component["status"]
+          raise "GitHub API status is #{status}" if api_status != "operational"
+
           "ok"
         end
 
-        ping.check :github_api do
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -76,24 +76,36 @@ class Application < Rails::Application
...
+
           "ok"
         end
 
-        ping.check :github do
-          uri = URI("https://status.github.com/api/status.json")
-          status = JSON.parse(Net::HTTP.get(uri))["status"]
-          raise "GitHub status is #{status}" unless status == "good"
+        ping.check :github, timeout: 5 do
+          status_components = JSON.parse(Net::HTTP.get(uri))["components"]
```
d12: Where does `uri` come from here?

talsafran: Good catch, I think it was accidentally removed from the diff.

```suggestion
          uri = URI("https://status.github.com/api/status.json")
          status_components = JSON.parse(Net::HTTP.get(uri))["components"]
```

MikeMcQuaid: @d12 @talsafran Yup, missed on my `git add`! Adjusted to use the new GitHub Status API URI (`uri = URI("https://www.githubstatus.com/api/v2/components.json")`)

MikeMcQuaid:
@d12 @BenEmdon @talsafran Oops, fixed now! Ready to 🚢 now, thanks!

config/application.rb
```diff
@@ -76,24 +76,37 @@ class Application < Rails::Application
...
-          status = JSON.parse(Net::HTTP.get(uri))["status"]
-          raise "GitHub status is #{status}" unless status == "good"
+        ping.check :github, timeout: 5 do
+          uri = URI("https://www.githubstatus.com/api/v2/components.json")
+          status_components = JSON.parse(Net::HTTP.get(uri))["components"]
+          raise "GitHub status format is invalid!" if status_components.blank?
+
+          github_status_api_id = "brv1bkgrwx7q"
+          api_component = status_components.find do |component|
+            component["id"] == github_status_api_id
```
BenEmdon: What is the id `brv1bkgrwx7q`? Is this global to all integrators? Is this a secret?

MikeMcQuaid: It is the `id` for the GitHub Status API component from https://www.githubstatus.com/api/v2/components.json.

MikeMcQuaid:
I don't have privileges to merge this. @talsafran @d12 @BenEmdon @tarebyte @srinjoym Could one of you who do do so? Thanks!

PR #1782, opened 2019-02-25
Fix template error

BenEmdon:
## What
* Add nil safety


BenEmdon:
This super safe 1 liner is safe to merge. It's impact is only on staff tools.

PR #1791, opened 2019-03-07
Google Classroom Roster Integration

srinjoym:
Roster import with syncing from Google Classroom! ✨ 

Depends on #1827 

Todo:
- [x] Full Test Coverage
- [x] Create feature flag
- [ ] Add Experimental Feedback flag in #1835 
- [ ] Merge #1827


app/views/orgs/rosters/select_google_classroom.html.erb
```diff
@@ -0,0 +1,38 @@
...
+
+    <div class="my-3">
+      <h2 class="h3 mb-2">Select the Google Classroom roster you want to import</h2>
+
+      <div class="Box overflow-auto">
+        <ul>
+          <% @google_classroom_courses.each do |course| %>
+            <li class="Box-row">
+              <label>
+                <input type="checkbox" class="mr-2" checked="checked" aria-describedby="help-text-for-checkbox">
```
brandonrosage: Since you can only import (and associate with) one Google Classroom (in the interest of syncing later), I suspect these should be radio buttons.

app/views/orgs/rosters/select_google_classroom.html.erb
```diff
@@ -0,0 +1,38 @@
...
+                <%= course.name %>
+                <%# <%= link_to course.name, {controller: "rosters", action: :import_from_google_classroom, course_id: course.id}, method: :patch %>
+              </label>
+            </li>
+          <% end %>
+        </ul>
+      </div>
+    </div>
+
+    <div class="f4 text-gray text-italic my-3">or <a href="" class="btn-link">manually add students</a></div>
```
brandonrosage: This needs to be linked back to the default "manual" view for creating a roster.

.env-example
```diff
@@ -1,6 +1,10 @@
 GITHUB_CLIENT_ID=
 GITHUB_CLIENT_SECRET=
 
+# Used for Google Classroom Roster Integration
+GOOGLE_CLIENT_ID=
```
srinjoym: Used for communicating with Google Classroom API

app/models/roster/creator.rb
```diff
@@ -57,12 +57,15 @@ def initialize(organization:, identifier_name:, **options)
...
     #
+    # rubocop:disable Metrics/MethodLength
     def perform
       ensure_organization_does_not_have_roster!
 
       ActiveRecord::Base.transaction do
         @roster = Roster.new(identifier_name: @identifier_name)
-        add_identifiers_to_roster(@options[:identifiers]) if @options.key?(:identifiers)
+
+        google_ids = @options[:google_user_ids].nil? ? [] : @options[:google_user_ids]
```
srinjoym: Optional `:google_user_ids` key is used to associate Google Classroom student ID with the Roster Entry

app/models/roster/creator.rb
```diff
@@ -72,13 +75,15 @@ def perform
...
 
     private
 
-    def add_identifiers_to_roster(raw_identifiers_string)
+    def add_identifiers_to_roster(raw_identifiers_string, google_ids: [])
       identifiers = raw_identifiers_string.split("\r\n").reject(&:blank?).uniq
-      identifiers.each do |identifier|
-        @roster.roster_entries << RosterEntry.new(identifier: identifier)
+
+      identifiers.zip(google_ids).each do |identifier, google_user_id|
```
srinjoym: If there are no `google_ids` this just puts a nil for each roster entry

config/initializers/google_classroom_client.rb
```diff
@@ -0,0 +1,21 @@
...
+    scope = [
+      Google::Apis::ClassroomV1::AUTH_CLASSROOM_COURSES_READONLY,
+      Google::Apis::ClassroomV1::AUTH_CLASSROOM_ROSTERS_READONLY
+    ]
+
+    client_id = Rails.application.secrets.google_client_id
+    client_secret = Rails.application.secrets.google_client_secret
+    client = Google::Auth::ClientId.new(client_id, client_secret)
+
+    token_store = Google::Auth::Stores::RedisTokenStore.new(redis: GitHubClassroom.redis)
```
srinjoym: Using our existing redis store to store Google Classroom user tokens

srinjoym:
@BenEmdon When you get a chance could you do a preliminary review of this PR? Sorry for the huge size, the actual changes are all pretty much in the RostersController but the cassettes and gems increased the number of changed files. Working on a follow up PR with UI updates for this + throughout Classroom in #1835

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +175,60 @@ def download_roster
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/AbcSize
 
+    def select_google_classroom
```
srinjoym: Route used when displaying list of Google Classrooms to select

Gemfile
```diff
@@ -44,7 +46,6 @@ gem "local_time", "~> 2.0"
 gem "octicons_helper", "~> 2.1"
 gem "octokit", github: "octokit/octokit.rb", ref: "ffae5ddd6fd9da6f4538ca7adeb5d1768011610b"
 gem "octopoller",      "~> 0.1"
-gem "omniauth",        "~> 1.6", ">= 1.6.1"
```
BenEmdon: Don't we need omniauth?

app/views/orgs/rosters/_google_classroom_collection.html.erb
```diff
@@ -0,0 +1,27 @@
...
+                new_roster_path,
+                class: "btn-link"
+    %>
+  </div>
+
+  <div class="d-flex flex-items-center border-top pt-5">
+    <%= submit_tag 'Import roster', class: 'btn btn-primary mr-3' %>
+    <%= link_to 'Skip', organization_path(current_organization), class: 'btn-link' %>
+  </div>
+<% end %>
```
BenEmdon: Nit: breakline before the EOF
```suggestion
<% end %>

```

package-lock.json
```diff
@@ -0,0 +1,945 @@
+{
```
BenEmdon: How come we haven't had a `package-lock.json` until now?

config/initializers/google_classroom_client.rb
BenEmdon: Why did you decide to use Redis instead of Postgres?

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +175,60 @@ def download_roster
...
+    # rubocop:disable Metrics/AbcSize
+    def search_google_classroom
+      courses_found = fetch_all_google_classrooms.select do |course|
+        course.name.downcase.include? params[:query].downcase
+      end
+
+      respond_to do |format|
+        format.html do
+          render partial: "orgs/rosters/google_classroom_collection",
+                 locals: { courses: courses_found }
```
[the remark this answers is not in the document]
srinjoym: Tried to add this but it hits a Linter Error for AlignHash https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_layout.md#layoutalignhash.

I think it makes sense to update the rubocop rule, didn't want to turn it off for everything because we still want the check but I'm sure there's some configuration we can find in a future PR.

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,88 @@ def get_user_to_group_hash(grouping)
...
+      @google_classroom_service.client_options.application_name = "GitHub Classroom"
+      @google_classroom_service.authorization = user_google_classroom_credentials
+    end
+
+    # Helper method for getting current user's google classroom credentials
+    def user_google_classroom_credentials
+      google_classroom_client = GitHubClassroom.google_classroom_client
+      user_id = current_user.github_user.login
+
+      google_classroom_client.get_credentials(user_id, request)
```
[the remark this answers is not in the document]
srinjoym: This is the key that is stored in Redis for looking up the Google Classroom user's credentials for signing into the API

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +175,60 @@ def download_roster
...
+        redirect_to roster_path(current_organization)
+      else
+        current_organization.update_attributes!(google_course_id: params[:course_id])
+        add_google_classroom_students(students)
+      end
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    def sync_google_classroom
+      return if current_organization.google_course_id.nil?
```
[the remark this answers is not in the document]
srinjoym: That's a good question, not sure what to do here because if we're trying to sync a google classroom roster and we don't have an ID it's in a pretty broken state 🤔 I think if we just flash an error it should be fine?

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +175,60 @@ def download_roster
...
+
+    def import_from_google_classroom
+      students = list_google_classroom_students(params[:course_id])
+      return if students.nil?
+
+      if students.blank?
+        flash[:warning] = "No students were found in your Google Classroom. Please add students and try again."
+        redirect_to roster_path(current_organization)
+      else
+        current_organization.update_attributes!(google_course_id: params[:course_id])
```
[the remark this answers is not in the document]
srinjoym: Great catch, just updated this to not skip vallidations

app/controllers/orgs/rosters_controller.rb
```diff
@@ -36,8 +59,9 @@ def new
     def create
       result = Roster::Creator.perform(
         organization: current_organization,
-        identifier_name: params[:identifier_name],
-        identifiers: params[:identifiers]
+        identifier_name: "Identifiers",
```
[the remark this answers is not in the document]
srinjoym: Great idea, just moved it downstream to `Roster::Creator`

package-lock.json
srinjoym: Whoops, must've used npm instead of yarn at some point, deleted the file

srinjoym:
@BenEmdon Just addressed a bunch of the comments, thanks for the great feedback! Couldn't reply to the PG vs Redis comment so I'll just post here. The Google APIs authentication module `Google::Auth` supports two main types of token storage: Redis and Files. I figured since we already have a Redis store we're using we could just use that for the user credentials as well. Here's some more documentation on `Google::Auth`: https://www.rubydoc.info/github/google/google-auth-library-ruby/Google/Auth/TokenStore

app/controllers/orgs/rosters_controller.rb
```diff
@@ -76,7 +99,7 @@ def link
       user_id = params[:user_id].to_i
       raise ActiveRecord::ActiveRecordError unless unlinked_user_ids.include?(user_id)
 
-      current_roster_entry.update_attributes!(user_id: user_id)
+      current_roster_entry.update_attributes(user_id: user_id)
```
BenEmdon: `update_attribute` skips validations. Use `update` instead to _not_ skip validations. 
See https://davidverhasselt.com/set-attributes-in-activerecord/ for details.
```suggestion
      current_roster_entry.update(user_id: user_id)
```

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,58 @@ def download_roster
...
+
+    def import_from_google_classroom
+      students = list_google_classroom_students(params[:course_id])
+      return if students.nil?
+
+      if students.blank?
+        flash[:warning] = "No students were found in your Google Classroom. Please add students and try again."
+        redirect_to roster_path(current_organization)
+      else
+        current_organization.update_attributes(google_course_id: params[:course_id])
```
BenEmdon: `update_attribute` skips validations. Use `update` instead to _not_ skip validations. 
See https://davidverhasselt.com/set-attributes-in-activerecord/ for details.
```suggestion
        current_organization.update(google_course_id: params[:course_id])
```

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,58 @@ def download_roster
...
+
+      new_student_ids = latest_student_ids - current_student_ids
+      new_students = latest_students.select { |s| new_student_ids.include? s.user_id }
+
+      add_google_classroom_students(new_students)
+    end
+    # rubocop:enable Metrics/AbcSize
+
+    def unlink_google_classroom
+      current_organization.update_attributes!(google_course_id: nil)
```
BenEmdon: `update_attribute` skips validations. Use `update` instead to _not_ skip validations. 
See https://davidverhasselt.com/set-attributes-in-activerecord/ for details.
```suggestion
      current_organization.update!(google_course_id: nil)
```

app/controllers/orgs/rosters_controller.rb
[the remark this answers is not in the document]
BenEmdon: Flashing an error sounds like a good idea 👍

app/controllers/orgs/rosters_controller.rb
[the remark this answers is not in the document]
BenEmdon: If a user changed their name would this break?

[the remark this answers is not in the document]
thebradbain: @srinjoym is there any reason we can't simply store a classroom user's `id` or, if it needs to use the github_user object, the (unchanging?) `user,github_user.user_id` as the key?

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +308,88 @@ def get_user_to_group_hash(grouping)
...
     end
+
+    # Returns name of the linked google course to current organization (for syncing rosters)
+    def current_organization_google_course_name
+      return nil if current_organization.google_course_id.nil?
+      authorize_google_classroom
+      course = @google_classroom_service.get_course(current_organization.google_course_id)
+      return course.name unless course.nil?
+      nil
+    rescue Google::Apis::Error
```
stephaniegiang: Needs another indentation

stephaniegiang:
@srinjoym won't be working on this PR anymore. So I'm going to take over!

app/models/roster/creator.rb
stephaniegiang: I believe so. If there is an existing roster and we import Google Classroom, there will be more roster entries than the google_ids.

d12:
Ping me when this is ready for review!

app/models/roster/creator.rb
stephaniegiang: Just talked to @srinjoym and there would not be a chance that these are uneven. We are grabbing the google classroom identifiers and ids to the roster entries.

stephaniegiang:
@d12 ready for review! Thanks!

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,61 @@ def download_roster
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/AbcSize
 
+    def select_google_classroom
+      @google_classroom_courses = fetch_all_google_classrooms
+    end
+
+    def search_google_classroom
+      courses_found = fetch_all_google_classrooms.select do |course|
+        course.name.downcase.include? params[:query].downcase
```
d12: Can we use a SQL `LIKE` condition here instead of doing filtering in Ruby?

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,61 @@ def download_roster
...
+        format.html do
+          render partial: "orgs/rosters/google_classroom_collection",
+                 locals: { courses: courses_found }
+        end
+      end
+    end
+
+    def import_from_google_classroom
+      students = list_google_classroom_students(params[:course_id])
+      return if students.nil?
```
d12:
```suggestion
      return unless students
```

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,88 @@ def get_user_to_group_hash(grouping)
...
+      nil
+    rescue Google::Apis::Error
+      nil
+    end
+
+    # Returns list of students in a google classroom with error checking
+    # rubocop:disable Metrics/MethodLength
+    def list_google_classroom_students(course_id)
+      response = @google_classroom_service.list_course_students(course_id)
+      response.students ||= [] # Set to empty array if no students in course
```
d12: Do you mind adding a slightly more explicit `response.students` line under this to show that's the return value? it's a little tricky following how `response.students ||= []` returns response.students if it exists.

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,61 @@ def download_roster
...
+        redirect_to roster_path(current_organization)
+      else
+        current_organization.update(google_course_id: params[:course_id])
+        add_google_classroom_students(students)
+      end
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    def sync_google_classroom
+      if current_organization.google_course_id.nil?
```
d12:
```suggestion
     unless current_organization.google_course_id
```

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,61 @@ def download_roster
...
+        flash[:error] = "No Google Classroom has been linked. Please link Google Classroom."
+        return;
```
d12: :fire: ; :fire:

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,88 @@ def get_user_to_group_hash(grouping)
...
+
+    # Add Google Classroom students to roster
+    def add_google_classroom_students(students)
+      names = students.map { |s| s.profile.name.full_name }
+      user_ids = students.map(&:user_id)
+      params[:identifiers] = names.join("\r\n")
+      params[:google_user_ids] = user_ids
+
+      if current_roster.nil?
+        create
```
d12: This pattern of calling controller actions from within controller actions can lead to some weird side effects. Can we abstract the logic you need into a method and then call that method from both controller actions?

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,88 @@ def get_user_to_group_hash(grouping)
...
+      else
+        add_students
```
d12: Same comment here

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,88 @@ def get_user_to_group_hash(grouping)
...
+      end
+
+      courses
+    end
+
+    # Authorizes current user through Google and sets google_classroom_service
+    # Used as a before_action before routes which require Google authorization
+    def authorize_google_classroom
+      google_classroom_client = GitHubClassroom.google_classroom_client
+      if user_google_classroom_credentials.nil?
```
d12:
```suggestion
      unless user_google_classroom_credentials
```

db/schema.rb
```diff
@@ -110,6 +110,7 @@
     t.integer "max_members"
     t.boolean "students_are_repo_admins", default: false, null: false
     t.boolean "invitations_enabled", default: true
+    t.integer "max_teams"
```
d12: I thought this already merged? You might have some other commits in this PR.

config/initializers/rack_rewrite.rb
```diff
@@ -2,7 +2,7 @@
 
 # URL for Nuts Server deployed on Heroku, redirects download/update requests to latest release
 # on GitHub Releases with platform parameters
-RELEASE_SERVER_URL = "http://classroom-assistant-releases.herokuapp.com"
```
d12: Can this move into another PR incase we need to revert this?

db/schema.rb
stephaniegiang: Oops I will fix that.

stephaniegiang:
@d12 ready for another round of review

.env-example
d12: Can we add something to the README about this in the section about filling out the .env? Mainly is this optional? I have no idea how I'd go about getting a client ID and secret.

app/controllers/orgs/rosters_controller.rb
```diff
@@ -146,6 +174,61 @@ def download_roster
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/AbcSize
 
+    def select_google_classroom
+      @google_classroom_courses = fetch_all_google_classrooms
+    end
+
+    def search_google_classroom
+      courses_found = fetch_all_google_classrooms.select do |course|
+        course.where("name ILIKE ?", "#{params[:query]}#")
```
d12: Ah sorry, I think I may have left bad feedback here, I thought this returned an AR relation. It looks like this fires a SQL query per course which isn't great for performance. If this is an array and the data is already loaded, lets just do the filtering in memory here. Something like `course.name.downcase.include?(params[:query].downcase)`

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,89 @@ def get_user_to_group_hash(grouping)
 
       mapping
     end
+
+    # Returns name of the linked google course to current organization (for syncing rosters)
+    def current_organization_google_course_name
+      return nil if current_organization.google_course_id.nil?
```
d12: If you find yourself writing `.nil?` in a conditional, there's usually a cleaner way to write it 😅 

```suggestion
     return unless current_organization.google_course_id
```

app/controllers/orgs/rosters_controller.rb
```diff
@@ -228,6 +311,89 @@ def get_user_to_group_hash(grouping)
...
+      authorize_google_classroom
+      course = @google_classroom_service.get_course(current_organization.google_course_id)
+      return course.name unless course.nil?
+      nil
```
d12: This is a little odd, the fall through case of the method is the error case, and we bail early with the expected return type earlier in the method (320). People usually skim to the bottom of the method to see what it returns, so this is a little confusing.

Instead, we could replace 320 and 321 with `return unless course`, and `course.name` below.

 If we want to take it one step further, we can use the [safe navigation operator](https://stackoverflow.com/questions/36812647/what-does-ampersand-dot-mean-in-ruby) here and replace 320 and 321 with 

```ruby
course&.name
```

As an example:

```ruby
irb(main):004:0> (-5).abs
=> 5

irb(main):005:0> nil.abs
NoMethodError: undefined method `abs' for nil:NilClass
	from (irb):5
	from /usr/bin/irb:11:in `<main>'

irb(main):006:0> nil&.abs
=> nil

irb(main):007:0> (-5)&.abs
=> 5
```

app/models/roster/creator.rb
```diff
@@ -45,24 +47,26 @@ def failed?
...
+    # rubocop:disable Metrics/MethodLength
     def perform
       ensure_organization_does_not_have_roster!
 
       ActiveRecord::Base.transaction do
-        @roster = Roster.new(identifier_name: @identifier_name)
-        add_identifiers_to_roster(@options[:identifiers]) if @options.key?(:identifiers)
+        @roster = Roster.new(identifier_name: DEFAULT_IDENTIFIER_NAME)
+
+        google_ids = @options[:google_user_ids].nil? ? [] : @options[:google_user_ids]
```
d12:
```suggestion
        google_ids = @options[:google_user_ids] || []
```

config/routes.rb
```diff
@@ -80,6 +84,11 @@
           patch :delete_entry
           patch :add_students
           patch :remove_organization
+          patch :import_from_google_classroom
+          patch :sync_google_classroom
+          patch :unlink_google_classroom
+          get   :select_google_classroom, as: "select_google_classroom"
```
d12: Do we need this `as:`?

.env-example
stephaniegiang: Is it possible to add this later? This google integration PR is not quite ready as I need to polish it a bit more (duplicate roster entries and allow teachers to import different identifiers).

I'll also make a separate PR for this. This PR is quite large and I'm getting lost in the files changed already 😂

config/routes.rb
stephaniegiang: No we don't. I'll remove it.

stephaniegiang:
@d12 Ready for another review!

app/controllers/orgs/rosters_controller.rb
```diff
@@ -109,9 +132,14 @@ def delete_entry
     # rubocop:disable Metrics/AbcSize
     def add_students
       identifiers = params[:identifiers].split("\r\n").reject(&:blank?).uniq
+      google_ids = params[:google_user_ids] ? params[:google_user_ids] : []
```
d12:
```suggestion
      google_ids = params[:google_user_ids] || []
```

package.json
```diff
@@ -9,7 +9,7 @@
     "jquery-readyselector": "git+https://git@github.com/Verba/jquery-readyselector.git",
     "jquery-ujs": " ~1.2.2",
     "jquery.turbolinks": "git+https://git@github.com/kossnocorp/jquery.turbolinks#2.1.0",
-    "primer": "^10.10.4",
+    "primer": "^10.10.5",
```
d12: Why was this change necessary?

db/schema.rb
d12: It looks like this reappeared again

spec/controllers/orgs/rosters_controller_spec.rb
```diff
@@ -135,8 +117,7 @@
           before do
             post :create, params: {
               id: organization.slug,
-              identifiers: "    \r\n ",
-              identifier_name: "emails"
```
d12: There are a bunch of changes here to identifier_name stuff, didn't this happen in a previous PR? If not, can we move to a different PR? It's unrelated to the Google Classroom stuff and this PR is already massive.

spec/controllers/orgs/rosters_controller_spec.rb
```diff
@@ -261,6 +242,189 @@
...
+      end
+    end
+
+    context "with flipper disabled" do
+      before do
+        get :show, params: { id: organization.slug }
+      end
+
+      it "404s" do
+        expect(response).to have_http_status(:not_found)
```
d12: Why is `:show` 404ing here? Shouldn't `show` render fine regardless of whether or not the flipper is enabled?

spec/controllers/orgs/rosters_controller_spec.rb
```diff
@@ -684,4 +848,415 @@
...
+                id: organization.slug
+              }
+            end
+
+            it "doesn't add any students" do
+              expect(organization.roster.roster_entries.count).to eq(1)
+            end
+          end
+
+          context "classroom has a linked google course" do
```
d12: Amazing tests in this context :sparkles:

spec/controllers/orgs/rosters_controller_spec.rb
stephaniegiang: Yeah we should be linking to the Google Classroom setup instead. Not sure why this is passing. Just tested it manually and it isn't 404ing 😕

d12: Could you dig into why that's happening in the test? That's suspicious...

stephaniegiang: This test was passing because we did not enable the `:student_identifier` flag. We should really get rid of all the feature flags that been at 100% for a while.

stephaniegiang: We also already have a test covering the google class setup route. So I'll just modify this test to pass if the `:google_classroom_roster_import` is disabled.

package.json
stephaniegiang: Ooops I'll remove that

stephaniegiang:
@d12 ready for another round of review!

db/schema.rb
```diff
@@ -163,6 +163,17 @@
     t.index ["user_id"], name: "index_invite_statuses_on_user_id"
   end
 
+  create_table "lti_configurations", force: :cascade do |t|
```
stephaniegiang: Not sure why everytime I pull master, this file always comes back. But I'll fix this 😄

d12:
Teachers are very excited for this ship! Let me know how I can help this move forward.

stephaniegiang:
Just pulled in master, so hopefully if the tests pass, another round of reviews. Then we can merge!

stephaniegiang:
@d12 officially ready for another review!

config/initializers/rack_rewrite.rb
```diff
@@ -2,7 +2,7 @@
 
 # URL for Nuts Server deployed on Heroku, redirects download/update requests to latest release
 # on GitHub Releases with platform parameters
-RELEASE_SERVER_URL = "https://classroom-assistant-releases.herokuapp.com"
+RELEASE_SERVER_URL = "http://classroom-assistant-releases.herokuapp.com"
```
d12: Why are we making this change?

stephaniegiang: Oopps the merge of master brought that back

stephaniegiang:
PR will be deployed once we get proper google credientials

PR #1794, opened 2019-03-11
Markdown Rendering for Help Pages

srinjoym:
Pulling in markdown rendering using ~~Redcarpet~~ [Kramdown](https://github.com/gettalong/kramdown) for the new help pages from @ddbeck!

Created new routes at `/help/(:article_name)` so we can render any markdown page in the Classroom Docs folder. If there's no `:article_name` specified (i.e. classroom.github.com/help) then we default to the help landing page. I also pulled in the top level header from our other static pages which has a `Sign In` button when the user is not logged into Classroom, and a `Back to Classroom` button when the user is logged in.

This is still a WIP, custom styles from primer and commands on the probot page like `title:` are not rendering correctly right now.

Todo:
- [x]  Fix rendering of custom styles from Primer
- [x]  Figure out a good way to navigate back through help pages (maybe a header with links to navigate back to help home)
- [x]  Make sure there aren't any issues with security on the route/markdown rendering
- [x]  Add tests for static routing

Screenshots:
screencapture-localhost-5000-help-2019-03-11-11_26_57


ddbeck:
Hi @srinjoym, I've made some progress on this, but I could use some guidance from you to really bring all these issues to completion. Mostly this relates to my inexperience with Rails (sorry!). Some notes on the work I've done so far:

* I swapped out the redcarpet gem for kramdown, so we could style paragraphs without resorting to using raw HTML everywhere. The API is very similar and the terms of the license are the same, so I don't think this represents any issues, but let me know if you want me to reevaluate that decision.
* I've added a proof-of-concept for the breadcrumbs, though the deepest breadcrumb (the current page) isn't terribly useful, since it's derived from the filename, rather than the actual title of the page. We could probably add some YAML frontmatter to the docs and set a title that way, but we'd probably need another dependency (maybe [F-3r/ruby_front_matter](https://github.com/F-3r/ruby_front_matter) or something like it). Any opnions on this?
* I'm having a little trouble with the URLs for pages beneath `/help`. Both `/help` and `/help/` work, which complicates relative links within the docs' Markdown source (i.e., you have to give an absolute path like `/help/settings-probot`, which is a bit limiting). Is there a way to redirect to a slashed or slashless URL so we can have a canonical URL to work from?
* I'm not sure how to get started testing the static routing because I've never done any Rails testing before. 😳 Can you point me to an example which might help me get started on that?
* Along the same lines, I'm not clear on how Rails handles image assets. Right now, the screenshot in `docs/create-group-assignments.md` uses the raw file URL on GitHub, rather than serving out of the app. I tried adding a gif to `docs/probot-settings.md` and I couldn't coax the app to serve the gif from `/assets`. Any hints on linking to assets from the Markdown source and having it actually work would be greatly appreciated.

Thanks in advance for your help on this stuff, Srinjoy!

ddbeck:
@srinjoym I'm sure you're busy at the moment, but if you get a chance, let me know if you have a particular timeline in mind for this—that way I can plan my follow-ups to avoid needlessly bothering you. Cheers! 😄

srinjoym:
Hi @ddbeck! Sorry for the delay on this, I've been busy working on some other things + school so haven't gotten a chance to look at this too in depth yet. 

I don't have a particular timeline in mind for this. Adding this in will be a big help for teachers but we also don't have to ship it asap since most teachers are wrapping up their Spring semesters. I think for a first goal, let's aim for the end of this month, let me know if this sounds good.

- Kramdown is a great find, looks like it'll be a better fit for these pages. We just have to take a look into the security of the HTML rendering and see if there's any way to restrict what elements are rendered and what links can be put in.
- The bread crumbs look great! Since most users will likely use it only to go back to a previous page, I don't think we need to prioritize fixing the deepest breadcrumb for now. Maybe we can follow this up with another PR to get that fixed?
- I took a look at trying to get relative URLs to work this morning and couldn't find a good solution either. For now, will we hit any problems if we prefix all the paths with `/help/<filename>`?
- I think the best place to start for testing would be to add a spec in https://github.com/education/classroom/blob/halp-redcarpet-test/spec/controllers/pages_controller_spec.rb to test the `#help` route. We can also add specs for each route we expect to work (like create-group-assignments) so it's easier to catch errors when we rename things in the future.

Let me know if you have any other questions, I'll try to follow up if I can find an actual solution for the relative links and images. Great work so far! ✨

ddbeck:
@srinjoym thank you! All of this sounds completely reasonable to me. Now that you say it, it seems obvious that we should do the basic things now and add more features and polish in subsequent PRs.

I'm going on a little vacation later this week, but I'll pick up some of these tasks starting late next week. Thank you again!

ddbeck:
Hey @srinjoym, it's my turn to apologize for the delay here. I've finally picked this back up and have some progress to show you:

* Fixed the issue with assets not loading ✔️ 
* Added tests for the pages we have now ✔️ 
* Switched to absolute links ✔️

The last thing is security. I'm not actually sure it's a problem that needs solving, really. As long as we're not accepting user-generated content as help pages, we don't need to sanitize the HTML.

Let me know if you have any more feedback or questions for me to move this along. And no rush on this—I'm happy letting this go on your schedule.

srinjoym:
Hi @ddbeck!

The help pages looks fantastic! I'll push up a commit with a small tweak to move the Rspec tests but other than that I think this is pretty much ready to ship. We can make small tweaks as we go along and get more feedback. The final step here is to run in through a design review. 

@brandonrosage could you take a look at this when you get a chance and let us know what tweaks we should make on the design front? One thing that we probably want to fix is that the root breadcrumb is a link to go back to the Classroom Dashboard, and we also have a button at the top of the page to go back to the Dashboard.

Screen Shot 2019-05-04 at 4 47 12 PM

ddbeck:
@srinjoym glad this is shaping up! I just realized that I made a mistake last week and made a change that never ended up in this PR. Is it OK if I add another doc to this PR? Or is it too late and I should wait to open a follow up PR?

srinjoym:
@ddbeck you can definitely add the change to the PR!

ddbeck:
@srinjoym thank you! Done!

Standing by for any changes needed in review.

srinjoym:
@brandonrosage friendly nudge 🙂 did you get a chance to take a look at this?

brandonrosage:
Just polished the `/help` page's layout 👍

ddbeck:
Hi @srinjoym, what's the next thing to do to wrap this up?

EricPickup:
I've replaced the `Video Tutorials` button on the header with `Help` button. I also moved it in the order of buttons (shown in the image below).

**Before**

image

**After**

image

EricPickup:
@srinjoym is there anything else that needs to be done for this?

app/controllers/pages_controller.rb
```diff
@@ -28,4 +28,16 @@ def assistant
     return not_found unless assistant_landing_page_enabled? || public_assistant_landing_page_enabled?
     render layout: "layouts/pagesv2"
   end
+
+  def help
+    file_name = params[:article_name] ? params[:article_name] : "help"
+    @file = File.read(Rails.root.join("docs", "#{file_name}.md"))
```
d12: We can't just blindly join with raw user input here, we're opening ourselves up to path traversal attacks. Can you find a way to sanitize this input?

```ruby
irb(main):003:0> Rails.root
=> #<Pathname:/Users/d12/dev/events_remote_loader_example>

irb(main):004:0> Rails.root.join("..")
=> #<Pathname:/Users/d12/dev>

irb(main):006:0> Rails.root.join("/etc/passwd")
=> #<Pathname:/etc/passwd>

irb(main):007:0> File.read(Rails.root.join("/etc/passwd"))
=> "##\n# User Database\n# \n# Note that this file is ..........
```

EricPickup: Good catch. How does this look? https://github.com/education/classroom/pull/1794/commits/147bfe97fe2e065b4af8a4ddf41c72b49d983193

PR #1817, opened 2019-05-01
Drop Ticker Stat Schema

srinjoym:
I brought in the `ticker_stats` table into `schema.rb` inadvertently when we were working on the Classroom Assistant launch. At the time we were planning on adding a repository ticker visualization to the Classroom homepage which has since been scrapped. The actual migration file was never committed so rails automatically deletes this snippet every time the app is run locally right now.


BenEmdon:
Maybe I have done something wrong –but every time I migrate the schema this schema reappears 🤔

PR #1818, opened 2019-05-03
Integrade Docker development environment, works cross platform (including Windows!)

DavidGriswoldTeacher:
As a windows user, Ruby work is hard for me, so I decided to expand the already-existing docker component of this project into an optional self-contained dev environment that allows me to do everything, including the ruby and rails work itself, in docker. It was a bigger project than I expected, but it now works very smoothly on my Windows machine, and could provide a nice entryway to developing on this for both Windows users and users who simply don't want to deal with a native ruby install.

I will leave comments explaining the elements briefly.


.gitignore
```diff
@@ -13,6 +13,7 @@ vendor/bundle
 !/log/.keep
 /tmp
 /coverage
+yarn-error.log
```
DavidGriswoldTeacher: this file came up in testing, and I realized it should be .gitignored.

Dockerfile-sc
```diff
@@ -0,0 +1,45 @@
+FROM ruby:2.4.2
```
DavidGriswoldTeacher: This is the docker file to build the docker image that will do the development. It starts with the base ruby image for the ruby version currently used in the project,  but for easy forward compatibility it installs rbenv as well.

Procfile.dev.sc
```diff
@@ -0,0 +1,2 @@
+rails: ./bin/rails server -p 5000 -b 0.0.0.0
```
DavidGriswoldTeacher: had to make a special procfile to get the binding right on rails. If there's a way to parse environment variables internally in the procfile we could do that here.

docker-compose.sc.yml
```diff
@@ -0,0 +1,70 @@
+version: '2.1'
+services:
+  rubyrails:
+    container_name: "classroom_rubyrails"
+    build: 
+      context: .
+      dockerfile: Dockerfile-sc
+    volumes: 
+      - .:/usr/src/app
```
DavidGriswoldTeacher: .bundle and vendor are copied into the docker image, then not mounted back onto the host box. That way dependencies we shouldn't be editing don't clutter up the filesystem.

docker-compose.sc.yml
```diff
@@ -0,0 +1,70 @@
...
+      - /usr/src/app/.bundle
+      - /usr/src/app/vendor
+    ports: 
+      - "5000:5000"
+    entrypoint: bash
```
DavidGriswoldTeacher: the dev environment simply opens a shell. Users attach to it and run script/setup and script/server etc just as they would for a native installation.

jeffrafter:
As a note we've updated Ruby to 2.6.4. It might make things easier (more modern distros available)

DavidGriswoldTeacher:
I won't be able to play with this for a while, unfortunately, as the school
year has started and my time spent USING github classroom has increased
while my time spent THINKING about it has shrunk. I still want to use this
environment to develop some more classroom features, but it might be next
summer before I can find the time again.

PR #1833, opened 2019-05-15
Added a flash message to direct long import times to GitHub support

BenEmdon:
## What
This PR proposes we add a new flash message which appears when importing starter code:

The flash message says:
> _"If importing starter code takes a very long time, please [contact GitHub Support](https://github.com/contact)."_

## Motivation
Whenever the source importer experiences slow import times, students and teachers flock to the [education/classroom](https://github.com/education/classroom) project to create [issues](https://github.com/education/classroom/issues?utf8=✓&q=is%3Aissue+code+import+). Since the source import API is maintained by GitHub.com, contacting GitHub support would be more desirable, since support is better equipped to address these situations.

### DIFF
--- a/app/assets/javascripts/setup.js
+++ b/app/assets/javascripts/setup.js
@@ -1,6 +1,7 @@
 (function() {
   var POLL_INTERVAL = 1000;
   var PROGRESS_HALF_LIFE = 1000;
+  var contact_support_message = "\nIf importing starter code takes a very long time, please <a href=\"https://github.com/contact\">contact GitHub support</a>.";
   var progress_asymptotically,
     create_flash_container,
     display_progress,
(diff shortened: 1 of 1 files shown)


srinjoym:
@BenEmdon Would it make sense to add the new message to the first flash message (the one that says Importing Starter Code)?

BenEmdon:
Whoops ☝️

BenEmdon:
## Update
Message now resides in the same flash as the _"Importing starter code"_ message:

BenEmdon:
After some discussions with GitHub support, adding the _contact support_ link might not help, since support doesn't have the tooling and knowledge to troubleshoot these situations.

Instead, we could provide more transparency into the repo which has already been created (which we removed in #1553).

PR #1836, opened 2019-05-17
Show Repo URL when created (while still importing)

BenEmdon:
## What
This PR adds the repo URL to the assignment invitation process when a repo has successfully been created:

## Motivation
This proposed change will increase transparency of the assignment setup process by letting students see their newly created GitHub repository (while the source import completes).

This will also allow students to report slow imports directly to GitHub support (if they experience a slow import) via the GitHub.com user experience.

This provides an alternative approach to #1833


BenEmdon:
Next, the same for group assignment repository setup 🔜

PR #1838, opened 2019-05-21
Show Repo URL when created (while still importing) [Group Assignment Setup]

BenEmdon:
## What
his PR adds the repo URL to the assignment invitation process when a repo has successfully been created:

Screen Shot 2019-05-21 at 4 54 33 PM

The same thing was already done for individual assignments in #1836 

## Motivation
This proposed change will increase transparency of the assignment setup process by letting students see their newly created GitHub repository (while the source import completes).

This will also allow students to report slow imports directly to GitHub support (if they experience a slow import) via the GitHub.com user experience.


srinjoym:
@BenEmdon Are the rerecorded cassettes a consequence of the change? Looks like this was just an UI addition

BenEmdon:
@srinjoym they are a result of the change in `app/jobs/group_assignment_repo/create_github_repository_job.rb` where I added a new `repo_url` field.

BenEmdon:
@srinjoym friendly nudge 😉

PR #1839, opened 2019-05-23
Use classroom token to register webhooks

BenEmdon:
## What
This PR is a WIP. Eventually Classroom will use its app token to register web-hooks instead of using an org admins token (which is less reliable).

This is a step towards solving #1647


BenEmdon:
I determined that you cant use an apps token to register webhooks, so this approach is no longer viable.

PR #1843, opened 2019-05-29
Link to issues page

BenEmdon:
## What ✨ 

Screen Shot 2019-05-29 at 10 50 41 AM

* Previously the link would take you page where you draft a new bug issue: https://github.com/education/classroom/issues/new?template=bug_report.md
* Now the link would take you to the page where you can see the issues board: https://github.com/education/classroom/issues

## Why 
By bringing future bug reporters to the issues board first, they will be more likely to see if the issue they are experiencing already exists. This will hopefully reduce the number of duplicate issues.

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -20,7 +20,7 @@
           </li>
 
           <li class="mx-3">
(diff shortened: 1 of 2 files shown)


srinjoym:
This is a great idea! ✨

PR #1845, opened 2019-05-30
Upgrade OmniAuth for CVE-2015-9284

tarebyte:
Got a security alert about this, and the security page said Dependabot couldn't fix it 😓


tarebyte:
Shit I totally missed that this didn't fix the patch, ok well we'll keep our eyes peeled.

Issue #1847, opened 2019-05-30
Remove the `webhook_id` and `is_webhook_active` column from the Organization model


BenEmdon:
### Feature request :sparkles:
Now that we have moved to using the `OrganizationWebhook` model as a source of truth for the webhook ID, we should remove the unused webhook_id column from `Organization`.

This should clean up some tech debt (and duplication).

## TODO

- [ ] Remove all usage of `Organization`'s `webhook_id`( #1850, #1848)
- [ ] Remove all usage of `Organization`'s `is_webhook_enabled`  (#1848)
- [ ] Create and run a migration to remove the columns (#1853)

PR #1850, opened 2019-05-30
Remove usage of Organization's `webhook_id`

BenEmdon:
## What
This PR removes all usage of the `Organization` model's `webhook_id` field, in favour of using the `OrganizationWebhook` model's `github_id`.

This PR also removes `lib/tasks/organization_webhook_backfill.rake` since the backfill was ran and completed a while ago, and is no longer needed.

## Why 
This is part of #1847


db/schema.rb
```diff
@@ -224,6 +224,13 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "ticker_stats", force: :cascade do |t|
```
srinjoym: I think this sneaked into the PR, this is the same way I committed it last time 😆 If you reset your DB based on the schema without this ticker stats table it should stop doing this I think

PR #1862, opened 2019-06-11
Fix setup failures

thebradbain:
## What
~ We the interns ~ encountered a few shared snaggles with the `script/setup` bootstrapping, and these changes slightly modified the setup script so that future setups will be more painless for future contributors (and interns!):

1. Moved `rbenv` and `nodenv` dependency checking to the beginning of `script/boostrap`. The same checks are still being done, but now they're being done first -- because some of the steps that were previously done before it would fail if these tools weren't already installed anyways.
2. Ensure `github/bootstrap` is tapped by homebrew, because `bootstrap-rbenv-ruby` and `bootstrap-nodenv-node` fail initially. It turns out this was because our homebrew installations were not tapped into [github/homebrew-bootstrap](https://github.com/github/homebrew-bootstrap). 
3. ~Unpin version 9.4 of postgresql within the Brewfile. It turns out this version is wholly incompatible with the latest Mac operating system, so as of now we are currently installing the most recent release of postgresql locally. If there's a reason we are still on version 9.4, feel free to revert this.~ See discussion below.

Additionally, a few of us ran into an issue with an incompatible version (< 2.0) of `bundler` since our `Gemfile.lock` was bundled with bundler 2.0.1. This could be fixed by manually running `gem install bundler` to update, but should we be looking for a better way? 

Finally, I tested this setup on as clean an environment I could (uninstalling everything initially installed by the unmodified script and running this version) and everything set up much more smoothly, but as it was a not clean slate Mac I'm wondering if that's enough.


script/bootstrap
```diff
@@ -7,11 +7,33 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+# Make sure the system has both rbenv 
+# and nodenv installed previously, otherwise we can't
+# ensure dependencies install correctly.
+if ! [ -f "$(which rbenv)" ]; then
+  echo "It looks like you don't have rbenv installed"
+  echo "Check out https://github.com/rbenv/rbenv#installation to get it running on your system"
```
BenEmdon: Nicely done ✨

EricPickup:
We all ran into these issues today when setting up classroom locally. Thanks for doing this! 🎉

Brewfile
```diff
@@ -7,6 +7,6 @@ cask "docker"
 cask "ngrok"
 
 brew "nginx", restart_service: :changed
-brew "postgresql@9.4"
+brew "postgresql"
```
BenEmdon: Looks like we are using Postgres `9.4.21` in production. Might be best if we can avoid changing this 😕 @d12 what do you think?

BenEmdon: Or maybe it's time to bump the production version?

thebradbain: Yeah I don't like unpinning the local version to be separate from the deployed version, and postgresql 9.4 will still install on each of our machines, but it seems the `pg` gem will not install with that version of psql properly on OSX Mojave, at least according to our experiences today... looking online to see if that's by design or we all just got unlucky in some way. 

In any event, postgresql 12 is about to release, so there's maybe that to consider too.

d12: Weird... Yeah if there's a way to make this work without unpinning the version, that'd be best. Let me know what you find.

thebradbain: Hello! It turns out it wasn't a compatibility issue between operating systems, but a subtlety (read: annoyance) in how Homebrew now manages pinned packages: when you pin a version, as we are with postgresql, Homebrew installs it as "keg-only" meaning it is not linked anywhere systemwide (meaning it's not on PATH, etc) out of fear it could interfere/conflict with the latest installed version of that same application, and leaves that decision to the user.

I've changed the bootstrap script such that it will install and link postgresql@9.4 _only_ if there is no other version of postgresql detected. I can also put it in a y/n prompt if that's of interest. 

Now pg builds properly against version 9.4!

d12: Nice find :+1: This seems perfect for now, we can always iterate on the script in the future. 💖

PR #1864, opened 2019-06-12
Avoid loading of all assignments objects multiple times when transferring assignments to new user.

shaunakpp:
## What

In `OrganizationEventJob` whenever we transfer assignments from one user to other, in order to check the creator we load all the assignments. This PR is a small optimization to not load these objects and use the `.count` query for the same purpose. It also updates the transfer assignments to first select all assignments of the user, and then update them instead of iterating over all assignments.

### DIFF
--- a/app/jobs/organization_event_job.rb
+++ b/app/jobs/organization_event_job.rb
@@ -29,11 +29,11 @@ class OrganizationEventJob < ApplicationJob
   end
 
   def user_owns_assignments?
-    @organization.assignments.where(creator_id: @user.id).count.positive?
+    @organization.assignments.where(creator_id: @user.id).any?
   end
 
   def user_owns_group_assignments?
-    @organization.group_assignments.where(creator_id: @user.id).positive?
+    @organization.group_assignments.where(creator_id: @user.id).any?
   end
 
   def transfer_assignments
--- a/app/jobs/organization_event_job.rb
+++ b/app/jobs/organization_event_job.rb
@@ -19,20 +19,28 @@ class OrganizationEventJob < ApplicationJob
 
     return true unless @user
 
-    transfer_assignments if user_owns_assignments?
+    transfer_assignments if user_owns_any_assignments?
     @organization.users.delete(@user)
   end
   # rubocop:enable Metrics/AbcSize
 
+  def user_owns_any_assignments?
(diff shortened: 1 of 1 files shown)


app/jobs/organization_event_job.rb
```diff
@@ -19,20 +19,28 @@ def perform(payload_body)
...
   end
   # rubocop:enable Metrics/AbcSize
 
+  def user_owns_any_assignments?
+    user_owns_assignments? || user_owns_group_assignments?
+  end
+
   def user_owns_assignments?
-    @organization.all_assignments.map(&:creator_id).include? @user.id
+    @organization.assignments.where(creator_id: @user.id).count.positive?
```
d12: Can we do `@organization.assignments.where(creator_id: @user.id).any?` here? This skips counting all the records by applying a `LIMIT 1` to the query.

app/jobs/organization_event_job.rb
```diff
@@ -19,20 +19,28 @@ def perform(payload_body)
...
+  end
+
+  def user_owns_group_assignments?
+    @organization.group_assignments.where(creator_id: @user.id).positive?
```
d12: Similar comment here as my comment above.

app/jobs/organization_event_job.rb
```diff
@@ -19,20 +19,28 @@ def perform(payload_body)
...
   end
 
   def transfer_assignments
```
d12: The stuff in this method was pretty rough, nice refactor. This looks much better!

app/jobs/organization_event_job.rb
d12: A quick test in Rails:

```ruby
  irb(main):003:0> User.all.any?
  User Exists (0.4ms)  SELECT  1 AS one FROM "users" LIMIT ?  [["LIMIT", 1]]
```

shaunakpp:
Hi @BenEmdon existing tests are showing full coverage in `SimpleCov`, should I still add tests for the new methods?

shaunakpp:
I realized that there's a slightly larger refactor that involves `OrganizationEventJob` and `OrganizationsController`. So I've pushed it to a new branch and here's the comparison: https://github.com/education/classroom/compare/transfer_assignment_service?expand=1

If y'all are okay with the new changes I'll close this PR and open that.

PR #1867, opened 2019-06-12
Add new service to transfer assignments, removes duplicate code

shaunakpp:
## What

We had duplicate code for transferring assignments from one admin to other. So, in this PR I have created a service `TransferAssignmentsService` which does the transfer process and is called instead of the duplicate code.

P.S. I discovered this when working on #1864 and hence closing the original PR in favor of this.

### DIFF
--- a/app/services/transfer_assignments_service.rb
+++ b/app/services/transfer_assignments_service.rb
@@ -10,7 +10,7 @@ class TransferAssignmentsService
 
   def transfer
     return false unless user_owns_any_assignments?
-    all_assignments_of_user = organization.all_assignments.select do |assignment| 
+    all_assignments_of_user = organization.all_assignments.select do |assignment|
       assignment.creator_id == old_user.id
     end
     all_assignments_of_user.map do |assignment|
@@ -19,6 +19,7 @@ class TransferAssignmentsService
   end
 
   private
+
   def user_owns_any_assignments?
     user_owns_assignments? || user_owns_group_assignments?
   end
@@ -30,5 +31,4 @@ class TransferAssignmentsService
   def user_owns_group_assignments?
     organization.group_assignments.where(creator_id: old_user.id).any?
   end
-
 end
--- a/app/services/transfer_assignments_service.rb
+++ b/app/services/transfer_assignments_service.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class TransferAssignmentsService
+  attr_accessor :organization, :old_user, :new_user
+  def initialize(organization, old_user, new_user = nil)
+    @organization = organization
+    @old_user = old_user
+    @new_user = new_user || @organization.users.where.not(id: old_user.id).first
+  end
+
+  def transfer
+    return false unless user_owns_any_assignments?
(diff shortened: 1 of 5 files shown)


app/services/transfer_assignments_service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class TransferAssignmentsService
+  attr_accessor :organization, :old_user, :new_user
```
d12: Can we add a newline after this, to be more consistent with the app?

app/services/transfer_assignments_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    @old_user = old_user
+    @new_user = new_user || @organization.users.where.not(id: old_user.id).first
+  end
+
+  def transfer
+    return false unless user_owns_any_assignments?
+    all_assignments_of_user = organization.all_assignments.select do |assignment|
+      assignment.creator_id == old_user.id
+    end
+    all_assignments_of_user.map do |assignment|
```
d12: I'd group this all together to be a little more concise:

```ruby
organization.all_assignments.map do |assignment|
  next unless assignment.creator_id == old_user.id
  assignment.update(creator_id: new_user.id)
end
```

app/services/transfer_assignments_service.rb
shaunakpp: done 👍🏼

app/services/transfer_assignments_service.rb
shaunakpp: done 👍🏼

spec/services/transfer_assignments_service_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    it "is expected to transfer assignments to specified new user" do
+      assignment = create(:assignment, organization: organization, creator: old_user)
+      TransferAssignmentsService.new(organization, old_user, new_user).transfer
+      assignment.reload
+      expect(assignment.creator_id).to eq(new_user.id)
+    end
+  end
+  context "when teacher owns group assignments" do
```
d12: Can we add a newline above here too?

spec/services/transfer_assignments_service_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+describe TransferAssignmentsService do
+  let(:organization) { classroom_org }
+  let(:old_user) { classroom_teacher }
+  let(:new_user) { create(:user) }
+  before(:each) do
+    organization.users << new_user
+    organization.save
+  end
+  context "when teacher owns no assignments" do
```
d12: Can we add a newline above here too?

spec/services/transfer_assignments_service_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe TransferAssignmentsService do
+  let(:organization) { classroom_org }
+  let(:old_user) { classroom_teacher }
+  let(:new_user) { create(:user) }
+  before(:each) do
```
d12: Can we add a newline above here too?

PR #1865, opened 2019-06-12
Check for invalid GroupAssignmentRepos

EricPickup:
## What

Fixes #1505.

#1346 was in production for a while and resulted in broken/invalid `GroupAssignmentRepo`s. These models had their associating `Group` model destroyed which eliminated the associating `github_team` for the repos.

This change makes us check for a nil `Group` and returns a `NullGitHubTeam` when requesting the repo's `github_team`, which will appear as a ghost team as shown in the screenshot below.

**Before:**

image

**After:**

image

PR #1868, opened 2019-06-12
Fix GroupAssignmentRepo tests/VCR

EricPickup:
Thanks to @d12 for pairing with me on this issue.

## Background

I was trying to create tests for another PR (#1865) and ran into issues with VCR when trying to add new tests to the `GroupAssignmentRepo` spec.

We were loading `Group`s from a [factory](https://github.com/education/classroom/blob/master/spec/factories.rb#L62) that generated a random team name and ID that didn't exist and was causing problems when I tried to make a new test and VCR was trying to call the API using the fake team ID's. The existing tests still work because we used cassettes.

## Solution

We're now creating a GitHub Team through the API and using that real team ID to run the tests. I've regenerated all the old cassettes using this new method, and new tests can now be created without issues.


EricPickup:
Thanks for catching that @d12, just fixed it on my machine and rebased it on this PR.

PR #1874, opened 2019-06-14
Sort assignment repos

### LINKED ISSUE #847: Sort assignment repositories

thebradbain:
_NOTE_ See #1889 for the latest discussion

## What
Closes #1182 and closes #847, incorporates the suggested UI improvements, necessary tests, and applies to both Group and Individual Assignments. 

Teachers are now able to sort the assignment list via roster identifier or time created at for Individual Assignments, and team name and time created at for Group Assignments. Additionally, functionality is shared as much as possible between `AssignmentsRepo` and `GroupAssignmentsRepo` by being abstracted into the `Sortable` model concern/mixin, so adding a new sort filter is as simple as writing one line in the relevant model. 

Screen Shot 2019-06-14 at 1 00 41 PM

cc @stephaniegiang, as we probably want to make sure the assignment search filtering she's implementing is compatible with this before merge


thebradbain:
vcr / cassettes ~ hate ~ me today... i'll figure out the issue on CI and re-push

thebradbain:
See #1889 for further discussion

d12:
Can we close this PR in favor of #1889?

PR #1875, opened 2019-06-14
Common code shifting in AssignmentRepo and GroupAssignmentRepo models

shaunakpp:
## What
This PR does the following things:

- Moves common code from `AssignmentRepo` and `GroupAssignmentRepo` to a mixin `AssignmentRepoable`
- removes `disabled?` method from the above models and move it their View models(which was a long time #TODO )


spec/models/assignment_repo_spec.rb
```diff
@@ -53,6 +53,12 @@
 
       expect(repo.valid?).to be_truthy
     end
+    it "expects github_team_id to be nil" do
```
d12: Can we add a newline above here? To be consistent with other code in this app, blocks should not touch each other.

PR #1876, opened 2019-06-14
Add ability to set maximum amount of teams in group assignments

### LINKED ISSUE #1871: Maximum number of groups
_Please provide a description of the feature you would like Classroom to have_
I would like to be able to limit the maximum number of groups that could be created for an assignment. This is very useful to avoid students creating lots of individual groups in a group assignment and force tham to join already created groups.

EricPickup:
## What
Closes #1871.

This PR adds the ability to set "Maximum number of teams" in your group assignments.

**Group assignment form:**
Screen Shot 2019-06-14 at 2 57 21 PM

**Group invitation view when max teams have been reached:** (team name field and create buttons are disabled)
image

**Attempting to create a team when limit is exceeded**
Normally the team name and create button are disabled when the limit is reached. The case below could only happen if a student loads the invitation page _before_ the limit was reached.


d12:
@EricPickup I'm not sure if you've explored Groupings yet, they allow teachers to reuse sets of teams in many assignments. How does this work if I were to select a grouping that has more teams than the setting permits?

EricPickup:
@d12 Good to know, I wasn't too sure what Groupings were before.

Since the groups are already created in the grouping, the limit wouldn't be checked (it's only checked when _creating_ the group). We could go about this a few ways:

1. Check if the chosen grouping exceeds the team limit when submitting the group assignment form and flash an error if it violates it (could also display the number of teams in the grouping on the flash error so teachers could easily adjust their limit in the form)
2. Since the teacher definitely wants the same groups as the other assignment (using the grouping), ignore the limit. We're currently unintentionally doing that.
3. If the number of groups in the chosen grouping exceeds the team limit, automatically adjust the team limit to that amount for them.

What do you think?

d12:
@EricPickup I like option 1, I don't think there's any reason a teacher couldn't just bump the number up (or not use a number). Option 2 could mislead teachers if they accidentally select a grouping with more groups than the maximum they've specified.

EricPickup:
@d12 I went ahead and added a validator to `GroupAssignment` that considers groupings and checks the grouping sizes and existing team sizes when adding a `max_teams` limit.

Selected grouping has more groups than the `max_teams` limit when creating a new assignment:
image

Adding a `max_teams` limit to an existing assignment that contains more teams than the limit:
image


On a side note you can see there's a style issue with that first screenshot. The error box is covering the next field's label. I'll create an issue for that.

EricPickup:
@d12 Migration is in production, this PR should be ready to go! 🎉

PR #1877, opened 2019-06-14
Use roster identifier instead of GitHub username

### LINKED ISSUE #1811: Use class roaster identifier rather than github account
_Please provide a description of the feature you would like Classroom to have_

As a teacher, I use the class roaster identifier to find, grade and even email students. I'd like to use the same identifier when I download repositories from github rather than having to convert them from github usernames.

stephaniegiang:
## What
Uses the student's identifier in the roster to identify students. The link on the roster identity will still link to the student's GitHub account. This only changes the username for organizations that use rosters. 

### Before
image

### After 
image

Not sure if there are any tests I need to write for this. 

#### This PR is in prep for being able to search and filter students
cc @thebradbain


d12:
Hey @stephaniegiang , could you add screenshots to show before/after?

stephaniegiang:
@d12 Added the screenshots to the PR 😄

stephaniegiang:
@d12 I added tests to check the views 😸

app/views/orgs/roster_entries/assignment_repos/_linked_not_accepted.html.erb
```diff
@@ -8,7 +8,11 @@
     <div class='d-flex flex-column flex-justify-around'>
       <div>
         <%= link_to current_roster_entry.user.github_user.html_url, class: "assignment-repo-github-url", target: '_blank' do %>
-          <h3 class="css-truncate css-truncate-target"><%= current_roster_entry.user.github_user.login %></h3>
+          <% if current_roster_entry.identifier %>
+          <h3 class="css-truncate css-truncate-target"><%= current_roster_entry.identifier %></h3>
```
d12: This line should be indented 1 more!

app/views/orgs/roster_entries/assignment_repos/_linked_accepted.html.erb
```diff
@@ -10,7 +10,11 @@
     <div class='d-flex flex-column flex-justify-around'>
       <div>
         <%= link_to view.github_user_url, class: "assignment-repo-github-url", target: '_blank' do %>
-          <h3 class="css-truncate css-truncate-target"><%= view.github_user_login %></h3>
+          <% if defined? current_roster_entry %>
```
d12: Can we bring this logic out into the view model? The method could be something like `view.user_display_name` or something similar.

stephaniegiang: I looked into it and it's going to be a bit tricky to do that. Because `current_roster_entry` is not always defined and it's going to be hard to pass that to the view model. We get an error that `current_roster_entry` is not defined. The error happens at the view level. So we will still need an `if` statement.

app/views/orgs/roster_entries/assignment_repos/_linked_not_accepted.html.erb
```diff
@@ -8,7 +8,11 @@
     <div class='d-flex flex-column flex-justify-around'>
       <div>
         <%= link_to current_roster_entry.user.github_user.html_url, class: "assignment-repo-github-url", target: '_blank' do %>
-          <h3 class="css-truncate css-truncate-target"><%= current_roster_entry.user.github_user.login %></h3>
+          <% if current_roster_entry.identifier %>
```
d12: Are there any roster entries without identifiers? Or can we :fire: this line? This view partial is specific to students linked roster entries, so I _think_ we're good.

app/views/orgs/roster_entries/assignment_repos/_linked_accepted.html.erb
d12: Oh I see. This is fine :+1:

BenEmdon: Why are you using `if defined?` instead of `blank?` or `present?`? Does the controller not define it sometimes?

spec/controllers/assignments_controller_spec.rb
```diff
@@ -204,6 +204,41 @@
...
 
+  describe "display student username", :vcr, type: :view do
+    before do
+      organization.users.push(create(:user, uid: 90, token: "asdfsad4333"))
+      organization.roster = create(:roster)
+      organization.roster.roster_entries.push(RosterEntry.create(
+                                                identifier: "student",
+                                                roster: organization.roster,
+                                                user_id: organization.users.last
+      ))
```
BenEmdon: Nit-pick: please use indented style as described in https://github.com/education/classroom/pull/1473

For example:
```diff
-      organization.roster.roster_entries.push(RosterEntry.create(
-	                                                identifier: "student",
-	                                                roster: organization.roster,
-	                                                user_id: organization.users.last
-	      ))
+      organization.roster.roster_entries
+        .push(RosterEntry.create(
+          identifier: "student",
+          roster: organization.roster,
+          user_id: organization.users.last
+        ))
```

Also it would be cool if you could make your editor use 2 spaces in place of tabs 😅

app/views/orgs/roster_entries/assignment_repos/_linked_accepted.html.erb
stephaniegiang: Opps I didn't see this. I replied to a similar question. The controller does not always define it.

spec/controllers/assignments_controller_spec.rb
stephaniegiang: I'll make a new PR for that.

PR #1878, opened 2019-06-15
Move assignment_repos relation helper from controller to RosterEntry model

thebradbain:
## What
Currently, `RosterEntriesController` has some helper methods regarding navigating model relationships in it that could benefit from being moved to the model level rather than being kept purely within the controller. One of these best candidates is the `assignment_repo` controller helper method.

Now, given an instance of a `RosterEntry`, its `assignment_repos` may be easily accessed anywhere, not just from within `RosterEntriesController`.

### DIFF
--- a/app/controllers/orgs/roster_entries_controller.rb
+++ b/app/controllers/orgs/roster_entries_controller.rb
@@ -61,7 +61,7 @@ module Orgs
     end
 
     def find_assignment_repo
-      current_assignment.repos.find_by(user: current_roster_entry.user)
+      current_roster_entry.assignment_repos.find_by(assignment: current_assignment)
     end
 
     def find_current_assignment_repo
--- a/app/controllers/orgs/roster_entries_controller.rb
+++ b/app/controllers/orgs/roster_entries_controller.rb
@@ -61,7 +61,9 @@ module Orgs
     end
 
     def find_assignment_repo
-      current_roster_entry.assignment_repos.find_by(assignment: current_assignment)
+      current_roster_entry
+        .assignment_repos(organization_id: current_assignment.organization_id)
+        .find_by(assignment: current_assignment)
     end
 
     def find_current_assignment_repo
--- a/app/models/roster_entry.rb
+++ b/app/models/roster_entry.rb
@@ -5,6 +5,8 @@ class RosterEntry < ApplicationRecord
   belongs_to :roster
   belongs_to :user, optional: true
 
+  has_many :assignment_repos, primary_key: :user_id, foreign_key: :user_id
+
   validates :identifier, presence: true
   validates :roster,     presence: true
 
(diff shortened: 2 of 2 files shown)


d12:
The actual diff here won't introduce any errors, but the association is misleading and technically not correct. A roster entry can `has many` assignment_repos, but they must be within the same Classroom. By using the user_id as the mapping, we'll have assignment_repos returned from other classrooms since 1 user can be mapped to many roster entries via multiple classrooms.

As long as every call site remembers to do`.find_by(assignment: assignment)` or `.find_by(organization: organization) we'd be fine, but we should avoid needing to rely on memory or context to use an association or method correctly.

Instead, can we rewrite this association to return _only_ assignment repos within the current classroom? We'll still need to filter by assignment in the controller, but then at least the relation is correct and we can use it in the future without unexpectedly returned records.

thebradbain:
All of that sounds 💯 . The problem I'm running into, however, is creating a valid `has_many` ActiveRecord association with the ability to pass in completely external context. We can _almost_ do something like: 

```
 has_many :assignment_repos, 
            ->(entry, organization_id) { joins(:assignments).where( assignments: {organization_id: organization_id} ) },
            primary_key: :user_id, 
            foreign_key: :user_id
```

But it appears the argument passed will always be the instance of the ActiveRecord being called (in this case RosterEntry) with no ability to pass extra arguments, in order to be compliant with how Rails normally handles `has_many` relationships. 

As such, I don't believe passing in external context is not possible from a purely standard association standpoint. However, I could create a custom `assignment_repos` model instance method which accomplishes a similar thing, but will unfortunately be unable to be used in `'join` requests for more complex queries. Is this a desirable solution, or do you think it would just cause confusion?

d12:
Oh I was wrong! A roster entry belongs to a roster which can belong to many organizations, which means roster entry repos shouldn't be scoped to a single classroom, but to all the classrooms the roster is part of. How confusing 😱 

That being said, I think the way we were doing this before was actually more simple 😅 Figuring out all the repos for a given entry is hard, but figuring out the repo for an entry and an assignment is pretty easy, we don't need any complicated logic or new associations for that. Since that's all we need in the controller, what do you think about writing a `for_assignment_and_roster_entry` method on `AssignmentRepo`? Then we can move the logic to the model without introducing unnecessary complexity.

thebradbain:
I totally agree for our current use case, especially after the latest commit. 

However, for the _other_ feature I'm working on regarding sorting AssignmentRepos associated with RosterEntries within the assignment view, it would be nice to have such an association because then for example a filter to sort repositories attached to a roster_entry could be a scope which looks like 

`left_outer_joins(:assignment_repos).where(assignment: @assignment).order("assignment_repos.created_at")`

As doing something like `@assignment.repos.where(user_id: roster_entry.user).order(:created_at)` would give us back a list of assignment_repos, not actually roster_entries, and I don't know of anything to nicely convert that Assignment-to-AssignmentRepos relation to AssignmentRepos-to-RosterEntry.

PR #1880, opened 2019-06-18
Migration for max_teams in group assignments

EricPickup:
## What
This migration is required for #1876. It adds the `max_teams` column to `GroupAssignments`, allowing teachers to set the maximum amount of teams allowed in a group assignment.


d12:
Hm CI is complaining because the version number of your migration is different than the version number in schema.rb. This will be :+1: once that's fixed up.

PR #1883, opened 2019-06-18
Fix spelling mistake in assignment view

EricPickup:
## What
"assignment" has been misspelled as "assigment" on the assignment page for 2 years 🤔


stephaniegiang:
Maybe we should put that sentence in the `config/locales/en.yml` file

EricPickup:
@stephaniegiang this view has a lot of hardcoded strings already, I think that'd belong in a separate PR where we can internationalize the entire view

PR #1879, opened 2019-06-17
Allow searching for individual and group assignments

stephaniegiang:
#### Will need to be merged after #1877

Closes #1873, closes #1300, and closes #847

Allows teachers to search for assignments based off of the roster identify. Also allows teachers to sort assignments as well.

cc @thebradbain as he is implementing a filter function that will work with the search as well

### Search in action
For group assignments
edited

For individual assignments
individualassignments

PR #1889, opened 2019-06-20
Merge assignment sort and search features

### LINKED ISSUE #847: Sort assignment repositories

### LINKED ISSUE #1300: Search submissions
It would be great if there was a way to search the submissions.
Let's say I want to find the repository that corresponds to homework 1 of a particular student.
There's no way to do this, and doing it manually by scrolling to the paginated assignment submissions is not really feasible (the interface is sluggish and I have 8 pages).
Sorting by the identifier as suggested in #847 would help a bit (I could do manual binary search instead of linear search) but really there should be a search.

### LINKED ISSUE #1873: Search functionality for students within an assignment
Creating a feature that allows teachers to search for a students within an assignment.

thebradbain:
## What

Closes #1874, closes #1879, closes #1873, closes #1300, closes #1182, closes #847 🚀 

This Mega-PR (which definitely could have benefitted from being.... broken up into smaller PRs. Thank you for your patience, reviewers! 😬😅) merges the work done by both @stephaniegiang and I on group/individual assignment searching (#1879) and sorting (#1874), respectively. In addition to to ensuring the two features play nice together, this PR makes a few UI tweaks to put these features side-by-side, refactors/streamlines shared code and common functionality, and dynamically reloads the assignment list on both search and sort ordering so that a full page refresh is not necessary.


app/controllers/assignments_controller.rb
```diff
@@ -28,26 +30,47 @@ def create
...
+  def filter_repos
+    return unless @organization.roster
+
+    matching_roster_entries = @organization.roster.roster_entries
+    if @query.present?
+      matching_roster_entries = matching_roster_entries.where("identifier LIKE ?", "%#{@query}%")
+    end
+
+    @roster_entries = matching_roster_entries
+      .page(params[:students_page])
```
d12: How does this work when we're not on the first page? It'd be weird to be paginating by created_at, then apply a search and be given results on the nth page. Or what if we're on the second page, and the results from search don't fill a full page? Will this render an empty set?

We probably want to reset the page param when we perform a search.

app/models/roster_entry.rb
```diff
@@ -10,6 +13,27 @@ class IdentifierCreationError < StandardError; end
...
 
+  scope :order_by_repo_created_at, lambda { |context|
+    assignment = context[:assignment]
+    sql_formatted_assignment_id = assignment.id
+
+    order("assignment_repos.created_at")
+      .joins <<~SQL
+        LEFT OUTER JOIN assignment_repos
+        ON roster_entries.user_id = assignment_repos.user_id
+        AND assignment_repos.assignment_id=#{sql_formatted_assignment_id}
```
d12: Can you replace this string interpolation with SQLs `?` interpolation?

app/models/roster_entry.rb
```diff
@@ -49,7 +73,6 @@ def self.order_for_view(assignment)
         WHEN roster_entries.user_id IN #{sql_formatted_users} THEN 0 /* Accepted */
         ELSE 1                                                       /* Linked but not accepted */
       END
-      , id
```
d12: Why was this removed?

app/views/assignments/_assignment_roster_list.html.erb
```diff
@@ -0,0 +1,5 @@
+<% roster_entries.each do |entry| %>
```
d12: Can we feature flag the entire search/filter feature? That way, we can slowly roll it out and monitor performance and feedback, and it'll be easy to rollback if we need to.

app/views/assignments/_filterable_assignment_roster_list.html.erb
```diff
@@ -0,0 +1,30 @@
+
```
d12: Extra new line here!

app/assets/javascripts/select_menu.js
```diff
@@ -0,0 +1,61 @@
...
+
+  $(document).click(function(event) {
+    if(select_menu.find(event.target).length < 1) {
+      closeSelectMenu();
+    }
+  });
+};
+
+
+$(document).ready(initialize_select_menu)
```
d12: There's quite a few files in here with missing newlines at the end, could you fix that up? Your editor should have a setting to add them automatically :)

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

app/views/group_assignments/show.html.erb
```diff
@@ -57,16 +57,10 @@
...
-                <div class="assignment-repo-list">
-                  <% @group_assignment_repos.each do |group_assignment_repo| %>
-                    <%= render partial: 'orgs/group_assignment_repos/group_assignment_repo', \
-                      locals: { url: organization_group_assignment_group_assignment_repo_path(@organization, @group_assignment, group_assignment_repo) } %>
-                  <% end %>
-                </div>
-
+                <%= render partial: 'group_assignments/filterable_group_assignment_repo_list' %>
                 <%= render partial: 'shared/pagination', locals: { collection: @group_assignment_repos, param_name: :teams_page } %>
+
```
d12: Extra newline!

app/views/shared/_select_menu.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <% options.each_with_index do |option, i| %>
+              <% conditional_selected_class = "selected" if option == selected_option %>
+
+              <%= link_to option_links[i], class: "select-menu-item #{conditional_selected_class}", remote: remote do %>
+                <% show_or_hide_class = "v-hidden" if option != selected_option %>
+
+                <%= octicon 'check', class: "select-menu-item-icon #{show_or_hide_class}" %>
+                <span class="select-menu-item-text"><%= option %></span>
+              <% end %>
+
```
d12: Extra newline!

app/models/roster_entry.rb
thebradbain: Unfortunately multi-line SQL statements in this format `?` is not possible -- I followed the same format that the rest of this file does for raw SQL queries (of which some also use string interpolation), but I can change it if you like.

app/assets/javascripts/roster_tabs.js
```diff
@@ -3,6 +3,7 @@ function selectTab(element){
   unlinkedTab = document.getElementById('unlinked-tab');
   studentsSpan = document.getElementById('students-span');
   unlinkedSpan = document.getElementById('unlinked-span');
+  searchBar = document.getElementById('js-search-form');
```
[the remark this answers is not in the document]
stephaniegiang: In classroom right now we actually have an organization search bar that looks like this. The search bar for assignments looks identical.
image
We want to keep it consistent with the organization search bar for now. But we can add the magnifying glass on the side in the next PR.

app/assets/javascripts/select_menu.js
```diff
@@ -0,0 +1,61 @@
+function initialize_select_menu() {
```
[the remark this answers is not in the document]
stephaniegiang: The select menu has a JS so that it can be consistent with the search bar. The search bar only refreshes the partial and not the entire page. So we wanted to do the same with the search menu.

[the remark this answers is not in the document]
thebradbain: Yes, I agree it's not ideal, and am very open to a different approach if necessary (though I couldn't think of any that wouldn't require changing the select menu in some way). The reasoning behind that is that to update the select menu links with the latest search query, re-rendering the select menu is necessary, which causes all existing jquery event listeners to be destroyed, hence the need to re-initialize all the jquery hooks. 

Other options I've thought of: 

1. Combine the select menu and search bar into a single form: `shared/_select-menu.html` as it is does not lend itself well to integration within existing forms, and I'd be open to exploring changing it but I want to make sure that we don't specialize this component so much it can't be used as a component in the general case

2. Re-render just the links in the search menu -- would require functional changes to _search_menu.js and probably refactoring the actual list links html into a separate partial so it can be re-renderd separately via ERB. 

3. Re-render the entire container partial containing the search bar / sort filter every submit -- icky, would require we re-hookup event listeners to both search and sort filter

app/views/assignments/_assignment_roster_list.html.erb
stephaniegiang: Yes we can definitely do that

app/models/roster_entry.rb
thebradbain: Pre-PR `order_for_view` would order by accepted/not accepted/not linked _and_ within these three groupings sort by `id`. Due to the way ActiveRecord handles ordering, our `order_for_view.order_by_sort_mode` query would essentially be ordering by accepted/not accepted/not linked, then `id`, then _only if two entities has the same `id`__ order by sort mode, which would mean something would never get sorted by the sort mode. If you'd like, we can explicitly add `id` to the end of this order chain within the controller's full query?

app/views/assignments/_assignment_roster_list.html.erb
thebradbain: Yes! I was going to ask: how specifically do I do that. Is it just a simple method conditional?

app/assets/javascripts/roster_tabs.js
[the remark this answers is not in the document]
thebradbain: ^ We can definitely open up a separate PR for the search bar component though in general to add one, if one exists

BenEmdon:
This PR is greater than 500 lines and I don't think I'll have the bandwidth to review ATM. 
If you need a second pair of eyes feel free to ask me specific questions 👀

app/assets/javascripts/roster_tabs.js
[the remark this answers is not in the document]
d12: Ah, gotcha. This is fine then :+1:

app/models/roster_entry.rb
d12: Relaying from Slack, looks like we can't do ? interpolation into `joins` with AR. This is fine

app/models/roster_entry.rb
d12: > would essentially be ordering by accepted/not accepted/not linked, then id, then, and only if two entities has the same id, order by sort mode

Ah yeah, you're totally right 😄 

> If you'd like, we can explicitly add id to the end of this order chain within the controller's full query to ensure the ordering preference is accepted/not accepted/not linked, sort mode, then id?

This would be good. SQL doesn't guarantee ordering of sets when there is no order, or when there are ties like there will often be here. Adding a `id` order at the end means we won't get assignment repo lists shuffling on reload.

app/views/assignments/_assignment_roster_list.html.erb
d12: You can look into https://github.com/education/classroom/blob/8b09ce135f92eb6ea3753df104e68ec847dfdd95/app/controllers/application_controller/feature_flags_dependency.rb which defines our flipper helpers.

More info: https://github.com/jnunemaker/flipper

I believe rosters are behind a feature flag right now, you can also use those as an example.

thebradbain:
@BenEmdon No worries! Fortunately I believe most of those lines are due to cassettes and refactoring view partials to be more DRY, most of the 'real' changes are contained to `RosterEntry.rb`, `AssignmentsController.rb`, and `GroupAssignmentsController.rb`

app/assets/javascripts/select_menu.js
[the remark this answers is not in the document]
thebradbain: @d12 I should have also mentioned that we're using I believe(?) a primer.style select menu component similar to the rest of our components and not an actual native select menu, so that's why jquery is used in the first place to interact with it

app/controllers/assignments_controller.rb
thebradbain: You're right, good catch

app/views/assignments/_assignment_roster_list.html.erb
stephaniegiang: Just added the feature flag

app/views/assignments/_filterable_assignment_roster_list.html.erb
```diff
@@ -1,29 +1,35 @@
+<<<<<<< Updated upstream
 <div class='TableObject'>
+=======
+<% if search_assignments_enabled? %>
+  <div class='TableObject'>
+>>>>>>> Stashed changes
```
thebradbain: It looks like some merge conflicts from applying a stash still exist :joy:

stephaniegiang: 😞I'll fix it

app/controllers/assignments_controller.rb
```diff
@@ -55,7 +55,7 @@ def show
 
   # rubocop:disable MethodLength
   def filter_repos
-    return unless @organization.roster
+    return unless @organization.roster || search_assignments_enabled?
```
thebradbain: Do you think this should be `@organization.roster && search_assignments_enabled?` instead? Otherwise this feature will technically be available for any assignment with an organization

stephaniegiang: Good point 👍

app/controllers/assignments_controller.rb
```diff
@@ -28,26 +30,49 @@ def create
...
     @unlinked_user_repos = AssignmentRepo
       .where(assignment: @assignment, user: @unlinked_users)
       .order(:id)
       .page(params[:unlinked_accounts_page])
   end
+  # rubocop:enable AbcSize
+  # rubocop:enable MethodLength
+
+  # rubocop:disable MethodLength
+  def filter_repos
```
d12: As we discussed in Slack, lets :fire: this endpoint if possible. Instead of setting page(1) in the controller, that can happen in the front end on search.

app/controllers/group_assignments_controller.rb
```diff
@@ -43,6 +45,26 @@ def show
       .page(params[:students_page])
   end
 
+  # rubocop:disable MethodLength
+  def filter_repos
+    return unless search_assignments_enabled?
+    matching_groups = @group_assignment.grouping.groups
+    if @query.present?
+      matching_groups = matching_groups.where("slug LIKE ?", "%#{@query}%")
```
d12: Can we search by the team name and not the slug? Searching for `Team` or `A` with a team called `Team A` returns no results since the slug is `team-a`.

app/assets/javascripts/select_menu.js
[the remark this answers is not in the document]
thebradbain: ugh and here I was thinking I had combed through them all 😅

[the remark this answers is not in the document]
stephaniegiang: You can do `rubocop -a FILE_NAME` and it should automatically fix them

@EricPickup actually made a PR https://github.com/education/classroom/pull/1888 that will do that automatically run the line when running `script/test`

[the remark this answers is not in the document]
thebradbain: I'm hoping I've finally did it. Someone please reserver the right to yell at me if not

stephaniegiang:
@d12 can we get another review on this?

app/views/assignments/_filterable_assignment_roster_list.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+    </div>
+  </div>
+<% end %>
+
+<div class='assignment-repo-list' id='assignment-repo-list'>
+  <%= render partial: 'assignments/assignment_roster_list', locals: {
+    organization: @organization,
+    assignment: @assignment,
+     roster_entries: @roster_entries
```
d12: Small indentation issue here

app/views/shared/_select_menu.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  </button>
+  <div class="select-menu-modal-holder">
+    <div class="select-menu-modal">
+      <div class="select-menu-header" tabindex="-1">
+        <%= octicon 'x', class: "select-menu-close-button" %>
+        <span class="select-menu-title"><%= title %></span>
+      </div>
+      <div class="select-menu-list">
+        <% options.each_with_index do |option, i| %>
+              <% conditional_selected_class = "selected" if option == selected_option %>
```
d12: This is indented one too far

app/views/shared/_select_menu.html.erb
```diff
@@ -0,0 +1,28 @@
...
+
+              <%= link_to option_links[i], class: "select-menu-item #{conditional_selected_class}", remote: remote do %>
+                <% show_or_hide_class = "v-hidden" if option != selected_option %>
```
d12: We can use `unless` here instead of `if !=`

```suggestion
                <% show_or_hide_class = "v-hidden" unless option == selected_option %>
```

phannan:
Last term my list was in classid order.  Apparently now it is in first name / last name order.  I can see why some instructors might want this.  AND, I don't.  I prefer classid order so that my grading / grader can look at the repositories anonymously.     The order is important because I translate the graders feedback into an issue for the students and want to be able to bop down the list in the same order that it was graded.  Please allow for choice on the sort order.  Class ID / github login / first name - last name all make sense based on your context.

d12:
@phannan :wave:

Thanks for the feedback. I have a couple of quick questions,

1) Does this classroom you're looking at use a Roster?
2) What do you mean by `classid`?

I'm sure we can add a sort option to match what the behavior was before, I'm mostly just unsure what `classid` refers to.

phannan:
Yes I use a roster.  When you download the roster, the field that I am talking about is labeled "identifier".   It is the one they have to choose when they come in the classroom for the first time.  For us its an anonymous identifier that we then use to refer to a specific student.

stephaniegiang:
@phannan Could you please provide your classroom name? 

We would love to take a look at how you are setting up your class.

phannan:
cst136-su19-oit-pdx

d12:
Are you looking to sort your assignment submissions by roster identifier? I believe the default is `created_at` right now (which I believe we should change to the roster identifier), but there is an option to sort by identifier in the drop down beside the search box.

Screen Shot 2019-07-16 at 4 37 01 PM

phannan:
Yes Yes Yes!
Thank you.  I didn't see the option on the right.  Student ID would be a better default to me, and this solves my problem.

d12:
Excellent! Thank you for the feedback, we'll have that default fixed very soon!

You can follow progress on that work here: https://github.com/education/classroom/issues/2019

d12:
@phannan This should be resolved now :)

PR #1890, opened 2019-06-21
Optimize js-search-form

thebradbain:
## What
Fixes some subtle bugs with the (relatively old) front-end search component. The way it previously responded to `keyup` events unintentionally created duplicate event handlers, which ultimately meant that identical `"ajax:success"` and similar handlers would get run multiple times per request, which eventually creates a noticeable, unnecessary lag for the user.


app/assets/javascripts/search.js
```diff
@@ -1,11 +1,13 @@
...
-  var delay, ready;
+  var debounce, ready;
 
-  delay = (function() {
+  debounce = (function() {
     var timer;
     timer = 0;
     return function(callback, ms) {
-      clearTimeout(timer);
+      if(timer) {
```
BenEmdon: Nitpick: add extra spacing after `if`
```suggestion
      if (timer) {
```

app/assets/javascripts/search.js
```diff
@@ -16,12 +18,22 @@
...
-        $this.submit();
-        return $this.on('ajax:success', function(e, data, status, xhr) {
+
+      debounce(function() { 
+        $this.one('ajax:beforeSend', function(e, data, status, xhr) {
+          $this.find(":input").prop("disabled", true);
+        });
+
+        $this.one('ajax:complete', function(e, data, status, xhr) {
+          $this.find(":input").prop("disabled", false);
```
BenEmdon: How come you are using `"` in some places but `'` in others? I don't know JS very well but are they interchangeable like this? And if they are interchangeable, should we strive for consistency?

thebradbain: Yes, they're interchangeable. And you're so right, I usually have my editor settings turned on to ensure consistency but it seems I forgot to enable that for my work computer. Changes incoming

PR #1892, opened 2019-06-21
Better error reporting on individual assignment creation

### LINKED ISSUE #1684: Better error reporting on Assignment Creation flow
It was difficult to determine the cause of #1575 because the error was obscured by the following non descriptive error strings:

https://github.com/education/classroom/blob/49864b3246ccc4d48829ddb6276a7bf98ebb2253/app/models/assignment_repo/creator.rb#L5-L8

Instead we should use errors straight from GitHub like we did (before moving it to the creator pattern) `GroupAssignmentRepo`s:
https://github.com/education/classroom/blob/49864b3246ccc4d48829ddb6276a7bf98ebb2253/app/jobs/group_assignment_repo/create_github_repository_job.rb#L41-L42

- [x] Make errors better for `AssignmentRepo`
- [ ] Make errors better for `GroupAssignmentRepo`

EricPickup:
## What

Part of #1684.

This PR adds the detailed GitHub error to the user-friendly (but very uninformative) error messages we're reporting when creating individual assignments.

image


d12:
@EricPickup You can use Draft PRs to open a PR without pinging codeowners :) I'll unassign myself for now, please re-ping when this is ready for review!

BenEmdon:
Neato ✨ Cant wait for this 🙌

Issue #1894, opened 2019-06-24
Upgrading our instance of PostgreSQL


tarebyte:
Currently, our Heroku instance of PostgreSQL is running at version `9.4.21` 😅 

This _is_ a supported version, but it appears that 9.4 [will stop being supported in early 2020](https://www.postgresql.org/support/versioning/).

While it's the summer time we should look into what it would take to upgrade Postgres to the latest version.

Here are the [Heroku docs](https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases#upgrading-with-pg-upgrade) about upgrading an instance.

Is someone interested in taking this on?

I'm happy to 🍐 on it.

thebradbain:
I’d be happy to take this up! 🚀

thebradbain:
_Note:_ We will be upgrading to PostgresQL v10.9, as that is the latest version currently supported by Travis CI. v10.9 will be supported until [Nov 10, 2022](https://www.postgresql.org/support/versioning/)

Upgrade is planned for **7:30 PM PST on Sunday, 06/30/2019**

**Pre-Upgrade Checklist**:

- [ ] Announce planned upgrade on GitHub Classroom website
- [ ] Upgrade `pg` gem from v0.21.0 to 1.1.4
- [ ] Configure CI to test against both current (9.4) and planned (10.9) versions
- [ ] Update local/dev installation db dependencies to keep in sync with production (edited) 

And _after_ all of that is done, at the time of the upgrade:
- [ ] Turn on heroku maintenance mode and upgrade the database https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases#upgrading-with-pg-upgrade

shaunakpp:
Attaching the links for release notes on the new postgres version(11.4) and also the migration changes. I don't think we need to make any changes for our instance, but it might come in handy for quick reference.
Release notes:
https://www.postgresql.org/docs/11/release-11-4.html

Migration reference:
https://www.postgresql.org/docs/11/release-11-1.html

thebradbain:
@tarebyte Since Classroom seems to be in a more stable place now, would you like to 🍐 on the actual db upgrade again sometime after you're back from summit?

tarebyte:
Sure thing! Grab a time on my calendar.

thebradbain:
Woo! Now that we've finally upgraded the db in production, I've created two followup issues re: tying up loose ends.

- [ ] Update pinned development environment version of postgresql to ensure consistency (#2058)
[...]

PR #1900, opened 2019-06-26
Configure CI to test against multiple Postgres database versions

thebradbain:
## What

Part of #1894 

In ensuring the planned postgresql upgrade goes as smooth as possible, configure CI to test against both the current (v9.4) and planned (v10.9) postgresql versions.

Note: Travis does not support this capability out of the box, but they do provide [documentation](https://docs.travis-ci.com/user/database-setup/#multiple-database-builds) to do so. Nevertheless, after the upgrade, we will probably want to revert these changes, as it doesn’t make sense to test against two versions after the fact.


tarebyte:
@thebradbain should we update the `pg` gem as well in this PR?

Awesome work btw ✨

BenEmdon:
This :clap: is :clap: SO 👏 dope 👏

thebradbain:
I was thinking upgrading the `pg` gem be a separate PR, as the current version is still technically compatible with the latest version of Postgres, it's just old. I can add it in this one though, if we feel they're related enough! @tarebyte

tarebyte:
@thebradbain I'm fine with it being a separate PR.

PR #1904, opened 2019-06-26
Make search not case sensitive

stephaniegiang:
## What
There is a bug with the organization search and assignment search as well. With the current query, the search is case sensitive. This PR removes the case sensitivity by changing `LIKE` to `ILIKE`.

Also, there were no tests for the organization search as well. So I added them 😄


app/controllers/organizations_controller.rb
```diff
@@ -103,11 +103,12 @@ def setup_organization
   end
 
   def search
-    orgs_found = current_user.organizations.order(:id).where("title LIKE ?", "%#{params[:query]}%")
+    orgs_found = current_user.organizations.order(:id).where("title ILIKE ?", "%#{params[:query]}%")
+    @organization = orgs_found.page(params[:page]).per(12)
     respond_to do |format|
       format.html do
         render partial: "organizations/organization_card_layout",
-               locals: { organizations: orgs_found.page(params[:page]).per(12) }
```
thebradbain: Is there a reason we're removing the pagination in the response?

stephaniegiang: I moved it into a variable so we can grab it in the tests

stephaniegiang:
@d12 made the changes!

app/controllers/organizations_controller.rb
```diff
@@ -102,15 +102,23 @@ def setup_organization
     end
   end
 
+  # rubocop:disable MethodLength
   def search
-    orgs_found = current_user.organizations.order(:id).where("title LIKE ?", "%#{params[:query]}%")
+    @organization = current_user
```
d12: Could you rename this to `@organizations`? That comment might have been lost with the refactoring comment :)

PR #1905, opened 2019-06-26
Assignment Search Component UX improvements

thebradbain:
## What
The way we were previously approaching the client-side interop between the sort menu and search filter bar within the assignment and group assignment views was unideal and relied upon re-rendering the select menu component with the latest `query` parameter every time a search was performed. It worked, but it also created a sub-par experience for the user.

This PR changes that, so search is now not at all "jumpy" in the slightest, and has the added benefit of decoupling client-side concerns from the server. 

Also, as a bonus, whenever a search returns empty, instead of rendering nothing we render a message saying there were no matching repos in the results list.


app/views/assignments/_assignment_roster_list.html.erb
```diff
@@ -1,5 +1,11 @@
-<% roster_entries.each do |entry| %>
+<% if roster_entries.present? %>
+  <% roster_entries.each do |entry| %>
     <%= render partial: 'orgs/assignment_repos/assignment_repo', locals: {
       url: organization_assignment_roster_entry_path(organization, assignment, entry)
     } %>
+  <% end %>
+<% else %>
+  <div class="blankslate">
+    <h3>"<%= assignment.title %>" does not have any repositories matching this query</h3>
```
d12: 🎉

app/views/group_assignments/show.html.erb
```diff
@@ -89,7 +88,6 @@
     <% else %>
       <% if @group_assignment_repos.present? %>
         <%= render partial: 'group_assignments/filterable_group_assignment_repo_list', locals: { param_name: :page } %>
-        <%= render partial: 'shared/pagination', locals: { collection: @group_assignment_repos } %>
```
d12: Did we mean to remove this? It looks like this removed pagination from this view.

thebradbain: Yup, in fact this should never have actually been here, because it's included in the `group_assignments/filterable_group_assignment_repo_list` partial, where it's re-rendered accordingly as the results returned from the search query change

PR #1908, opened 2019-06-27
Let AssignmentRepo#create_github_repository_job to use the #perform method form AssignmentRepo::Creator

### LINKED ISSUE #1886: When creating AssignmentRepo and GroupAssignmentRepo, use the #perform method
Currently, when we create new `AssignmentRepo` and `GroupAssignmentRepo` from the `CreateGithubRepositoryJob` we do not use the `perform` method. Instead we have repeated code. There are broadcast messages sent in between which are not in the `perform` method but apart from that the code is essentially the same. We can move these messages to the creator and use perform method in the job.

shaunakpp:
## What

This closes #1886, which essentially removes the duplication of code in the `AssignmentRepo::CreateGitHubRepositoryJob` and `AssignmentRepo::Creator`. 
This is one step closer to code sharing between `AssignmentRepo` and `GroupAssignmentRepo` creators.


app/jobs/assignment_repo/create_github_repository_job.rb
```diff
@@ -13,150 +13,37 @@ class CreateGitHubRepositoryJob < ApplicationJob
     # user       - The User that the AssignmentRepo will belong to
     # retries    - The number of times the job will automatically retry
     #
-    # rubocop:disable MethodLength
-    # rubocop:disable AbcSize
     def perform(assignment, user, retries: 0)
```
d12: This method is so clean now 😍 

Generally, keeping logic out of our jobs is a great thing. Brought out into a model or class of it's own, the logic can be reused elsewhere. ✨

spec/jobs/assignment_repo/create_github_repository_job_spec.rb
```diff
@@ -347,7 +353,7 @@
 
       it "fails to save the AssignmentRepo and reports" do
         allow_any_instance_of(AssignmentRepo).to receive(:save!).and_raise(ActiveRecord::RecordInvalid)
-
+        expect(GitHubClassroom.statsd).to receive(:increment).with("exercise_repo.create.fail")
```
d12: @BenEmdon Does this look right to you? I'm not sure if both of these metrics should be getting hit, or just the v2 one.

shaunakpp: Even I had the same doubt, this is present in `GroupAssignmentRepo::Creator` too, should we stop sending stats for v1 in both? Also, if we're changing this, are any other places we can add more tracking?

BenEmdon: I had the `v2` prefixed metrics to compare changes I made last summer. I think its time to drop the `v2` prefix. 
If we do choose to do so we must make sure our data dog dashboards are pointed at the correct metrics ⚠️

shaunakpp: So should I just remove `v2` prefix on these stats? Or i could send both `v2` prefixed and not prefixed stat in this PR and then send a separate PR to remove `v2`. Just so that we don't lose any tracking. What do you think?

BenEmdon: > Or i could send both v2 prefixed and not prefixed stat in this PR and then send a separate PR to remove v2 prefix. Just so that we don't lose any tracking.

This ☝️ 🎉

shaunakpp: awesome! thanks! 🎉

PR #1911, opened 2019-06-27
Create LTI Auth Client

thebradbain:
## What
Creates an OmniAuth authentication strategy for LTI Tool launches from a Learning Management System, which is required regardless of the path an instructor takes to interact with GitHub classroom via an LMS. 

_Note_ Uses the [ims-lti](https://github.com/instructure/ims-lti) gem to do the heavy lifting, which is licensed under the MIT permissive license, so we should be good to use it.


d12:
Would you mind using a draft PR next time? Ping me when this is ready for review :)

thebradbain:
TIL about Draft Pull Requests 😆

thebradbain:
@d12 Ok _now_ it's ready for review 😉

app/controllers/sessions_controller.rb
```diff
@@ -28,11 +30,37 @@ def create
     redirect_to url
   end
 
+  def lti_launch
+    auth_hash = request.env["omniauth.auth"]
```
d12: Mind calling this `auth_obj` or something other than `auth_hash`? It looks like it's not actually a hash based on the `.uid` below.

app/controllers/sessions_controller.rb
```diff
@@ -28,11 +30,37 @@ def create
...
   end
 
   def failure
     redirect_to root_path, alert: "There was a problem authenticating with GitHub, please try again."
   end
+
+  private
+
+  def verify_lti_launch_enabled
+    return not_found unless lti_launch_enabled?
```
d12: :+1:

lib/omniauth/strategies/lti.rb
```diff
@@ -0,0 +1,69 @@
...
+          last_name: @authenticator.params["lis_person_name_family"],
+          image: @authenticator.params["user_image"]
+        }
+      end
+
+      credentials do
+        { token: @authenticator.consumer_key }
+      end
+
+      extra do
```
d12: What does `extra` do here? I mainly want to make sure we're handling data from @authenticator correctly. E.g. adding a secret to a query param would be bad.

d12:
Also looks like you have a small conflict, probably the faraday update or something similar. Merging master and re-bundling should fix.

app/controllers/sessions_controller.rb
thebradbain: No problem! Though we also name this same variable ‘auth_hash’ in the action relying upon the standard GitHub OmniAuth strategy (in the ‘sessions#create’ action) — I believe both are based on the standard OmniAuth authentication object, which they (perhaps misleadingly) [title the “Auth Hash”](https://github.com/omniauth/omniauth/wiki/Auth-Hash-Schema). 

Should I change the variable name in ‘sessions#create’ too for consistency?

lib/omniauth/strategies/lti.rb
thebradbain: The ‘secret’ is fortunately not returned nor available in the request parameters. But essentially, we are storing the entire raw request parameters (a bulky response containing _all_  LMS specific data about the authenticated user and actions available to them) in the `extra` field, which changes depending on the LMS-defined role of the user who authenticated. There is no way to re-request this data unless the User relaunches Classroom from the LMS, so we will want to persist this data (which will be my next PR :) )

app/controllers/sessions_controller.rb
d12: Ah dang. Yeah, if thats the standard, lets do `auth_hash`. Good call :)

config/routes.rb
```diff
@@ -19,8 +19,9 @@
   get  "/login/oauth/authorize", to: "oauth#authorize"
   post  "/login/oauth/access_token", to: "oauth#access_token"
 
-  match "/auth/:provider/callback", to: "sessions#create",  via: %i[get post]
-  match "/auth/failure",            to: "sessions#failure", via: %i[get post]
+  match "/auth/lti/callback",       to: "sessions#lti_launch",  via: %i[get post]
```
forced-request: Hello from Appsec!

I was hoping to get some clarification as to why we're accepting both `GET` and `POST` requests to these three routes. 

* Supporting `GET` requests will bypass the `verify_authenticity_token` checks, and leave the application potentially vulnerable to [CSRF](https://github.com/github/security-docs/blob/master/standards/Secure%20Coding%20Principles.md#csrf-tokens), which likely is not a concern for these integrations.
* Supporting `POST` requests is preferred by default because it enables us to pass sensitive data in the body parameters as to not leak the values in the query string. This would also execute the `verify_authenticity_token` to protect from CSRF. My concern here is that the LTI integration shouldn't have knowledge of the user's `authenticity token` to provide and as such, likely would raise an exception.

I guess my question is: Do we *need* to support both `GET` and `POST` requests to all three of these routes? Or is this just a quick catch-all?

PR #1912, opened 2019-06-28
Make search in group assignment not case sensitive

stephaniegiang:
## What
In my last PR https://github.com/education/classroom/pull/1904, I changed `LIKE` to `ILIKE` to make searches not case sensitive.

In this PR, I am doing the same with group assignment search.


spec/controllers/group_assignments_controller_spec.rb
```diff
@@ -149,7 +149,7 @@
         organization_id: organization.slug, id: group_assignment.slug, query: "test"
       }
 
-      expect(assigns(:group_assignment_repos)).to_not be_nil
```
d12: Can we use a more specific assertion here? Something like

```ruby
expect(assigns(:group_assignment_repos).first.id).to equal(group_assignment.groups.first.id)
```

spec/controllers/group_assignments_controller_spec.rb
```diff
@@ -159,6 +159,14 @@
 
       expect(assigns(:group_assignment_repos)).to eq([])
     end
+
+    it "search is not case senstive" do
+      get :show, xhr: true, params: {
+        organization_id: organization.slug, id: group_assignment.slug, query: "TEST"
+      }
+
+      expect(assigns(:group_assignment_repos)).to_not equal([])
```
d12: Likewise here

stephaniegiang:
@d12 changes were made!

PR #1917, opened 2019-06-28
Disable more caches, remove the entire http_cache gem

d12:
## What
_What does your pull request propose?_ :sparkles:


BenEmdon:
Go for it 👍

PR #1928, opened 2019-06-30
Update Ruby to 2.6.4

tarebyte:
We're a few versions behind, let's starting updating our dependencies.


tarebyte:
@BenEmdon I guess so, but I'm surprised by this failure.

I'm gonna push up a ruby 2.5 branch and see if there was a breaking change between versions.

jeffrafter:
Thanks @tarebyte and @BenEmdon 🌮 

As a side-note you may need to run:

```bash
rbenv install 2.6.4
gem install bundler:2.0.2
```

locally and you may need to:

```bash
bundle install
```

Most of the default bootstrapping will take care of this for you; but doing it manually might save you some seconds.

PR #1929, opened 2019-07-01
Add feature flag for LTI roster integration

stephaniegiang:
## What
This PR adds the feature flag for the LTI roster integration


stephaniegiang:
Going to use the flag in https://github.com/education/classroom/pull/1911 instead

PR #1931, opened 2019-07-01
Prevent using empty repositories as starter code repos

### LINKED ISSUE #1906: Starter code importer hangs when importing from empty repository
**Describe the bug**
The starter code importer hangs/breaks when the starter code repo is empty.

**To Reproduce**
1. Create a new assignment
2. Select an empty repo to use as starter code
3. Accept invitation 
4. Starter code importer hangs

**Expected behavior**
We should warn teachers if the starter code repo is empty and prevent them from choosing it.

We can easily fix this by checking the `size` attribute on the `github_repository`. I've confirmed that empty repos have a size of 0.

EricPickup:
## What
Fixes #1906, which causes the importer to hang when importing from an empty repo.

This PR adds a validator to `Assignment` and `GroupAssignment` that checks if their starter code repo is empty. It will prevent teachers from saving/creating an assignment with an empty repo selected.

I had to re-generate a bunch of cassettes since this added one more http call when creating an assignment.


spec/controllers/api/group_assignment_repos_controller_spec.rb
```diff
@@ -7,7 +7,8 @@
   let(:user)              { classroom_teacher }
   let(:student)      { classroom_student }
   let(:grouping)     { create(:grouping, organization: organization) }
-  let(:group) { create(:group, grouping: grouping, github_team_id: 3_284_880) }
+  let(:github_team_id) { organization.github_organization.create_team(Faker::Team.name).id }
+  let(:group) { create(:group, grouping: grouping, github_team_id: github_team_id) }
```
EricPickup: FYI: I'm working on splitting this test fix in a different PR that will fix this issue in all of our specs. It keeps coming up so I'm going to eliminate it completely in another PR.

PR #1934, opened 2019-07-01
Unify Creators for AssignmentRepo and GroupAssignmentRepo

### LINKED ISSUE #1948: Proposed changes for unifying repository creators
For unifying the existing creators, we need to make a few changes, especially to determine if we're creating a `GroupAssignmentRepo` or an `AssignmentRepo`. Also based on that, we need to make sure that appropriate API calls are made, to do this I'm proposing the following changes:

Move the unified repo creator from `models` to `services` as the creator isn't necessarily a value object or represents a database object. Implement the following classes to communicate with each other and have concise, well defined responsibilities. This in turn will also help in testing them better:

### CreateGitHubRepoService
- Communicates with GitHub API and has public facing method to start creation
- Has all the methods required for creation
- essentially a modified `Creator` class

### Result
- Manages status and error handling(use existing code, modify if necessary)

### Broadcaster
- Sends messages to the ActionCable
- manages correct channels
- manages message content

### StatsSender
- Send stats and timings to Datadog
- based on `Entity` send appropriate stats

### Entity
- Gets attributes like `GroupAssignment` or `Assignment`, `User`or `Group` and based on this decides the appropriate methods to use and names to use for content
- Example: if its a `GroupAssignmentRepo` then our messages should contain `team` instead of `user`, so entity will determine that
- Essential Object for unifying both the existing creators.

Since this is not implemented yet, all suggestions are welcome and appreciated😄. Especially when it comes to names and responsibilities of these objects. I'm not entirely happy about naming of `Entity` and `StatsSender` 😅

_Originally posted by @shaunakpp in https://github.com/education/classroom/pull/1934#issuecomment-508186700_

shaunakpp:
## What
Resolves #1948 
TODO:

- [x] ~~Move shared code to a module~~
- [x] Replace the module with a Creator service
- [x] handle dependencies and non-shared code
- [x] Creators should identify `Assignment` and `GroupAssignment` and call appropriate methods
- [x]  Add feature flag for gradual rollout
- [x] Write tests
- [x] Test this a lot 😅


shaunakpp:
For unifying the existing creators, we need to make a few changes, especially to determine if we're creating a `GroupAssignmentRepo` or an `AssignmentRepo`. Also based on that, we need to make sure that appropriate API calls are made, to do this I'm proposing the following changes:

Move the unified repo creator from `models` to `services` as the creator isn't necessarily a value object or represents a database object. Implement the following classes to communicate with each other and have concise, well defined responsibilities. This in turn will also help in testing them better:

## Classes

### CreateGitHubRepoService
- Communicates with GitHub API and has public facing method to start creation
- Has all the methods required for creation
- essentially a modified `Creator` class

### Result
- Manages status and error handling(use existing code, modify if necessary)

### Broadcaster
- Sends messages to the ActionCable
- manages correct channels
- manages message content

### StatsSender
- Send stats and timings to Datadog
- based on `Entity` send appropriate stats

### Entity
- Gets attributes like `GroupAssignment` or `Assignment`, `User`or `Group` and based on this decides the appropriate methods to use and names to use for content
- Example: if its a `GroupAssignmentRepo` then our messages should contain `team` instead of `user`, so entity will determine that
- Essential Object for unifying both the existing creators.


Since this is not implemented yet, all suggestions are welcome and appreciated😄. Especially when it comes to names and responsibilities of these objects. I'm not entirely happy about naming of `Entity` and and `StatsSender` 😅

app/controllers/assignment_invitations_controller.rb
```diff
@@ -35,6 +35,8 @@ def setup; end
 
   # rubocop:disable MethodLength
   # rubocop:disable AbcSize
+  # rubocop:disable CyclomaticComplexity
+  # rubocop:disable PerceivedComplexity
```
shaunakpp: These are temporary disables. Once we rollout to 100% we can remove the old job and these disables.

app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -40,6 +40,7 @@ def setup; end
 
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable MethodLength
+  # rubocop:disable PerceivedComplexity
```
shaunakpp: As mentioned above, these are temporaray.

d12:
It looks like specs are failing right now, please ping me when this is ready for :eyes:!

d12:
Oh it's just a linting error. I'll :eyes: this today!

app/services/create_github_repo_service.rb
```diff
@@ -0,0 +1,208 @@
+# frozen_string_literal: true
+
+# rubocop:disable ClassLength
+class CreateGitHubRepoService
+  attr_reader :entity, :stats_sender
+  delegate :assignment, :collaborator, :organization, :invite_status, to: :entity
+
+  def initialize(assignment, collaborator)
+    @entity = Entity.build(assignment, collaborator)
```
d12: Can you give `Entity` a more specific, descriptive name? I'm not sure what this class does.

app/services/create_github_repo_service.rb
```diff
@@ -0,0 +1,208 @@
...
+
+  # Public: Add user/team to the GitHubRepository based on the type of assignment.
+  #         Calls #add_user_to_github_repository! if it is an Assignment.
+  #         Calls #add_group_to_github_repository! if it is a GroupAssignment.
+  #
+  # github_repository - GitHubRepository in which we need to add collaborator.
+  #
+  # Returns true if collaborator added or raises a Result::Error.
+  def add_collaborator_to_github_repository!(github_repository)
+    send("add_#{entity.humanize}_to_github_repository!", github_repository)
```
d12: Can we write this without `send` and string interpolation? This makes for code that's difficult to read and search for. Explicit conditionals are preferred.

app/services/create_github_repo_service.rb
```diff
@@ -0,0 +1,208 @@
...
+    end
+  end
+
+  # Internal: Method for error messages, modifies error messages based on the type of assignment.
+  #
+  # error_message - A symbol for getting the  appropriate error message.
+  # rubocop:disable LineLength
+  # rubocop:disable MethodLength
+  # rubocop:disable AbcSize
+  def errors(error_message, options = {})
```
d12: Can this method be re-organized as a case statement? Then we're not initializing and interpolating a bunch of unneeded strings every time this gets called.

app/services/create_github_repo_service/broadcaster.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class CreateGitHubRepoService
+  class Broadcaster
+    def self.call(entity, message, message_type, repo_url = nil)
+      new(entity, message, message_type, repo_url).call
+    end
+
+    attr_reader :entity, :message, :message_type, :repo_url
```
d12: Can this go at the top of the class, above `self.call`?

app/services/create_github_repo_service/broadcaster.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class CreateGitHubRepoService
+  class Broadcaster
+    def self.call(entity, message, message_type, repo_url = nil)
+      new(entity, message, message_type, repo_url).call
```
d12: The fact that we're initializing an instance and then immediately throwing it a way is a smell that we should probably get rid of the instance methods here and just use class methods. I don't see any places in the codebase where we're relying on `Broadcaster` holding any state.

app/services/create_github_repo_service/broadcaster.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def channel_hash
+      {}.tap do |msg|
+        msg["#{entity.humanize}_id".to_sym] = entity.collaborator.id
+        msg["#{entity.assignment_type}_id".to_sym] = entity.assignment.id
+      end
+    end
+
+    def messages
```
d12: This looks like it should be a frozen constant instead of a method.

app/services/create_github_repo_service/entity.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    def repo_name
+      @repo_name ||= generate_repo_name
+    end
+
+    def default_repo_name
+      "#{assignment.slug}-#{slug}"
+    end
+
+    def owner
```
d12: This is an organization login, could you make the method name more descriptive?

app/services/create_github_repo_service.rb
shaunakpp: Even I'm not very happy with this name 😢 , do you have any suggestions? the job of this class is to manage `Assignment` and `GroupAssignment` related methods that have the same use, but different methods. e.g to get repos for `Assignment` would be `assignment.assignment_repos` and for `GroupAssignment` it would be `group_assignment.group_assignment_repos`. So with `Entity` we get a single method `repos`. And more such kind of methods.

d12: Possibly `Exercise` or `GenericAssignment`?

shaunakpp: I'm in favour of `Exercise`! Also, a few more renames like moving all related classes under the namespace `GitHubRepositoryCreator` and then rename `CreateGitHubRepositoryService1 -> `GitHubRepositoryCreator::Service` and rename `CreateGitHubRepositoryNewJob` -> `GitHubRepositoryCreator::Job`, what do you think?

app/services/create_github_repo_service/broadcaster.rb
shaunakpp: I agree, but can we change it have an object with state and re-use it for wherever required something like:
```
broadcaster = Broadcaster.new(entity)
# for message
broadcaster.report_message(:repository_creation_complete, repo_url)
# for error
broadcaster.report_error(error_message)
```

app/services/create_github_repo_service/entity.rb
shaunakpp: renaming it to `organization_login`

app/services/create_github_repo_service/exercise.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    def suffixed_repo_name(suffix_count)
+      return default_repo_name if suffix_count.zero?
+
+      suffix = "-#{suffix_count}"
+      default_repo_name.truncate(100 - suffix.length, omission: "") + suffix
+    end
+  end
+
+  class Individual < Exercise
```
d12: These classes should be in their own files. Rails autoloading will not be able to load these classes if they haven't already been loaded, and it makes locating the definitions harder.

app/services/create_github_repo_service/exercise.rb
```diff
@@ -0,0 +1,105 @@
...
+    def humanize
+      "user"
+    end
+
+    def stat_prefix
+      "exercise_repo"
+    end
+  end
+
+  class Team < Exercise
```
d12: Same here

app/services/create_github_repo_service/stats_sender.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+# Wrapper class over GitHubClassroom.statsd for creating repositories
+# This enables us to add extra stats based on the progress of the repository being created
+#
+class CreateGitHubRepoService
+  class StatsSender
+    class InvalidStatError < StandardError; end
+    attr_reader :exercise
+    def initialize(exercise)
```
d12: Can you add a newline after `class InvalidStatError < StandardError; end` and `attr_reader :exercise`

app/services/create_github_repo_service/stats_sender.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    # Public: Method for reporting a stat to GitHubClassroom.statsd,
+    #         raises an InvalidStatError if appropriate method not found.
+    #
+    # stat - symbol for calling a stat.
+    #
+    # Returns the current value of increment from statsd.
+    def report(stat)
+      send(stat)
+    rescue NoMethodError
```
d12: Can this be designed in a way that doesn't use `send` and `rescue NoMethodError`? If we do development on any methods called by `send` here and we introduce a NoMethodError during development, it will get swallowed here and make debugging a pain.

It looks like we have a mapping from symbols to error messages. Can this be a case statement?

app/services/create_github_repo_service/exercise.rb
shaunakpp: Thanks for pointing this out! I totally forgot to move them to new files once I'd written them

app/services/create_github_repo_service/broadcaster.rb
d12: We could, but I don't see anywhere in the app where we're actually re-using a Broadcaster instance. As is, it's just unnecessary overhead we go through in Ruby every time we want to broadcast a message.

app/services/create_github_repo_service/broadcaster.rb
```diff
@@ -0,0 +1,58 @@
...
+    MESSAGES = {
+      repository_creation_complete: "Your GitHub repository was created.",
+      import_ongoing: "Your GitHub repository is importing starter code.",
+      create_repo: "Creating GitHub repository.",
+      importing_starter_code: "Importing starter code."
+    }.freeze
+
+    attr_reader :exercise, :message, :message_type, :repo_url
+
+    def self.call(exercise, message, message_type, repo_url = nil)
```
d12: See my earlier comment, we aren't re-using `Broadcaster` instances anywhere in the app, I think we should remove the overhead of initializing a object every time we want to broadcast a message and just use class methods instead.

PR #1935, opened 2019-07-02
LTI integration User Interface

stephaniegiang:
## What
This PR creates the initial views for the LTI integration. Back-end will be another PR. Everything is behind the feature flag `lti_launch`.

## Flow

This is the first screen that users will see when click `LMS configuration`.

image

When clicking the `Connect to LMS` it will bring users to this screen where they will get their consumer key and shared secret. Because there is no backend for this, I just put in a placeholder for the key and secret. All will be added in the next PR. 

image

Users will always see this page as long as there is a current LMS configuration


stephaniegiang:
Going to use the flag in https://github.com/education/classroom/pull/1911 instead

stephaniegiang:
Ops wrong PR

stephaniegiang:
This PR is ready for review! There are some tests that are failing from a previous PR but @thebradbain is on it!

app/controllers/application_controller/feature_flags_dependency.rb
```diff
@@ -65,7 +69,7 @@ def search_assignments_enabled?
   helper_method :search_assignments_enabled?
 
   def lti_launch_enabled?
-    GitHubClassroom.flipper[:lti_launch].enabled?
```
thebradbain: It appears that those tests within `spec/integration/lti_launch_spec.rb` fail because of this change to this feature check. 

It's important that possibly-unauthenticated users still be able to launch into GitHub Classroom from their LMS so we can immediately persist that launch data for the session and _then_ log the user in normally if they are not already, and we definitely want to be able to test that this is indeed what happens (otherwise we have no idea who the session launch data belongs to).

Perhaps we want something like `GitHubClassroom.flipper[:lti_launch].enabled? || (logged_in? && current_user.feature_enabled?(:lti_launch))` to ensure in testing on production we can still enable it just for staff?

app/views/orgs/lti_configurations/show.html.erb
```diff
@@ -0,0 +1,45 @@
+<% organization = @organization || current_organization %>
```
thebradbain: I think setting this variable within the view may be unnecessary, we could probably just use `current_organization` directly since it's guaranteed to be set by this point courtesy of the `before_action` on Orgs::Controller

app/views/orgs/lti_configurations/new.html.erb
```diff
@@ -0,0 +1,40 @@
+<% organization = @organization || current_organization %>
```
thebradbain: See other comment re: this variable

stephaniegiang:
@d12 @BenEmdon @srinjoym ready for reviews!

spec/controllers/orgs/lti_configurations_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Orgs::LtiConfigurationsController, type: :controller do
+  let(:organization) { classroom_org                                   }
+  let(:user)         { classroom_teacher                               }
```
BenEmdon: Nit: this looks a little janky
```suggestion
  let(:user)         { classroom_teacher }
```

spec/controllers/orgs/lti_configurations_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Orgs::LtiConfigurationsController, type: :controller do
+  let(:organization) { classroom_org                                   }
```
BenEmdon: Nit: this looks a little janky
```suggestion
  let(:organization) { classroom_org }
```

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Orgs
+  class LtiConfigurationsController < Orgs::Controller
+    before_action :ensure_lti_launch_flipper_is_enabled
+    before_action :ensure_current_lti_configuration, except: %i[new create]
+
+    def create
+      # Create a new lti configuration for the current organization here
+      # redirect_to lti_configuration_path(current_organization) after
```
BenEmdon: Is this a `TODO`? Or are you documenting this method?

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Orgs
+  class LtiConfigurationsController < Orgs::Controller
+    before_action :ensure_lti_launch_flipper_is_enabled
+    before_action :ensure_current_lti_configuration, except: %i[new create]
```
BenEmdon: Does this controller support the `new` route? Is it convention to have a `#new` method even when it does nothing (like you did for `#create`)?

app/controllers/orgs/lti_configurations_controller.rb
stephaniegiang: This is a TO-DO. This PR just focuses on the front end of the application. The next PR will include the back-end logic.

app/controllers/orgs/lti_configurations_controller.rb
stephaniegiang: The controller does.

That is a good question. Maybe we should move the information page to a different route and have the `#new` be where teachers actually create the new config. What do you think @thebradbain?

thebradbain: `#new` as an action is kind of awkward in this flow because it essentially just displays a view with a button, that, once an instructor clicks "generate LMS config" will simply call out to the `#create` action and generate its `shared_secret` and `consumer_key` fields with no other user input, and then direct them to the #show action.

I could get behind `#index` instead, but as this is a page we would only show in the context of when an `LtiConfiguration` doesn't exist for the organization, I also wonder if that makes sense

stephaniegiang:
@BenEmdon added screenshots and a gif of the flow

app/controllers/orgs/lti_configurations_controller.rb
stephaniegiang: Yeah that makes sense. I'm ok with the info view being under `#new` function. It is a bit awkward but I think that's ok.

app/controllers/orgs/lti_configurations_controller.rb
BenEmdon: Typically the convention for leaving todos is to start the comment with `TODO` so that a other developers can discover the TODO's with [`rake notes`](https://guides.rubyonrails.org/command_line.html#notes)

stephaniegiang:
I'll took into the button

config/routes.rb
```diff
@@ -84,6 +84,8 @@
           patch :add_students
           patch :remove_organization
         end
+
+        resource :lti_configuration, controller: "orgs/lti_configurations"
```
BenEmdon: Looks like you aren't exposing all of the resource actions on `lti_configuration`, so maybe you should only allow `show`, `new`, and `create`?
```suggestion
        resource :lti_configuration, only: %i[show new create], controller: "orgs/lti_configurations"
```

stephaniegiang: Can that be in the next PR when we work on those actions?

app/views/orgs/lti_configurations/new.html.erb
```diff
@@ -0,0 +1,38 @@
...
+          <p>You or your system administrator will need to add a new application to
+          your Learning Management System with the following credentials generated by
+          GitHub Classroom.</p>
+          <ul>
+            <li>Consumer Key</li>
+            <li>Shared secret</li>
+            <li>Link to the LMS classroom page</li>
+          </ul>
+
+          <%= button_to "Connect to LMS", lti_configuration_path(current_organization), class: 'btn btn-primary' %>
```
srinjoym: Hitting this error when I click this button in development: `No template found for Orgs::LtiConfigurationsController#create`

app/controllers/orgs/lti_configurations_controller.rb
BenEmdon: The purpose of the `new` action in Rails is create an HTML form for creating a new resource ([src](https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions)). I would say that a form containing only a button counts 👍  We are good to resolve this discussion.

BenEmdon: Should this instead be the inverse?
```suggestion
    before_action :ensure_current_lti_configuration, only: :show
```

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def create
+      # TODO: Create a new lti configuration for the current organization here
+      # TODO: redirect_to lti_configuration_path(current_organization) after
+    end
+
```
BenEmdon: Nit: In this project we usually implement empty methods for actions which have routes. This slightly improves readability.
```suggestion
    def show; end

    def new; end

```

stephaniegiang:
Just talked to @srinjoym there are some things that I'm going to change.

- [ ]  Move the `new` to `info` to make it more clear
- [ ]  Disable the shared secret and consumer key fields.

app/views/orgs/lti_configurations/new.html.erb
stephaniegiang: This is expected since we don't have a view for this and the method in the controller is empty. This will be fixed in our next PR.

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    def show; end
+
+    def info; end
+
+    def edit;
```
srinjoym:
```suggestion
    def edit; end
```

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190709163639) do
+ActiveRecord::Schema.define(version: 20190709163513) do
```
srinjoym: We should probably figure out why merging master brings this change in but for now we can just remove it

spec/controllers/orgs/lti_configurations_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+RSpec.describe Orgs::LtiConfigurationsController, type: :controller do
+  let(:organization) { classroom_org }
+  let(:user)         { classroom_teacher }
+
+  before(:each) do
+    sign_in_as(user)
+  end
+
+  describe "GET #info", :vcr do
```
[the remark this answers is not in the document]
stephaniegiang: Do we need to if these pages don't exist? And the methods are empty? You would just hit a template error.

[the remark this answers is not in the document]
stephaniegiang: Those methods will be worked on the next PR, so I will definitely have tests for that in the next one.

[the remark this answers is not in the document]
BenEmdon: I see that they don't do or render anything yet 👀 my bad!

PR #1947, opened 2019-07-02
Initial model for LTI Configuration

### LINKED ISSUE #1945: LMS-Classroom Configuration Modeling
A `Classroom` instance should have (at most) one `LMSConfiguration` model, which contains parameters specific to that instance of the classroom organization to tie it to an LMS.

Required model columns include: 

`consumer_key` -- the classroom-specific identifier linking a classroom to an LMS
`shared_secret` -- should _never_ be sent over the wire; used to decrypt the payload from the LMS, also specific to each classroom

stephaniegiang:
## What
Closes #1945
Creating initial model `lti_configuration.rb` that will be used to store LMS consumer key, shared secret, and lms link.


db/migrate/20190702222417_create_lti_configurations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateLtiConfigurations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :lti_configurations do |t|
+      t.text :consumer_key
```
BenEmdon: Should this part of the model be non nullable?

spec/models/lti_configuration_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe LtiConfiguration, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
BenEmdon: You could add a shoulda matcher test like so: https://github.com/education/classroom/blob/1f81278aca77b523d8b821ce206d6a35e523217e/spec/models/invite_status_spec.rb#L11

Suggestion:
```suggestion
  it { should belong_to(:organization) } 
```

db/migrate/20190702222417_create_lti_configurations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateLtiConfigurations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :lti_configurations do |t|
+      t.text :consumer_key, null: false
```
thebradbain: I think we will actually also want to ensure this is unique, as according to the [Security Specification](http://ltiapps.net/guide/LTI_Best_Practice.pdf) best practices dictate that the `oauth_consumer_key` is a "unique identifier for the tool consumer"

stephaniegiang: Yes, that is something we should add. I'll also make sure the `lms_link` is unique too since we don't want teachers to be linking a LMS classroom that was already added.

thebradbain: I think actually the link itself should be fine to be duplicated, it's only what we redirect professors to to make their lives easier, and there's no guarantee the URL on an LMS will be unique for classes (Sakai, my school's LMS, for example operates exclusively through POST requests, so there is essentially only one URL visible in the address bar at all times)

thebradbain:
Closes #1945

stephaniegiang:
Pulled out the migration into this PR https://github.com/education/classroom/pull/1950

Issue #1951, opened 2019-07-03
Remove feature flags at 100%


stephaniegiang:
We should remove flags that have been at 100% for a while. Feel free to add to this list. 

A list of flags we should remove: 
- [x] student_identifier
- [x] public_home_v2
- [x] public_assistant_landing_page
- [x] multiple_classrooms_per_org
- [x] import_resiliency
- [x] group_import_resiliency
- [x] download_repositories
- [x] deadlines
- [x] dashboard_search

d12:
`deadlines` was removed a while ago here https://github.com/education/classroom/commit/6c09575f96121f775bb8813a1bb3dece222769ad

I'll remove the feature from production flipper.

PR #1952, opened 2019-07-04
Remove assistant_landing_page and public_assistant_landing_page

d12:
cc https://github.com/education/classroom/issues/1951

Removes:
- public_assistant_landing_page
- assistant_landing_page

assistand_landing_page is not 100%, but it's also not being used in the codebase. Is this OK @srinjoym ?


BenEmdon:
> assistand_landing_page is not 100%, but it's also not being used in the codebase. Is this OK @srinjoym ?

Weird that it's not at 100%. I think the intention was that it should be?

PR #1958, opened 2019-07-05
Kill dashboard_search flag

d12:
cc https://github.com/education/classroom/issues/1951

Removes old `dashboard_search` flipper logic.


app/views/organizations/index.html.erb
```diff
@@ -1,11 +1,7 @@
 <div class="d-sm-flex flex-items-center flex-justify-between my-4">
   <% if @organizations.present? %>
-    <% if dashboard_search_enabled? %>
-      <%= form_tag search_organizations_path, id: 'js-search-form', class: 'form-block col-md-4', method: :get, remote: true do %>
-        <%= text_field_tag :query, params[:query], class: 'form-control width-full', placeholder: 'Search Classrooms...', autofocus: true, autocomplete: :off %>
-      <% end %>
-    <% else %>
-      <h2 class="f2 text-normal text-gray"><%= t('views.organizations.classrooms') %></h2>
```
stephaniegiang: Why are we removing this?

d12: Because `dashboard_search_enabled` is always true (and we're removing the feature now), so we don't need the logic in the `else`.

PR #1959, opened 2019-07-05
Persist LTI identifying credentials for user session upon launch

thebradbain:
## What
Persists LTI data for the current session/user upon LMS launch, tying the functionality in #1949 and #1911 together.

## Why Redis over Postgres

We’re using redis to primarily persist the given LTI data for the length of the user’s session since that’s the only time it’s guaranteed to be valid (but the response is too big in itself to store in the session, so we store it in redis, keyed by the unique nonce given to us by the response).

Furthermore, the concrete data given about the LMS-user from a launch is dependent on multiple, possibly-changing LMS-specific factors: the school-defined role of the user logging in (e.g. instructor, student, TA) and what they have permissions to do, school-specific security settings, and the fact the user may or may not be able change information (e.g. name, email) at any time. 

Fortunately, most of the changing/non-guaranteed information is not stuff we care about too much on our end: every LTI Launch regardless of institution will have a 

- `consumer_key`: unchanging and unique to the corresponding Classroom organization connected to the LMS, per the standard 
- `oauth_nonce`: a unique key used to fetch a launch message in redis, changes every new launch, 
- `oauth_signature`: used for signing requests 

However, it still does not make sense to persist any of this information to Postgres (minus the `consumer_key`, which we use to tie an `Organization` to an `LtiConfiguration`), as once the `oauth_signature` expires or is invalidated by a user logging into their LMS to perform any action, even if not related to GitHub classroom, the received message is no longer guaranteed to be accurate and the `oauth_signature` may no longer be used to sign requests.

**TLDR** LTI makes very little guarantees about the permanence of the actual information it's passing over. Storing the authentication information in Redis and keying it to the `oauth_nonce` for the duration of the user’s session is the best path forward, since it will most likely expire sooner rather than later.


app/controllers/sessions_controller.rb
```diff
@@ -30,25 +30,44 @@ def create
...
+    strategy = request.env["omniauth.strategy"]
+    strategy.options.consumer_key = consumer_key
+    strategy.options.shared_secret = shared_secret
+
+    head :ok
+  end
+
+  # rubocop:disable MethodLength
+  # rubocop:disable AbcSize
   def lti_launch
```
d12: It looks like there's some untested logic in this method, would you mind adding some tests to fix that up?

d12:
Also, would you mind reminding me why we're using Redis instead of our normal Postrges DB here?

thebradbain:
So we’re using Redis to primarily persist the given LTI data for the length of the user’s session since that’s the only time it’s guaranteed to be valid (but the data is too big in itself to store in the session, so we use redis keyed by a unique nonce given to us in the launch request stored in the session). 

Furthermore, the concrete data given about the LMS-user from a launch  is dependent on multiple, possibly-changing LMS-specific factors: the role of the user logging in (instructor, student, TA) and what they have permissions to do, school-specific security settings, and the fact the user may be able change information (e.g. name, email) at any time.

Fortunately, most of the changing/non-guaranteed information is not stuff we care about too much on our end: every LTI Launch regardless of institution will have a `consumer_key` (unique to the instance installed on the LMS, per the standard), an `oauth_nonce` (used to key the message in redis), and an `oauth_signature` (used for signing requests). However, it still does not make sense to persist any of this information to Postgres (minus the `consumer_key`, which we do to tie an Organization to an LTI Configuration), as once the `oauth_signature` expires or is invalidated by a user logging into their LMS to perform any action, even if not related to GitHub classroom, the received message is no longer

thebradbain:
@d12 started explaining and ended up writing an entire essay 😅

Moved comment to the description of the PR

thebradbain:
Should be ready for a second round of review!

thebradbain:
@d12 ready for a second round of review!

PR #1960, opened 2019-07-06
Adding LTI Configuration Reference to Organization Migration

stephaniegiang:
## What
This PR adds a LTIConfiguration `belongs_to` relationship to Organization. So we can access the data like `current_organization.lti_configuration`.


thebradbain:
I’ve left a comment on the other PR, but a database migration won’t be necessary if we use `has_one :lti_configuration, optional: true` on organization instead of `belongs_to` since the association and key already exists on `lti_configuration`

PR #1961, opened 2019-07-06
Add "has_one" relationship on organization to lti_configuration

stephaniegiang:
## What
This PR adds the `has_one` relationship between `organization` and `lti_configuration`.


thebradbain:
I _think_ this should be a `has_one :lti_organization, optional: true` instead of a `belongs_to`, as we already have the foreign key on the `:lti_organization` table

stephaniegiang:
@d12 ready for another review

PR #1964, opened 2019-07-08
Denormalize GitHub attributes on User

### LINKED ISSUE #1726: Roster management Page is too slow
When opening the roster page while you have a lot of unlinked users it will take a lot of time loading 
It takes around 30-45 Seconds with me.

I tracked what is causing this delay and found that the issue is on this line below:

https://github.com/education/classroom/blob/6d3fa4663268345a264a0e77e031eb0da81471ca/app/views/orgs/rosters/_link_to_github_account_modal.html.erb#L14

It's not a bug but the way its done now slow things down. 
It have to connect to github.com for every unlinked user.

What do you think?

d12:
Closes #1726

Build a write-through cache of GitHub attributes on the User model. The interface doesn't change at all besides a new optional `use_cache: true` kwarg on `github_` methods.

When `use_cache` is true and we don't have the attribute cached, we make an API request. While updating the requested field, we also update all the other cacheable fields on the model so that we don't get 5 API requests when fetching 5 uncached fields.

Example:

```
[33] pry(main)> Benchmark.ms { u.github_user.login }
   (3.7ms)  BEGIN
  User Exists (5.5ms)  SELECT  1 AS one FROM "users" WHERE "users"."token" = $1 AND ("users"."id" != $2) LIMIT $3  [["token", "fee5f8634f28c1d8f0c10d811d841cd09f15f7cc"], ["id", 1], ["LIMIT", 1]]
  User Exists (5.9ms)  SELECT  1 AS one FROM "users" WHERE "users"."uid" = $1 AND ("users"."id" != $2) LIMIT $3  [["uid", 30920216], ["id", 1], ["LIMIT", 1]]
  SQL (7.3ms)  UPDATE "users" SET "updated_at" = $1, "github_login" = $2, "github_name" = $3 WHERE "users"."id" = $4  [["updated_at", "2019-07-08 15:18:40.587189"], ["github_login", "d12"], ["github_name", "Nathaniel Woodthorpe"], ["id", 1]]
   (8.4ms)  COMMIT
=> 252.147999957203865

[34] pry(main)> Benchmark.ms { u.github_user.login }
=> 0.01200009137392044

[35] pry(main)> Benchmark.ms { u.github_user.name }
=> 0.03400002606213093

[36] pry(main)> Benchmark.ms { u.github_user.name(use_cache: false) }
=> 323.61600013636051
```

This is pretty easily extensible, all we need to do is add `def local_cached_attributes` to the GitHub resources we want to use this with. User causes the most API requests so I started here, but once this is in place I'll add organization, team, etc.

Once we've fully moved to locally cached values, we can :fire: some old stuff in `github_model.rb` that was fetching all the attributes from the API on every single instance load.


app/models/assignment_repo/creator.rb
```diff
@@ -116,7 +116,7 @@ def add_user_to_repository!(github_repository_id)
       options = {}.tap { |opt| opt[:permission] = "admin" if assignment.students_are_repo_admins? }
 
       github_repository = GitHubRepository.new(organization.github_client, github_repository_id)
-      invitation = github_repository.invite(user.github_user.login_no_cache, options)
```
d12: `(use_cache: false)` is the new API for skipping local + GitHub caches. I'll be removing the definitions for these `_no_cache` methods if we're able to convert all github_ resources to the new API.

app/models/github_model.rb
```diff
@@ -49,7 +52,33 @@ def initialize(client, id_attributes, **options)
         attr[gh_attr.to_sym] = github_response(client, id_attributes.values.compact, options).send(gh_attr)
       end
 
-      remove_instance_variable("@response")
+      local_cached_attributes.each do |gh_attr|
```
d12: This takes the list of attributes defined in `local_cached_attributes` and defines the caching method for each one.

app/models/github_model.rb
```diff
@@ -49,7 +52,33 @@ def initialize(client, id_attributes, **options)
...
+        define_singleton_method(gh_attr) do |use_cache: true|
+          field_name = "github_#{gh_attr}".to_sym
+
+          no_cache_options = options.dup
+          no_cache_options[:headers] = GitHub::APIHeaders.no_cache_no_store
```
d12: We add this to the options hash because if we make an API request, it's either because we want to skip the cache or we need to warm our cache. In both cases, we should ask GitHub for the most recent data.

lib/collab_migration.rb
```diff
@@ -34,7 +35,7 @@ def github_organization
   end
 
   def github_user
-    @github_user ||= GitHubUser.new(@repo_access.user.github_client, @repo_access.user.uid)
+    @github_user ||= GitHubUser.new(@user.github_client, @user.uid, classroom_resource: @user)
```
thebradbain: This is probably something completely obvious I'm overlooking and more out curiosity, but why are we passing in a `classroom_resource` reference to the classroom user?  I'd imagine its because that's where we're cacheing the fields, but the only changes I see in reference to it is deleting it within the [initializer for GitHubModel](https://github.com/education/classroom/pull/1964/files#diff-6a9532badda0e079e68cf8a3cbdf8766R37)

d12: That `delete` serves two purposes. One, it removes the key from the hash, but `Hash#delete` also returns the value deleted. So `resource = options.delete(:classroom_resource)` removes the key from the hash and assigns the value to `resource`. We then use resource when building the cache here: https://github.com/education/classroom/pull/1964/files#diff-6a9532badda0e079e68cf8a3cbdf8766R69

PR #1972, opened 2019-07-09
Add stats for repo creation time with importer

EricPickup:
## What
This PR adds a metric for the repo creation time when using the source importer.

I'll also add a metric for template repo creation which lets us compare the time/performance difference when using template repos vs the source importer.

### DIFF
--- a/app/models/assignment_repo/creator/reporter.rb
+++ b/app/models/assignment_repo/creator/reporter.rb
@@ -34,10 +34,14 @@ class AssignmentRepo
 
       # Reports the elapsed time to Datadog
       #
-      def report_time(start_time)
+      def report_time(start_time, assignment)
         duration_in_millseconds = (Time.zone.now - start_time) * 1_000
-        GitHubClassroom.statsd.timing("v2_exercise_repo.create.time", duration_in_millseconds)
-        GitHubClassroom.statsd.timing("exercise_repo.create.time", duration_in_millseconds)
+        if group_assignment.starter_code?
+          GitHubClassroom.statsd.timing("exercise_repo.create.time.with_importer", duration_in_millseconds)
+        else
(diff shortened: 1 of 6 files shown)


EricPickup:
@d12 I already have them separated in the PR that will introduce template repos, it's just not up yet!
```
if assignment.use_importer?
  GitHubClassroom.statsd.timing("v2_exercise_repo.create.time.with_importer", duration_in_millseconds)
elsif assignment.use_template_repos?
  GitHubClassroom.statsd.timing("v2_exercise_repo.create.time.with_templates", duration_in_millseconds)
else
  GitHubClassroom.statsd.timing("v2_exercise_repo.create.time", duration_in_millseconds)
end
```

app/models/assignment_repo/creator/reporter.rb
```diff
@@ -34,10 +34,14 @@ def broadcast_message(type: :text, message:, status_text:, repo_url: nil)
 
       # Reports the elapsed time to Datadog
       #
-      def report_time(start_time)
+      def report_time(start_time, assignment)
         duration_in_millseconds = (Time.zone.now - start_time) * 1_000
-        GitHubClassroom.statsd.timing("v2_exercise_repo.create.time", duration_in_millseconds)
-        GitHubClassroom.statsd.timing("exercise_repo.create.time", duration_in_millseconds)
+        if group_assignment.starter_code?
+          GitHubClassroom.statsd.timing("exercise_repo.create.time.with_importer", duration_in_millseconds)
```
EricPickup: cc @shaunakpp how does this look? Using v1 for the new stat and keeping v1/v2 for the existing ones?

d12: :+1: To this

shaunakpp: yes that looks good 👍

Issue #1969, opened 2019-07-09
Slow assignment page load for large rosters


d12:
When rendering roster entries in the assignment view, we execute this query to properly order the relation: https://github.com/education/classroom/blob/17848e109b977bd19c8cc26cf80ed8438f602639/app/models/roster_entry.rb#L66-L78

If `sql_formatted_users` is too large, this query becomes ridiculously slow. We need to find a better way to order this relation without passing so many IDs into IN, which is known to be slow. https://stackoverflow.com/questions/4514697/mysql-in-operator-performance-on-large-number-of-values

Example:

```ruby
Benchmark.ms { RosterEntry.order_for_view(assignment).to_a }
=> 11483.21          # This is 11 seconds
```

d12:
To solve this issue properly, we might want to consider redesigning the assignment show page a little bit. Roster entries were jammed into an existing view and made a bit of a mess. We may be able to organize this data in a better way such that we don't need these complex queries anymore.

The main thing is that we have 1 list of:

- Repos not linked to a student
- Repos linked to a student
- Roster entries that don't have a repo yet

Then in a second tab, we have roster entries that aren't linked yet. I'll do some thinking into how we can structure this differently to be less confusing (and more performant).

PR #1982, opened 2019-07-10
Search by GitHub attributes on non-roster AssignmentRepos

### LINKED ISSUE #1978: Allow teachers to search and sort by GitHub handles
We now have caching for GitHub handles. So we should now allow teachers to search and sort by GitHub handles. So we can now rollout this feature to teachers that do not have rosters.

thebradbain:
## What
Closes #1978 

Because we now cache GitHub handles, we can now enable teachers to search and sort by GitHub login for classroom organizations that do not have a roster linked to them.

This PR is also an opportunity to DRY the search functionality into a model concern (`Searchable`) much like the sort functionality (`Sortable`) already was.


app/controllers/assignments_controller.rb
```diff
@@ -129,8 +129,12 @@ def set_assignment
     @assignment = @organization.assignments.includes(:assignment_invitation).find_by!(slug: params[:id])
   end
 
+  def set_list_type
+    @list_type = @organization.roster ? :roster_entries : :assignment_repos
```
thebradbain: I've created a `@list_type` instance variable set in a `before_action` to help make the view logic a little more readable, whereas previously were only relying `if @organization.roster` scattered everywhere for the same functionality

thebradbain:
As an aside, you'll notice the actual backend changes to enable search/sort on unlinked assignment repos is pretty trivial and straightforward, _but_ the view logic as it currently exists, while ultimately trivial changes, is not very DRY. This is perhaps an indication that we should consider splitting up `AssignmentsController#show` into two actions depending on whether or not  `@organization.roster` is present for the assignment, because the views are differing enough now that scattered `if/else` statements are beginning to be insufficient.

d12:
> This is perhaps an indication that we should consider splitting up AssignmentsController#show into two actions depending on whether or not @organization.roster is present for the assignment, because the views are differing enough now that scattered if/else statements are beginning to be insufficient.

+ 100, I'd like to redesign these views to make more logical sense, you can tell that we're overcomplicating the design by the horrible spaghetti code we have. I mention it a little bit here: https://github.com/education/classroom/issues/1969#issuecomment-509711272

Also a note, lets not merge this until I populate the caches. Right now, the cache get populated when `github_user.login` gets fetched. That means some users have logins, but a lot still don't. I'll write a task to go ahead and backfill them all, and make sure we're filling in the login on creation.

d12:
Issue to populate the caches: https://github.com/education/classroom/issues/1984

thebradbain:
@d12 The tests have been added so whenever the cache is warmed we should be ready to rollout

d12:
Thanks @thebradbain ! I have something working locally, just gonna do a bit of local testing and then I'll push that up. The task may take a very long time to run (days), we'll see 😬

Issue #1926, opened 2019-06-29
Remove Elasticsearch dependency


d12:
Elasticsearch can be replaced by Postres full text search to remove a dependency that's caused availability issues in the past.

https://www.postgresql.org/docs/9.5/textsearch.html

d12:
Re-opening, #1983 removed our dependency on ElasticSearch but lets not close this until we've actually removed ES from our Gemfile, cleaned up any ES files/logic, and unprovisioned ES in Heroku.

d12:
As an update, I _believe_ all uses of ElasticSearch have been removed from Classroom. To close out this issue:

- [ ] Check ES provider for read/write activity. (We use a cloud provider through Heroku)
- [ ] Remove old ES logic in the app
- [ ] Remove ES related bootstrapping logic
- [ ] Remove ES gems
- [ ] Deprovision the service on Heroku
- [ ] Profit.

@andrewbredow Can I hand this off to you and your team?

shaunakpp:
I believe https://github.com/education/classroom/pull/2107 takes care of most of the above checkboxes

JessRudder:
I was going through the checklist and I'm fairly certain this is all done. @jeffrafter it looks like you removed elasticsearch from Heroku - if so (and you're comfortable that there's nothing else we need to do to remove ES), feel free to close this out!

PR #1983, opened 2019-07-11
Move Stafftools from ElasticSearch to PostgreSQL Full Text Search

d12:
Closes https://github.com/education/classroom/issues/1926

This PR _finally_ removes our dependency on ElasticSearch.

I'm curious to see how this performs on our production-sized database. We may need to add some lexeme indexes but since this is admin-facing, let's just deploy and find out how it performs.

We should also be using a better dictionary than "english". The english dictionary uses stopwords that make sense when looking at sentances and paragraphs, but not logins and slugs. We might be best off just defining out own dictionary, I just haven't dug into how to do this yet.

I'm not yanking our elasticsearch gem and other related files yet, I want to see how this performs first.


spec/models/assignment_spec.rb
```diff
@@ -9,6 +9,34 @@
...
+      expect(results.to_a).to include(searchable_assignment)
+    end
+
+    it "searches by slug" do
+      results = Assignment.search(searchable_assignment.slug)
+      expect(results.to_a).to include(searchable_assignment)
+    end
+
+    it "does not return the assignment when it shouldn't" do
+      results = Assignment.search("spaghetto")
```
EricPickup: 🍝

PR #1985, opened 2019-07-11
Implement Template Repositories in Classroom

EricPickup:
The moment we've been waiting for...

## What
This PR adds the option to import starter code using GitHub's new [Template Repository](https://github.blog/2019-06-06-generate-new-repositories-with-repository-templates/) feature.

This feature will launched be under a flag.

## Why
Our current method of importing starter code is through GitHub's source importer. This method is unfortunately very flaky and is the reason for a **huge** chunk of our reported bugs (import freezing, slow, or not working at all).

Template repos is a new feature shipped by GitHub this year which is much more reliable and fast.

## Images

Import options show/hide when starter code repo is filled/empty
nANJoc082t

Checking if a repo is a template repository
image

Accepting an assignment with starter code
2O47aDQlhV


app/controllers/application_controller/feature_flags_dependency.rb
```diff
@@ -28,4 +28,9 @@ def lti_launch_enabled?
     GitHubClassroom.flipper[:lti_launch].enabled?
   end
   helper_method :lti_launch_enabled?
+
+  def github_template_repos_enabled?
```
d12: How does this work? It looks like in the code, we're checking if template repos are enabled on an organization, not a user. We should probably always check on orgs and not users.

app/models/assignment_repo/creator.rb
```diff
@@ -132,6 +133,26 @@ def delete_github_repository(github_repo_id)
       true
     end
 
+    # Public: Clone the GitHub template repository for the AssignmentRepo.
+    #
+    # Returns an Integer ID or raises a Result::Error
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+    def clone_github_template_repository!
```
d12: This is confusing since `clone` means to pull a repository down from a remote, which isn't happening here. What about `create_repo_from_template!`?

app/models/assignment_repo/creator.rb
```diff
@@ -132,6 +133,26 @@ def delete_github_repository(github_repo_id)
...
+        name: repository_name,
+        owner: organization.github_organization.login,
+        private: assignment.private?,
+        description: "#{repository_name} created by GitHub Classroom",
+        accept: "application/vnd.github.baptiste-preview"
+      }
+
+      client.post("https://api.github.com/repositories/#{assignment.starter_code_repo_id}/generate", options)
+    rescue GitHub::Error => error
+      raise Result::Error.new REPOSITORY_CREATION_FAILED, error.message
```
d12: Can we put a statsd call in here?

app/models/concerns/starter_code_importable.rb
```diff
@@ -0,0 +1,44 @@
...
+  def starter_code_repository
+    return unless starter_code?
+    @starter_code_repository ||= GitHubRepository.new(creator.github_client, starter_code_repo_id)
+  end
+
+  def template_repos_enabled?
+    template_repos_enabled
+  end
+
+  def template_repos_disabled?
```
d12: We typically don't define methods like this in Ruby. Prefer `!template_repos_enabled?`

app/models/concerns/starter_code_importable.rb
```diff
@@ -0,0 +1,44 @@
...
+  def starter_code?
+    starter_code_repo_id.present?
+  end
+
+  def starter_code_repository
+    return unless starter_code?
+    @starter_code_repository ||= GitHubRepository.new(creator.github_client, starter_code_repo_id)
+  end
+
+  def template_repos_enabled?
```
d12: This can be an `alias`

thebradbain:
This is so exciting! 🎇 ✨

app/controllers/application_controller/feature_flags_dependency.rb
EricPickup: Good catch!

app/models/concerns/starter_code_importable.rb
```diff
@@ -0,0 +1,36 @@
...
+    return unless use_template_repos?
+
+    options = { accept: "application/vnd.github.baptiste-preview" }
+    endpoint_url = "https://api.github.com/repositories/#{starter_code_repo_id}"
+    starter_code_github_repository = creator.github_client.get(endpoint_url, options)
+
+    return if starter_code_github_repository.is_template
+    errors.add(
+      :starter_code_repository,
+      "is not a template repository. Make it a template repository to use template cloning."
```
d12: I'd like to point to documentation here, but this can be done in the future PR.

EricPickup: I was planning to do that, but wouldn't that require adding HTML markup to the error message to link to the docs? It's easy/doable but it might get ugly for tests

d12: I'd like to err on the side of UX over having all our tests be 100% pretty

app/assets/javascripts/assignment_form.js
```diff
@@ -0,0 +1,7 @@
+function import_options(starter_code_field_value) {
+  if (starter_code_field_value.length > 0) {
+    document.getElementById('import-options').style.display ="";
+  } else {
+    document.getElementById('import-options').style.display ="none";
```
cheshire137: A few style nitpicks: could use `'` or `"` consistently, and add a space after each `=`. What do you think about pulling out `const importOptions = document.getElementById('import-options');` to reduce the line length on lines 3 and 5?

app/assets/javascripts/assignment_form.js
```diff
@@ -0,0 +1,7 @@
+function import_options(starter_code_field_value) {
```
cheshire137: Is snake_case used elsewhere in the app? If not, could go with camelCase to match JavaScript style, for both the variable and the function name.

app/models/assignment.rb
```diff
@@ -37,8 +38,11 @@ class Assignment < ApplicationRecord
 
   validate :uniqueness_of_slug_across_organization
 
+  validate :starter_code_repository_is_a_template_repository
```
cheshire137: Could rename to `starter_code_repository_is_template` to shorten it, or even use `repo`.

app/models/assignment_repo/creator.rb
```diff
@@ -132,6 +134,26 @@ def delete_github_repository(github_repo_id)
...
+      repository_name = generate_github_repository_name
+      options = {
+        name: repository_name,
+        owner: organization.github_organization.login,
+        private: assignment.private?,
+        description: "#{repository_name} created by GitHub Classroom",
+        accept: "application/vnd.github.baptiste-preview"
+      }
+
+      client.post("https://api.github.com/repositories/#{assignment.starter_code_repo_id}/generate", options)
```
cheshire137: Could `https://api.github.com` be pulled out to a constant, like `GITHUB_API_HOST`?

app/assets/javascripts/assignment_form.js
EricPickup: Good catch! Looks like we've used both in our `assets/javascripts` but I'll switch to camelCase here.

app/models/assignment_repo/creator.rb
EricPickup: Good idea. I'll also move the preview accept header into a constant

EricPickup:
@BenEmdon The issue with the error dialogue box exists with all form fields unfortunately and is reported in https://github.com/education/classroom/issues/1988. I'll add a hotfix that removes it for this field on input but this will eventually need to be fixed properly.

db/schema.rb
```diff
@@ -196,6 +196,7 @@
     t.string "github_global_relay_id"
     t.bigint "organization_webhook_id"
     t.string "google_course_id"
+    t.boolean "archived"
```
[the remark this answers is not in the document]
EricPickup: I didn't, thanks for catching this 😅

app/views/group_assignments/_group_assignment_form_options.html.erb
```diff
@@ -43,10 +43,25 @@
 
 <dl class="<%= view.form_class_for(:starter_code_repository) %>">
   <dt><label>Add your starter code from GitHub <small>(optional)</small> </label></dt>
-  <dd><input class="textfield js-autocomplete-textfield form-control input-contrast input-block" type="text" name="repo_name" value="<%= @group_assignment.starter_code_repository.try(:full_name) %>" placeholder="<%= @group_assignment.starter_code_repository.try(:full_name) || 'Search repositories'%>" autocomplete="off" data-autocomplete-search-endpoint="github_repos"></dd>
+  <dd><input class="textfield js-autocomplete-textfield form-control input-contrast input-block" type="text" name="repo_name" value="<%= @group_assignment.starter_code_repository.try(:full_name) %>" placeholder="<%= @group_assignment.starter_code_repository.try(:full_name) || 'Search repositories'%>" autocomplete="off" data-autocomplete-search-endpoint="github_repos" oninput=<%= "importOptions(this.value);removeErrorBox();" if @organization.feature_enabled?(:template_repos) %>></dd>
```
BenEmdon: Nit: add spacing 👀 

```suggestion
  <dd><input class="textfield js-autocomplete-textfield form-control input-contrast input-block" type="text" name="repo_name" value="<%= @group_assignment.starter_code_repository.try(:full_name) %>" placeholder="<%= @group_assignment.starter_code_repository.try(:full_name) || 'Search repositories'%>" autocomplete="off" data-autocomplete-search-endpoint="github_repos" oninput=<%= "importOptions(this.value); removeErrorBox();" if @organization.feature_enabled?(:template_repos) %>></dd>
```

EricPickup: You actually can't have spacing between calls like this 😕

EricPickup:
@BenEmdon Sure! It hides the error box on input in this field.
Jpis5s5rIM

app/views/group_assignments/_group_assignment_form_options.html.erb
BenEmdon: Oh my bad!

PR #1987, opened 2019-07-11
[SPIKE] Roster Import from LMS

thebradbain:
## What
This is a spike to demonstrate the feasibility of importing the roster from an LMS into GitHub classroom, but is in no way meant to be merged in its current form (or in one PR).  It makes for a fun gif, though! 

In a nutshell -- once the LMS is already configured to talk with Classroom and the user is authenticated, the process is straightforward! To see how everything comes together, check out `LtiConfiguration#import`. 

Al7tXFZTbw


app/services/lti_membership_service.rb
```diff
@@ -0,0 +1,66 @@
...
+      symbolized_membership[:member] = parsed_member
+
+      parsed_membership = IMS::LTI::Models::MembershipService::Membership.new(symbolized_membership)
+      memberships_list.push(parsed_membership)
+    end
+
+    memberships_list
+  end
+
+  def make_signed_membership_request(endpoint)
```
thebradbain: I'm not a fan of having to manually sign any outgoing requests, but the LTI standard requires a little-used extension of OAuth 1.0 called the `oauth_body_hash` to be present when generating the signature, and I don't know of any mainstream OAuth libraries makes it as simple as a boolean -- so I used ruby's [go-to OAuth library](https://rubygems.org/gems/oauth/versions/0.5.1) to generate it.

This method should probably be abstracted to a generic `LtiRequestService` class so this method can be reused for future lti services, too.

app/services/lti_membership_service.rb
```diff
@@ -0,0 +1,66 @@
+class LtiMembershipService
```
thebradbain: Another big question I have that would be helpful in proceeding: do you think this be a service, or should we make it a controller concern? 

Right now, aside from `LtiMembershipService`, we have
-  `LTI OmniAuth strategy` - for launch/auth handshake
- `GitHubClassroom::LtiMessageStore`  - for abstracting launch message redis persistence/retrieval 

And, if we do decide to keep `LtiMembershipService` as a service, should I create an umbrella `GitHubClassroom::LTI` umbrella module that would hold LtiMembershipService, LtiMessageStore, and any other Lti services we may decide to create in the future?

d12:
This looks super cool! We'd need some sort of direction on our side so that teachers know what to do. How do they install the app on their LMS side, for e.g.?

thebradbain:
Great question! As a start, some of those instructions are visible to the user on #1990 and hopefully that process is made easier/smoother in #1998, though we'll probably want to revise and be very intentional with our UI, as all the various LMSs don't make the configuration process user-friendly for the uninitiated (and the LTI standard isn't too prescriptive when it comes to configuration, unfortunately). 

We'll also definitely want to write support docs separate the UI.

PR #1989, opened 2019-07-11
Prevent teachers from using empty repos as starter code repositories

### LINKED ISSUE #1906: Starter code importer hangs when importing from empty repository

EricPickup:
## What
#1931 aimed to fix this but we ran into edge cases when using the number of commits to determine if a repo was empty. We also tried using the `size` attribute which was unreliable, described in #1970.

Fixes #1906, which causes the importer to hang when importing from an empty repo.

This PR adds a validator to Assignment and GroupAssignment that checks if their starter code repo is empty. It will prevent teachers from saving/creating an assignment with an empty repo selected.

## How
This time we'll be checking the repository's [`Contents`](https://www.rubydoc.info/github/pengwynn/octokit/Octokit/Client/Contents#contents-instance_method), which lists the actual file contents of the repo. When the repository contains no content, the GitHub api returns an error stating that the repo is empty.

ex:
Empty repo
https://api.github.com/repos/classroomtestorg/daasdsadsadasdasd-ericpickup-test3/contents

Populated repo
https://api.github.com/repos/education/classroom/contents


app/models/github_repository.rb
```diff
@@ -161,7 +161,9 @@ def number_of_commits
   end
 
   def empty?
```
EricPickup: Reviewers - this method is the only thing changed from the original PR

srinjoym: Could we use the `GitHub::Errors` class here to follow the pattern in the other methods in the file? Not sure if there's anything different about this check but it could possibly look like this:

```
  def empty?
      @client.contents(full_name).empty?
    end
  rescue GitHub::Error
    return true
  end
```

d12: :+1:

EricPickup: @srinjoym great suggestion!

PR #1990, opened 2019-07-11
Implement LtiConfigurations Controller

stephaniegiang:
## What
Creating the `#create`, `#edit`, and `#delete` actions that will connect with the front-end. Also moved `#info` back to `#new`

### Creating Configuration
bfb1G5YxUQ

### Updating Configuration
jo08w3M9cC

### Deleting Configuration
JmTprVfdQV


thebradbain:
Based on @BenEmdon's comment on this PR, https://github.com/education/classroom/pull/1935#discussion_r301831432, I think it would be best if we rename the `info` action back to `new`, just because right now we're redirecting to `info` if an `LtiConfiguration` doesn't exist when in practice we do really mean `new`

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -3,27 +3,59 @@
...
+
     private
 
     def current_lti_configuration
       @current_lti_configuration ||= current_organization.lti_configuration
     end
 
     def ensure_current_lti_configuration
-      redirect_to info_lti_configuration_path(current_organization) if current_lti_configuration.nil?
+      redirect_to new_lti_configuration_path(current_organization) if current_lti_configuration.nil?
```
d12:
```suggestion
      redirect_to new_lti_configuration_path(current_organization) unless current_lti_configuration
```

thebradbain:
🎉 ✨

PR #1992, opened 2019-07-12
Consolidate LTI Services

thebradbain:
## What
Proactively restructures LTI services / utility classes. 

Right now, regarding general LTI functionality not-intrinsically tied to a specific controller, we have:

- `LTI OmniAuth strategy` - for launch/auth handshake
- `GitHubClassroom::LtiMessageStore` - for abstracting launch message persistence/retrieval of a launch message for duration of the session

Looking at #1987, it looks like we'll also be creating something akin to a `LtiMembershipService` to 
fetch a roster for a given LMS class. I'd imagine further down the line when we implement fetching/submitting assignments to the LMS we'd add another service. 

This PR moves the freestanding `LtiMessageStore` utility class under the new `GitHubClassroom::LTI` umbrella module (though I'm also fine with separating it entirely and moving it into a separate `LTI` module, if we wish), so in the future it's clear where more LTI-specific api functionality will fall under. Feedback is much appreciated!


lib/github_classroom/lti/message_store.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+
+      def message_valid?(lti_message)
+        # check for duplicate nonce
+        return false if nonce_exists?(lti_message.oauth_nonce)
+
+        # check if nonce too old
+        return false if DateTime.strptime(lti_message.oauth_timestamp, "%s") < 5.minutes.ago
+
+        true
```
BenEmdon: I like the _happy path_ you have going on here 🙌

lib/github_classroom/lti/message_store.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module GitHubClassroom
+  module LTI
+    class MessageStore
+      attr_reader :consumer_key
+
+      def initialize(consumer_key: nil, redis_store: nil)
+        raise(ArgumentError, "consumer_key may not be nil") if consumer_key.blank?
+        raise(ArgumentError, "redis_store may not be nil") if redis_store.blank?
```
BenEmdon: Nice validation 💥

config/initializers/lti_message_store.rb
```diff
@@ -2,11 +2,16 @@
 
 module GitHubClassroom
   def self.lti_message_store(options = {})
+    if options[:lti_configuration].present?
+      options[:consumer_key] = options[:lti_configuration].consumer_key
+      options.delete(:lti_configuration)
+    end
```
BenEmdon: Does this need a test?

thebradbain: Great question! Meant to highlight this when I submitted the PR. I didn't see any tests for initializers within the specs, is there a specific place I should create a test for this?

BenEmdon: You should create a test at `spec/config/initializers/lti_message_store_spec.rb`

BenEmdon: Since this is an initializer I suppose it doesn't need a test 🤷‍♂

PR #1998, opened 2019-07-13
Implement LTI's auto-configure LMS capabilities

### LINKED ISSUE #1994: Implement LTI's Auto-configure Application capabilities
LTI has the ability to semi-automatically configure Classroom on an instructor's LMS such that they would only have to provide their `consumer_key`/`public_key` and a link to our application to configure the app on their LMS.

It's a non-required capability for LMSs to implement, but it will make it a lot easier for the instructors to integrate Classroom on LMSs which do

thebradbain:
## What
Closes #1994. 

Configuring an LTI Application on the LMS-side consists of a lot of moving parts that may confuse instructors, so most LMSs provide functionality to make that process smoother by allowing LTI Tools to auto-configure themselves on the LMS via generating XML. This XML essentially tells the LMS which users can access the application, where the application shows up on the UI, declares intent/permissions and capabilities, etc.

This PR will enable, on LMSs which support it, an instructor to completely configure Classroom in their LMS by only providing the generated consumer key, shared secret, and auto-configuration url.

demo configure gif


app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -5,6 +5,9 @@ class LtiConfigurationsController < Orgs::Controller
     before_action :ensure_lti_launch_flipper_is_enabled
     before_action :ensure_current_lti_configuration, except: %i[new create]
 
+    skip_before_action :authenticate_user!, only: :autoconfigure
+    skip_before_action :ensure_current_organization_visible_to_current_user, only: :autoconfigure
```
thebradbain: I'm wondering if `#autoconfigure` makes sense within the `#lti_configurations_controller` mainly because whenever the LMS makes its initial request to Classroom to generate a configuration file it will be a plain old non-authenticated GET request, meaning the general assumptions around a subclass of `Orgs::Controller` don't fully apply. 

My followup question would be: If not in this controller, where?

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -44,6 +47,39 @@ def destroy
...
+        vendor_name: "GitHub Classroom",
+        vendor_url: "https://classroom.github.com/"
+      )
+
+      ## LMS Specific Attributes ##
+      # Note: LMS's will ignore vendor identifiers they do not understand
+
+      # Canvas will not display an LTI application
+      # unless specify a course_navigation location.
+      builder.add_vendor_attributes(
```
thebradbain: Do we think refactoring the meat of this entire method into a controller concern is worth consideration? It's already pretty long as it is, but I could easily imagine a future in which we have custom attributes for all of our most popular LMSs to ensure the best experience possible, and in that case this controller becomes a glorified config file 😬

d12: I think this is fine for now :+1:

PR #2009, opened 2019-07-15
Fix "lti_launch" flag

stephaniegiang:
## What
This fixes a bug that I found. The flag was checking to see if Classroom had it the `lti_launch` flag enabled. This is because we want the teacher to launch this outside of GitHub Clasroom, as they don't always have to be logged in. But this resulted to staff not being able to see the `LMS configuration` tab.

In this PR, we are also checking to see if the user has the flag enabled as well.


app/controllers/application_controller/feature_flags_dependency.rb
```diff
@@ -34,7 +34,7 @@ def search_assignments_enabled?
   helper_method :search_assignments_enabled?
 
   def lti_launch_enabled?
-    GitHubClassroom.flipper[:lti_launch].enabled?
+    GitHubClassroom.flipper[:lti_launch].enabled? || (logged_in? && current_user.feature_enabled?(:lti_launch))
```
BenEmdon: When the `:lti_launch` is enabled `current_user.feature_enabled?(:lti_launch)` will return true. Heres a suggestion to fix that redundancy: 
```suggestion
    logged_in? && current_user.feature_enabled?(:lti_launch)
```

stephaniegiang: We also want this feature to be able to launch when the user isn't logged in. It's kinda a weird flow with LTI integration. So the user is able the launch Classroom from the respective LMS and that will lead them to log in. The addition I'm adding is to check once the user is logged in. @thebradbain can explain more if you need more information.

thebradbain: Linking https://github.com/education/classroom/pull/1935#discussion_r301357639 for context, but specifically: 

> It's important that possibly-unauthenticated users still be able to launch into GitHub Classroom from their LMS and then be required log in normally if they are not already, and we definitely want to be able to test that this is indeed what happens (otherwise we have no idea who the LTI launch data belongs to).

But it's also important that we can only test out this integration in production for staff members without enabling it for everyone, so hence the awkward semi-redundant check

BenEmdon: Thanks for the context 👍 This looks fine then! Dismiss my suggestion

PR #2011, opened 2019-07-15
Backfill for cached github_user fields

d12:
Successor to https://github.com/education/classroom/pull/1964

This backfills `github_` fields on the User model which will allow us to begin using these columns for sort/search.

This will be a super long running rake task, and will probably have to be run over the course of many evenings (and then paused in the day while engineers are deploying). It's written to be re-runnable and resumable, it'll just pick up where it left off if cancelled and re-run.

cc https://github.com/education/classroom/pull/1982

I gave it a go locally with BATCH_SIZE = 1:

```
✗ bundle exec rake backfill_github_cached_fields_for_user
Backfilling GitHub-cached fields for Users
There are 3 users to update.
Updating batch 1, starting with user ID 1
Updating batch 2, starting with user ID 2
Updating batch 3, starting with user ID 3
```


lib/tasks/backfill_github_cached_fields_for_user.rake
```diff
@@ -0,0 +1,26 @@
...
+  batch_num = 1
+
+  users_to_update = User.where(github_login: nil)
+  puts "There are #{users_to_update.count} users to update."
+
+  users_to_update.find_in_batches(batch_size: BATCH_SIZE) do |users|
+    puts "Updating batch #{batch_num}, starting with user ID #{users.first.id}"
+
+    users_updated += users.length
+    batch_num += 1
```
EricPickup: Shouldn't we update these numbers _after_ the logins are fetched? If anything went wrong in our user loop these numbers would become inaccurate.

EricPickup: Shouldn't we update these numbers _after_ the logins are fetched? If anything went wrong in our user loop these numbers would become inaccurate.

PR #2013, opened 2019-07-15
Create LTI::MembershipService

thebradbain:
## What
Refactors the `MembershipService` previously seen in the spike #1987. 

Allows for easy and robust retrieval of a course roster from an endpoint provided by a Learning Management system through LTI, and handles any request signing, response parsing, and possible error handling/raising which may occur.


lib/github_classroom/lti/request_signing.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      def signed_request(endpoint, consumer_key, secret, headers: {})
+        headers = { "Authorization": signed_auth_header(endpoint, consumer_key, secret) }.merge(headers)
+        opts = {
+          url: endpoint,
+          headers: headers
+        }
+
+        Faraday.new(opts) do |conn|
+          conn.response :raise_error
```
thebradbain: Faraday has error handling for requests/response built in as a [middleware](https://lostisland.github.io/faraday/middleware/list), so we may as well use that for error handling rather than re-invent the wheel

lib/github_classroom/lti/membership_service.rb
```diff
@@ -0,0 +1,63 @@
...
+      def filtered_membership(roles: nil)
+        unfiltered_membership = membership
+        return unfiltered_membership unless roles
+
+        unfiltered_membership.select do |membership|
+          # LMS is allowed to prefix a role with anything it wants
+          # so the filter method simply matches on substrings
+          roles.any? do |desired_role|
+            membership.role.any? do |actual_role|
+              actual_role.include? desired_role
```
d12: What will `roles` and `unfiltered_membership` look like here?  We have 4 levels of iteration here, can we improve this at all? If these are long lists, this method could be extremely slow.

lib/github_classroom/lti/request_signing.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module GitHubClassroom
```
d12: Are you able to write any tests for this Module? It may be possible by stubbing out the network requests.

lib/github_classroom/lti/membership_service.rb
thebradbain: I agree that this method is... unideal :grimace:

`Roles` is a list of strings. LTI defines 8 standard roles (Student, Admin, Learner, Instructor, etc) that may or may not have custom prefixes, but also allows an LMS to define any custom roles they want. In my limited experience testing and researching I haven’t seen a list of roles longer than 2 elements.

`unfiltered_membership` is just a collection of all the members associated with a given course, so I’d imagine depending on the university it could get pretty large, but not more than a few hundred?

d12: Thats awesome 😄 Please ping me again when this is ready for review again!

lib/github_classroom/lti/request_signing.rb
thebradbain: I've tried multiple approaches to writing tests for this module, and it seems that every time I'm essentially re-writing implementation of this mixin itself to test agains itself, which doesn't make too much sense.

No network requests are actually necessary so there's nothing to mock, as validating the signature only requires a consumer key, secret, request body, url, oauth_body_hash, and timestamp. 

Perhaps the quasi-integration tests of the services which rely on this mixin suffice?

thebradbain:
@d12 ready for a second round of review!

PR #2018, opened 2019-07-16
Limit teachers to only one integration 

### LINKED ISSUE #1995: Only allow one integration
With Google Classroom and LMS integration being shipped around the same time, we want to make sure we are not allowing users to have more than one integration.

Requirements
- User cannot create another integration if one exists
- If user wants to connection to a different integration, they have to delete existing one beforehand

stephaniegiang:
## What
Closes #1995 
This limits the user to only one integration.


app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Orgs
   class LtiConfigurationsController < Orgs::Controller
     before_action :ensure_lti_launch_flipper_is_enabled
+    before_action :ensure_no_google_classroom
```
thebradbain: Should this be only before the `#create` action?

PR #2021, opened 2019-07-16
Ensure there is no existing roster when adding integration

stephaniegiang:
## What
This PR makes sure that for the Google Classroom and LMS integration, we check to see if there is an existing roster. 

If there is, we display an alert and tell the user to delete their roster before proceeding.

If there is none, we continue with the integration process.


app/controllers/orgs/rosters_controller.rb
```diff
@@ -398,6 +401,12 @@ def user_google_classroom_credentials
     def redirect_if_roster_exists
       redirect_to roster_url(current_organization) if current_organization.roster.present?
     end
+
+    def google_classroom_ensure_no_roster
+      return unless current_organization.roster
+      redirect_to edit_organization_path(current_organization),
+        alert: "Google Classroom connection failed as you already have a roster. Please delete roster and try again."
```
thebradbain: This is _very_ pedantic, but maybe we should present this alert in a friendlier manner. Maybe something like


> "We are unable to link your organization to Google Classroom because a roster already exists for this classroom organization. Please delete your current roster and try again."

app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -93,5 +94,11 @@ def ensure_current_lti_configuration
     def lti_configuration_params
       params.require(:lti_configuration).permit(:lms_link)
     end
+
+    def ensure_no_roster
+      return unless current_organization.roster
+      redirect_to edit_organization_path(current_organization),
+        alert: "LMS configuration failed as you already have a roster. Please delete roster and try again."
```
thebradbain: See related comment for GClassroom

app/controllers/orgs/rosters_controller.rb
stephaniegiang: Yeah I had a linter issue with how long that message was so I decided to shorten it. I can change it back.

app/controllers/orgs/lti_configurations_controller.rb
stephaniegiang: I don't think we need to. Since we define what LMS is on the `#new` page is

Issue #2019, opened 2019-07-16
Assignment sort default to student identifier


stephaniegiang:
There was an issue presented in https://github.com/education/classroom/pull/1889

The default for assignment sort is `created at` but we should default to `student identifier` instead.

PR #2022, opened 2019-07-16
Student identifier default for sort

stephaniegiang:
## What
Closes https://github.com/education/classroom/issues/2019

This PR makes the student identifier the default. 

I didn't see any tests for this, but correct me if I'm wrong @thebradbain


app/controllers/assignments_controller.rb
```diff
@@ -132,7 +132,7 @@ def set_assignment
   def set_filter_options
     @assignment_sort_modes = RosterEntry.sort_modes
 
-    @current_sort_mode = params[:sort_assignment_repos_by] || @assignment_sort_modes.keys.first
+    @current_sort_mode = params[:sort_assignment_repos_by] || @assignment_sort_modes.keys.second
```
thebradbain: Instead of relying on the order of the keys in a hash (my sin originally), maybe we should explicitly write `"Student identifier"` in place of `@assignment_sort_modes.keys.second` ?

stephaniegiang: I do like that idea. It does make the code easier to read. But I also don't know if I like having the hard coded string there.

d12: Sorting on the `second` key seems pretty arbitrary, + this will error if we ever add a sort with only 1 option (which is weird, but it's a step to 2 or more options!). 

Relying on the `first` key is a reasonable default I think, so why don't we change the order of the sort modes https://github.com/education/classroom/blob/1e2a4f2fc944a4338eda33b22b66180ba53a6f3d/app/models/roster_entry.rb#L33 here instead? Then we can leave this code alone and we won't swap the ordering of every single orderable thing with this PR.

PR #2028, opened 2019-07-17
Add archived column to Organization

EricPickup:
## What
Related to #602 and #1996.

Adds the `Archived` (boolean) column to `Organization` which will eventually be used to determine if a classroom is archived.

Issue #1996, opened 2019-07-12
Classroom After-life Project


EricPickup:
Related to #602, #1976, and #438 

Many teachers have requested the ability to archive/hide old classrooms. Currently their only option is to delete the classroom and all of its repos entirely, or leave the dead classroom in their dashboard which makes it difficult to find their active classrooms.

There's many ways we can handle the after-life of a classroom.

Teachers may want to archive the classroom and the repositories with it.
Teachers may want to archive the classroom and transfer the ownership of the repos to the students.
Teachers might want to publicize the assignment repos after the class is completed.

Steps:
- [ ] Add ability to archive classroom (hide or distinguish archives classrooms in list)
- [ ] Add ability to archive all assignment repos through github API (prevents any further commits/changes to repo)
- [ ] Add ability to transfer ownership of assignment repos to the students (would only work for individual assignments)
- [ ] Add ability to publicize all assignment repos

EricPickup:
Quick note: 
- We can archive repos through the api however we **cannot** un-archive them through the github api. If teachers archive all of the student repos it cannot be undone through classroom.

I'm going to start on the first step (adding ability to hide/archive classroom in list). @d12 what do you think about the rest of these goals?

d12:
This is great! Thanks for writing this up. The main thing I'd like to change is to move that step 2 `Add ability to archive all assignment repos through github API` to the very end. I'd say it's the least important and least requested in the list.

When archiving Classrooms, let's do it in a way where we don't delete anything, we just move the classes to a different "archived classrooms" view or tab.

When doing the transfers/publicizing/archiving of repos, we need to make sure this is written in a retryable, idempotent way. Also, all 3 are essentially the same, just a different API endpoint. We could probably implement all 3 with mostly the same code.

Issue #602, opened 2016-05-25
Archiving a Classroom


tabinks:
Add functionality to "archive" a Classroom. For example, once CS101-Autumn is completed, provide some way to lock the repos and potentially hide them (or provide some other visual cue that the course is over).  

This request may be related to Issue #595 or a potential solution to #595 where an organization can only have a single classroom (the current behavior) but you can archive the previous classroom and create a new one for that organization.

Potential workflow:
- I have an organization named _BigU-CS101_.
- I create a classroom for _BigU-CS101_ named _Fall-2015_.
- At the end of the quarter, I archive _Fall-2015_.  This locks it and prevents any new repos being created.
- I create a classroom for _BigU-CS101_ named _Winter-2016_.

At any given time, I only have a single active Classroom for the organization.

johndbritton:
Thanks for opening this @tabinks.

@NickTikhonov This is something you might want to consider as you're thinking about how the end of an assignment / course works.

robwhess:
I think something like this is a great idea.  I've been using Classroom for a while now, and I have lots of individual classrooms.  Since the list of classrooms doesn't appear to be organized in any way I can discern, it's getting hard for me to locate my current classrooms within the larger list.  Even just having that list split into "active" classrooms and "inactive" classrooms (where I can change a setting to mark a classroom as inactive) would be useful for me.

Locking down each inactive classroom so that no new repos can be created would be a bonus.  I currently just disable all of the invitation links for a given classroom.  While this takes a little effort to iterate through all of the assignments, it works fine.

robwhess:
As an added note: I use a separate GitHub organization for each course I teach.  This works well for me.  My main hope is to be able to mark classrooms as inactive.

sfogarty:
Here is a VERY hacky python script, with no error testing or anything of the sort. Use at your own risk.

[archive_classroom.txt](https://github.com/education/classroom/files/2078437/archive_classroom.txt)

sfogarty:
Also wonderful would be the ability to archive the organization, so that it did not clutter my drop-downs on github.com.

phannan:
This would be great.  I teach 2 courses a term.  I have one organization per class per term.  I'm starting to get a long list and came over here to add an issue to sort the classrooms newest first.  Archiving would be even better.

stevenbitner:
[...]

PR #2030, opened 2019-07-17
Add archive button to classroom settings

EricPickup:
## What
Related to #602 and #1996. Migration separated in #2028.

This PR adds a button to the classroom settings page that will toggle whether the classroom is archived.

This will change the `archived_at` attribute in `Organization`.

image

image


d12:
Nice, this is super simple! 🤘 

The way this danger zone was designed in the past isn't looking too great with multiple buttons. What do you think about layout out the text/buttons similarly to how we do it on GitHub.com?

Screen Shot 2019-07-18 at 9 18 00 AM

d12:
Also can we specify that this won't _delete_ anything?

EricPickup:
@d12 I've updated it with the GitHub.com style and added separate titles/descriptions for archived vs unarchived classrooms.
image

image

I also moved this button under a feature flag since it doesn't actually do anything user-facing yet.

Issue #2029, opened 2019-07-17
Ensure LMS User Flow consistent with Google Classroom


thebradbain:
**Action Items**
 
- [x] Move initial LMS Configuration setup flow into the Classroom Profile (Issue #2050)
- [x] When creating a new classroom, ask for the user about integration at the beginning
- [x] Remove dedicated LMS Configuration page from sidebar 
- [x] Refer to the LMS by its official platform name on the front-end (e.g. Canvas, Blackboard, Moodle) rather than generic LMS/Learning Management System (Issue #2057, PR #2056, PR #2071)

**Further Context**

After speaking with @brandonrosage, we ironed out some details on where the configuration of LMS should live, and how it should be treated alongside Google Classroom integration from the user's perspective.

In short, we should ensure the setup process is the same and common functionality is consistent and accessible in the same ways.

See #2027 for further context re: Google Classroom

PR #2033, opened 2019-07-18
LMS Roster Import

thebradbain:
## What
✨ Allows importing a class roster from any LMS which implements the LTI standard ✨ 

The backend changes are relatively minor, and focus on connecting previously existing LTI functionality into the relevant controller actions. Also includes UI changes to make the roster import flow similar to that of Google Classroom, but is not meant to fulfill #2029 so much as be a launching point.

_Call to action on new roster page and corresponding redirect if LMS is not configured_
Call to action

_Selecting how to identify students and successful initial import_
Successful first sync

_Syncing new students from an LMS on an already existing roster_
Sync

_User facing error messages when GitHub Classroom is not properly configured in the LMS_
Not Configured Error

Closes #1987


app/models/lti_configuration.rb
```diff
@@ -7,4 +7,19 @@ def self.find_by_auth_hash(hash)
     consumer_key = hash.credentials.token
     find_by(consumer_key: consumer_key)
   end
+
+  def context_membership_url(use_cache: true, nonce: nil)
```
thebradbain: I took heavy inspiration for this from #1964, in that it will use the persisted attribute if it exists or fallback on the LTI launch information if it doesn't and update the model accordingly (returning `nil` in the sad path that it exists on neither postgres or the lti launch context).

d12:
I love how small this PR is given what it's adding, nice work :) Would you mind adding tests for the new logic you've introduced?

app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module Orgs
+  class RostersController
+    class LtiImportError < StandardError; end
```
stephaniegiang: Love this structure ❤️

app/controllers/orgs/rosters_controller.rb
```diff
@@ -38,9 +38,10 @@ class RostersController < Orgs::Controller
     before_action :google_classroom_ensure_no_roster, only: %i[
       select_google_classroom
     ]
-
     helper_method :current_roster, :unlinked_users, :authorize_google_classroom
 
+    depends_on :lti
```
thebradbain: In the interest of keeping `rosters_controller` more or less untouched and decrease mental overhead, I used the awesome `depends_on:` controller helper we use in a few other large controllers -- it essentially splits up the controller, kind of like a mixin, and allows us to compartmentalize controller logic

app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -0,0 +1,73 @@
...
+      lti_configuration = current_organization.lti_configuration
+      redirect_to new_lti_configuration_path(current_organization) unless lti_configuration
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
+    def lms_membership
+      membership_service_url = current_organization.lti_configuration.context_membership_url(nonce: session[:lti_nonce])
+      unless membership_service_url
+        msg = "GitHub Classroom is not configured properly on your Learning Management System.
```
stephaniegiang: Can we test specifically for these messages?

app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -0,0 +1,73 @@
...
+      membership_service = GitHubClassroom::LTI::MembershipService.new(
+        membership_service_url,
+        current_organization.lti_configuration.consumer_key,
+        current_organization.lti_configuration.shared_secret
+      )
+
+      begin
+        membership_service.students
+      rescue
+        msg = "GitHub Classroom is unable to fetch membership from your Learning Management System at this time.
```
stephaniegiang: Same as above.

app/controllers/orgs/rosters_controller/lti_dependency.rb
thebradbain: This method itself is private and so should be covered by the controller tests regarding `#import_lms_roster` (in fact we even stub to simulate message failures), and we already do test `LTI::MembershipService` thoroughly on its own (pr #2033), so I don't know if there's much to test here that hasn't been already.

app/controllers/orgs/rosters_controller/lti_dependency.rb
thebradbain: I'm so happy we both discovered it. For another time, but GClassroom's integration with RostersController could probably easily be refactored into a similar dependency.

stephaniegiang: Yes! Once we get the UI changes in.

app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      respond_to do |format|
+        format.js
+        format.html
+      end
+    end
+
+    private
+
+    def ensure_lti_configuration
```
d12: This method can just be 

```ruby
     redirect_to new_lti_configuration_path(current_organization) unless current_organization.lti_configuration
```

app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      membership_service = GitHubClassroom::LTI::MembershipService.new(
+        membership_service_url,
+        current_organization.lti_configuration.consumer_key,
+        current_organization.lti_configuration.shared_secret
+      )
+
+      begin
+        membership_service.students
+      rescue
```
d12: Can you rescue specific error classes here? We usually try to avoid general rescues like this

app/models/lti_configuration.rb
```diff
@@ -7,4 +7,19 @@ def self.find_by_auth_hash(hash)
...
+    return cached_value if cached_value
+
+    message_store = GitHubClassroom.lti_message_store(consumer_key: consumer_key)
+    message = message_store.get_message(nonce)
+    return nil unless message
+
+    membership_url = message.custom_params["custom_context_memberships_url"]
+    return nil unless membership_url
+
+    self[:context_membership_url] = membership_url
```
d12: Does this actually save anything?

app/views/orgs/rosters/show.html.erb
```diff
@@ -12,8 +12,12 @@
             Classroom roster
           </span>
           <div class='flex-justify-between'>
-            <span class="btn btn-primary btn-sm", role="button" data-remodal-target="new-student-modal">Update students</span>
-             <span class="btn btn-primary btn-sm", role="button" data-remodal-target="download-csv-modal">Download</span>
+            <% if current_organization.lti_configuration %>
+              <%= link_to "Sync from LMS", import_from_lms_roster_path, remote: true, method: :get, class: "btn btn-primary btn-sm" %>
+            <% else %>
+              <span class="btn btn-primary btn-sm", role="button" data-remodal-target="new-student-modal">Update students</span>
```
d12: With this, teachers won't be able to manually add students to their roster when they have an LTI tool configured. That _could_ be a design decision we're OK with for now, but what about in the future when someone might link an LTI tool for something other than roster support? (E.g. they want to link assignments with Moodle, but don't have rosters set up there). Now they can't use the roster feature 😢 

This isn't a problem yet, and I think while roster support is the only LTI feature available, this actually makes a lot of sense. We'll need to do some investigation in the future though to figure out how to handle this case gracefully.

spec/controllers/orgs/rosters_controller_spec.rb
```diff
@@ -209,6 +209,104 @@
...
+            end
+
+            it "format.html: presents an error message to the user" do
+              get :import_from_lms, params: { id: lti_configuration.organization.slug }
+              expect(flash[:alert]).to be_present
+            end
+
+            it "format.js: presents an error message to the user" do
+              get :import_from_lms, format: :js, xhr: true, params: { id: lti_configuration.organization.slug }
+              expect(response.status).to be(428)
```
d12: 422 is preferred when missing a required parameter, 428 is for something different.

app/views/orgs/rosters/show.html.erb
thebradbain: Yeah, I totally agree this is something worth greater discussion! 

From a technical perspective, there currently is no difference between rosters inputted manually and imported from an LMS, so it wouldn’t cause any technical problems to even allow teachers to mix and match unrestricted (though we probably want to ensure consistency with Google Classroom). 

In the future, for assignment or submission syncing, the flows will ultimately depend upon the GitHub student whose assignment it is to be identifiable to a LMS’s `user_id`, but we don’t have store that piece of information on the roster (which we currently do for Google Classroom) and (I’d actually prefer, to avoid making the roster something different than originally designed) store that information separately on the user scoped to the linked classroom org.

In the meantime, do we want instructors to add students manually in addition to being able to sync, or do we want to leave as is?

app/models/lti_configuration.rb
thebradbain: It appears not, good catch! 😅 And also a good reminder that I haven’t written tests for this model yet.

app/views/orgs/rosters/show.html.erb
stephaniegiang: Oh that is a good point @d12. Hmm I think for now, we shouldn't allow teachers to manually input new students. Because that will leave us with some weird data.

thebradbain: To clarify: from an LMS perspective, we currently do the act of adding students to the roster via the standard #add_students action that manual entry also uses on the backend, so there is nothing technical keeping us from allowing them to mix. For Google Classroom, we tie those roster entries to a google client id, and all (or none) of those roster entries are either linked to Google or not, and mixing would produce inconsistent data.

d12: I'd like to strive for consistency between the two features whenever possible. Can we disallow manually adding students when either LTI or Google is enabled?

thebradbain: For LMS import specifically, do you mean:

1. If an LTI Configuration exists for the classroom, and a roster hasn't yet even been created, we only allow a teacher to create a roster via import from LMS

OR

2. If an LTI Configuration exists for the classroom, and a roster hasn't yet even been created, we allow the teacher to choose between manual entry or LMS import, and depending on what they choose means that method going forward will be the only way to add to their existing roster?

stephaniegiang: I like number 1 the best as it makes more sense. If we allow them to manually input, then they can technically also sync from LMS. Which doesn't really make sense, since I added the check to make sure no other configuration and roster currently exists.

d12: :+1:, # 1 sounds good. We'll need to tackle this in the future as we add more integration features, but in the mean time I think this is the most simple way to implement.

thebradbain: #1 it is, then! 

I also just want to clarify that option #2 would avoid mixing manual/imported entries. We would just be letting the user choose whether they want a manual or imported entries upon first sync, which would cover the scenario @d12 first described and also keep roster entries from intermingling, as @stephaniegiang brought up.

thebradbain:
@d12 should be ready for another round! I've incorporated the feedback from that large discussion, too

PR #2034, opened 2019-07-18
Add Barnes gem for beta Ruby metrics in Heroku

tarebyte:
See https://devcenter.heroku.com/articles/language-runtime-metrics-ruby for more context.

This adds the Barnes gem, which will allow our application to send metrics to Heroku about Ruby usage so that we can gain better insight into the system.


tarebyte:
Screen Shot 2019-07-19 at 13 12 38

Issue #2027, opened 2019-07-17
Move configuration of Google Classroom to a higher level 


stephaniegiang:
## TO-DO

- [x] Move the authorization of Google Classroom into the application controller
- [x] When creating a new classroom, ask for the user for the integration at the beginning
- [x] With existing classroom, add a button to `Classroom profile` tab/page so that they can connect Google Classroom

## CONTEXT
After speaking with @brandonrosage, we ironed out some details on where the configuration of Google Classroom should live.

Currently Google Classroom only interacts with the roster, but in the future we would like to expand on the functionality of what the integration can do.

So to set us up for that, we decided that we should move the interaction to the classroom level. 

When creating a new classroom, we should ask the user to choose their integration. This way we can fill in the same of the classroom to match the one on their Google Classroom. It will be on this page (presented below) and above the text field. 

![image](https://user-images.githubusercontent.com/22784140/61408446-e1e4b880-a894-11e9-93dd-4e6a5c5bde9e.png)

With an existing classroom, we would present it on this page. 

![image](https://user-images.githubusercontent.com/22784140/61408322-a21dd100-a894-11e9-936b-ff815b5ee4b6.png)

thebradbain:
And it appears #2023 will be partly taken care of by this, too! 🥳

stephaniegiang:
This is done in https://github.com/education/classroom/pull/2038

PR #2038, opened 2019-07-18
Move google auth to application controller

stephaniegiang:
## What
Part 1 of https://github.com/education/classroom/issues/2027

Moves Google authentication into `ApplicationController`


app/controllers/application_controller.rb
```diff
@@ -45,8 +45,30 @@ def failbot_context
     }
   end
 
+  def authorize_google_classroom
```
stephaniegiang: with `app/controllers/application_controller/google_authentication_dependency.rb` not working, I just added the methods to the actual controller. This seemed to fix the issue I was having with the file.

app/controllers/application_controller/google_authentication_dependency.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ApplicationController
```
stephaniegiang: So I tried making a new dependency to add it to the `ApplicationController` but this didn't really work. The methods weren't added for some reason. Am I missing something?

thebradbain: Ah! I ran into this issue too -- 

you need to add your file to the `depends_on:` line within `ApplicationController`

stephaniegiang: Oh really? I knew I was missing something 😂

stephaniegiang:
@d12 this PR is ready!

PR #2041, opened 2019-07-18
Add type column to LMS Configuration

stephaniegiang:
## What
With the UI changes, we want to add a type to `LTIConfiguration` so we can make the experience more customized for the user.


thebradbain:
Instead of a `text` type, what if we use Postgres's enum type instead? We can easily extend the enum later, but I'm thinking

`:blackboard`, `:canvas`, `:moodle`, `:brightspace`, and `:other`, with the default being `:other`. Any other big ones we're missing?

stephaniegiang:
Ohhh I like that idea

d12:
Don't forget what we want to use this column for! I think there were a couple of things, correct me if I'm wrong:

- Populating views like `<h3>Import your students from <%= lms_config.type %></h3>`.
- Determining which features to offer users (in the future)

We can't do # 1 with a enum type like `other`, I'd much prefer to see "Import your students from D2L" than "import your students from LMS tool", even if we don't have images in these "other" cases. Are we able to fetch the LMS name when going through the authorization stuff and use that instead?

stephaniegiang:
@d12 I think we could get the name of the LMS through the authentication. Correct me if I'm wrong, @thebradbain. 

I like that idea so we get a better idea of the next most popular LMS that teachers use outside of the 4 main ones.

db/schema.rb
```diff
@@ -169,10 +169,11 @@
     t.text "consumer_key", null: false
     t.text "shared_secret", null: false
     t.text "lms_link"
+    t.bigint "organization_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.bigint "organization_id"
```
BenEmdon: Lol why did this column want to move so badly 😂

app/models/lti_configuration.rb
```diff
@@ -3,6 +3,14 @@
 class LtiConfiguration < ApplicationRecord
   belongs_to :organization
 
+  enum lms_type: [
+    :other,
```
BenEmdon: Nit: Is this the order which we would display them in? Usually we would put an `unknown` or `other` option at the end of an enum

db/schema.rb
stephaniegiang: I don't know 😭 This also happened with @thebradbain's PR too.

app/models/lti_configuration.rb
stephaniegiang: Yea however I also display a default. 

https://github.com/education/classroom/blob/b63bb2d26aafe1d372cf8c04ffcac5c103b23188/db/schema.rb#L176

To make it easier to know where the default is, I just made it 0. 

I can still fix this to add it to the end, but because this is going to be changing (adding more LMSs) I think it sits better at the top.

thebradbain:
@d12 @stephaniegiang we _do_ get the platform name on launch, but there's no guarantees around what it is (and it may just be an unfriendly GUID): https://www.imsglobal.org/spec/lti/v1p3/#platform-instance-claim

We also _may_ get the `tool_consumer_instance_name`, but it appears that it's not required to be sent, though most do https://www.imsglobal.org/toolcompare.html

app/models/lti_configuration.rb
thebradbain: We should probably declare the enum as a hash with explicit orderings rather than array, as that way we're not dependent on implicit ordering: 

https://naturaily.com/blog/ruby-on-rails-enum (see section 1)

stephaniegiang:
@thebradbain Sounds good! That is something that can added in the next PR.

stephaniegiang:
@d12 @BenEmdon ready for another round of review!

app/models/lti_configuration.rb
```diff
@@ -3,6 +3,14 @@
...
   belongs_to :organization
 
+  enum lms_type: {
+    canvas: 1,
+    blackboard: 2,
+    brightspace: 3,
+    moodle: 4,
+    other: 5
+  }
+
```
thebradbain: What if we added a method here that would handle print human readable strings? Something like

```
def lms_type_string
case lms_type
when :canvas 
   "Canvas"
when :blackboard
   "Blackboard"
when :brightspace
   "D2L Brightspace" 
[..etc..]
default: 
   "Learning Management System"
end
```

app/models/lti_configuration.rb
```diff
@@ -3,6 +3,14 @@
 class LtiConfiguration < ApplicationRecord
   belongs_to :organization
 
+  enum lms_type: {
+    canvas: 1,
+    blackboard: 2,
+    brightspace: 3,
+    moodle: 4,
+    other: 5
```
thebradbain: I know @BenEmdon mentioned this, but what if we put `:other` as 0 (since enums can start at 0) but keep it at the end of the hash? That way we can add a 5th lms and it can be number 5, etc. I also kind of like explicitly keeping `other: 0` because 0 seems more of a "special" case than the other natural numbers.

app/models/lti_configuration.rb
stephaniegiang: Actually found a different way to do that. Added the changes!

app/models/lti_configuration.rb
thebradbain: omg love the change to strings didn't know that was even possible ✨

app/models/lti_configuration.rb
thebradbain: I _love_ the change to strings instead of numbers for the enum indexes. I guess my big question now is: does that actually accomplish the same thing as that method?

According to [https://stackoverflow.com/questions/32938729/how-to-store-enum-as-string-to-database-in-rails/32939934`](this Stack Overflow post), it seems to actually access the corresponding string we would have to do `lti_configuration[:lms_type]` instead of `lti_configuration.lms_type` since that would still give us a string representation of the symbol, which I don't think would work because we do care about capitalization here _and_ would like the default `other` type to correspond to "Learning Management System"

stephaniegiang: Ahhh ok. Maybe we can do a method called `lms_type` not sure if that would cause Rails to not know what is happening. But I can look into it. 

I don't think I like using `Learning Management System` as other. Because looking at a dropdown or whatever we choose will be confusing to the user. You would see all the names of the LMS and then on the bottom you will just see `Learning Management System`. To me that doesn't connect with `other`.

thebradbain: creating `lms_type` on the model should be totally fine, it'll just be treated as a regular method (a "computed attribute" is I think what it's called in Rails terminology) -- look at `LtiConfiguration#context_membership_url` in #2033 for an example.


As far as naming goes, based on [@d12's reply](https://github.com/education/classroom/pull/2041#issuecomment-513240019) 

>>>
> - Populating views like `<h3>Import your students from <%= lms_config.type %></h3>`
>
> We can't do # 1 with a enum type like other, I'd much prefer to see "Import your students from D2L" than "import your students from LMS tool"

Maybe "Other Learning Management System" or "Other LMS" works in context?

app/models/lti_configuration.rb
```diff
@@ -3,6 +3,19 @@
...
 
+  enum lms_type: {
+    canvas: "Canvas",
+    blackboard: "Blackboard",
+    brightspace: "Brightspace",
+    moodle: "Moodle",
+    other: "other"
+  }, _prefix: true
+
+  def lms_name(default_name: "Other Learning Management System")
```
thebradbain: I've gone ahead and added a very simple `lms_name` helper method. Essentially it will return the name corresponding the LMS platform type _unless_ that type is `:other`. In that case, it will either return a default or let us provide it depending on the view context at the time. Also added some simple tests for it.

stephaniegiang: Love it!

thebradbain:
@d12 as soon as this is approved and merge, we're ready to review #2056, which changes the configuration setup language to be more friendly / LMS-specific

app/models/lti_configuration.rb
d12: If the value of `:other` in our enum was "Other Learning Management System", wouldn't that kill the need for this method?

thebradbain: @d12 Sometimes (see PR #2056) we want a different text representation for `:other` than “Other Learning Management System” on the UI — like “your Learning Management System” or “a learning management system”, etc.

d12:
```
D, [2019-07-22T17:43:08.590427 #4] DEBUG -- :    (1.6ms)  SELECT pg_try_advisory_lock(2076930691114369125)
D, [2019-07-22T17:43:08.625722 #4] DEBUG -- :    (2.3ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
I, [2019-07-22T17:43:08.628765 #4]  INFO -- : Migrating to LmsType (20190718211229)
D, [2019-07-22T17:43:08.633797 #4] DEBUG -- :    (1.9ms)  BEGIN
== 20190718211229 LmsType: migrating ==========================================
-- add_column(:lti_configurations, :lms_type, :text, {:null=>false, :default=>"other"})
D, [2019-07-22T17:43:08.657994 #4] DEBUG -- :    (19.5ms)  ALTER TABLE "lti_configurations" ADD "lms_type" text DEFAULT 'other' NOT NULL
   -> 0.0230s
== 20190718211229 LmsType: migrated (0.0240s) =================================

D, [2019-07-22T17:43:08.668004 #4] DEBUG -- :   SQL (1.9ms)  INSERT INTO "schema_migrations" ("version") VALUES ($1) RETURNING "version"  [["version", "20190718211229"]]
D, [2019-07-22T17:43:08.674315 #4] DEBUG -- :    (5.6ms)  COMMIT
D, [2019-07-22T17:43:08.686249 #4] DEBUG -- :   ActiveRecord::InternalMetadata Load (2.7ms)  SELECT  "ar_internal_metadata".* FROM "ar_internal_metadata" WHERE "ar_internal_metadata"."key" = $1 LIMIT $2  [["key", "environment"], ["LIMIT", 1]]
D, [2019-07-22T17:43:08.701509 #4] DEBUG -- :    (2.9ms)  BEGIN
D, [2019-07-22T17:43:08.704629 #4] DEBUG -- :    (1.6ms)  COMMIT
D, [2019-07-22T17:43:08.706777 #4] DEBUG -- :    (1.8ms)  SELECT pg_advisory_unlock(2076930691114369125)
```

PR #2051, opened 2019-07-19
Add template? method to GitHubRepository

EricPickup:
## What
This PR adds a `template?` method to `GitHubRepository` that gives us a clean/easy way to check if a repository is a template.


app/models/github_repository.rb
```diff
@@ -228,6 +230,12 @@ def temp_clone_url
     end
   end
 
+  def template?
+    options = { accept: TEMPLATE_REPOS_API_PREVIEW }
+    endpoint_url = "#{GITHUB_API_HOST}/repositories/#{@id}"
```
BenEmdon: Why not rather use:
```suggestion
    endpoint_url = "#{StarterCodeImportable::GITHUB_API_HOST}/repositories/#{@id}"
```

app/models/github_repository.rb
```diff
@@ -228,6 +230,12 @@ def temp_clone_url
     end
   end
 
+  def template?
+    options = { accept: TEMPLATE_REPOS_API_PREVIEW }
```
BenEmdon: Why not rather use:
```suggestion
    options = { accept: StarterCodeImportable::TEMPLATE_REPOS_API_PREVIEW }
```

app/models/github_repository.rb
```diff
@@ -5,6 +5,8 @@ class GitHubRepository < GitHubResource
   depends_on :import
 
   DEFAULT_LABEL_COLOR = "ffffff"
+  GITHUB_API_HOST = "https://api.github.com"
+  TEMPLATE_REPOS_API_PREVIEW = "application/vnd.github.baptiste-preview"
```
BenEmdon: If you accept my previous suggestions you could also do this:
```suggestion
```

app/models/github_repository.rb
```diff
@@ -5,6 +5,8 @@ class GitHubRepository < GitHubResource
   depends_on :import
 
   DEFAULT_LABEL_COLOR = "ffffff"
+  GITHUB_API_HOST = "https://api.github.com"
```
BenEmdon: If you accept my previous suggestions you could also do this:
```suggestion
```

app/models/github_repository.rb
EricPickup: Good idea. I'll actually do the reverse and keep these constants in `github_repository.rb` since `StarterCodeImportable` doesn't use those constants.

PR #2056, opened 2019-07-22
Specify LMS Platform on Creation

### LINKED ISSUE #2057: Refer to an LMS by its Platform name
Per #2029, we should refer to a Learning Management System by its official platform name on the front-end (e.g. Canvas, Blackboard, Moodle) rather than generic"/Learning Management System" or "LMS".

thebradbain:
## What
Per #2029, we would like to refer to a Learning Management System by its official platform name on the front-end (e.g. Canvas, Blackboard, Moodle) rather than generic"/Learning Management System" or "LMS". Also closes #2057.

This PR creates the logic around specifying which platform the LMS is within the configuration creation flow if it's not already specified. It also changes the language around configuring an LMS in classroom to be friendlier and based upon the platform name.

Also compatible with what @stephaniegiang is currently working on regarding platform-specific buttons/options (e.g. "Connect to Canvas", "Moodle", etc) that a user would click on during initial classroom setup if they're interested in setting up a specific LMS, and would allow skipping over the "Choose LMS Provider" step in that case.

**AFTER**
_Note the new LMS Platform Selection screen._
We still have to change the initial `info` view, too, but we should decide exactly what we want to say there separate from this PR.
oQ8THFcBOf

**BEFORE**
_Note the old language versus the new above_
Screen Shot 2019-07-21 at 11 19 52 PM

_To be merged after #2041_


app/helpers/lti_configuration_helper.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module LtiConfigurationHelper
+  def lms_type_select_options(default_name: "Other Learning Management System")
+    LtiConfiguration.lms_types.collect do |k, v|
```
d12: `map` and `collect` are aliases, but it's more standard to use `map` in Ruby.

app/helpers/lti_configuration_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+      if k == "other"
```
d12: I mentioned this on my other PR, but if we change the value of `other` in the enum to the `default_name` here, we get to remove a lot of special case logic like this.

app/views/orgs/lti_configurations/_form.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<% allow_cancel = local_assigns.fetch(:allow_cancel, false) %>
+
+<%= form_with model: model, url: lti_configuration_path, method: method do |f| %>
+  <div class="form-group">
+    <dt><label>Learning Management System Provider</label></dt>
+    <dt>What platform does your institution use for its Learning Management System?</dt>
+    <dd><%= f.select :lms_type, lms_type_select_options, { prompt: "Select from the options listed" }, { class: "form-select" } %></dd>
+  </div>
+
+  <!--<div class="form-group">
```
d12: Is this commented out stuff supposed to be here?

app/views/orgs/lti_configurations/show.html.erb
```diff
@@ -1,25 +1,27 @@
...
+
   <div class="columns">
     <div class="one-fourth column">
       <%= render partial: 'organizations/side_menu' %>
     </div>
     <div class="three-fourths column">
       <div class="site-content">
         <div class="site-content-cap">
-          <h2 class="site-content-heading">Learning Management System Configuration</h2>
+          <h2 class="site-content-heading">Connect to <%= lms_name %></h2>
```
d12: This will say `Connect to Other Learning Management System` if they're using a non top-5 LMS. This reads strangely, shouldn't it be something more like `Connect to Learning Management System`? If a user was using one of these other systems, it'll be referred to as a `Other` system throughout the app, forever.

thebradbain: Yeah, so here we’re actually using the feature for `lms_name` that allows us to specify a different default depending on context. Here specifically it would say, “Connect to your Learning Management System”

app/helpers/lti_configuration_helper.rb
thebradbain: I also left another reply there, but in the #show view specifically we use the functionality of changing the default name from “Other Learning Management System” to “your Learning Management System”.

This view helper is primarily for the drop down in selecting which type of LMS they use, and so I think “Other Learning Management System” makes more sense here, and you’re right, in this case we don’t need to change the default.

stephaniegiang:
Love this PR! In my PR #2063, could the buttons also link back to these individual LMS? Just curious on how this will work when you are creating a classroom.

app/views/orgs/lti_configurations/show.html.erb
```diff
@@ -1,25 +1,27 @@
 <div class="remodal-bg">
   <%= render 'organizations/organization_banner' %>
 
+  <% lms_name = current_lti_configuration.lms_name(default_name: "your Learning Management System") %>
```
thebradbain: @d12 here specifically is where we set the default name in the `#show` view.

app/views/orgs/lti_configurations/_form.html.erb
thebradbain: We actually had this field on the model as a rough idea for functionality it we currently don't need / may not ever need, and I've temporarily commented it out because it was just cluttering up the UI. Still deciding if we should leave it or delete it!

thebradbain:
@d12 one more screenshot, I promise 😅 

Here's the generic/unlisted LMS platform case: 

Screen Shot 2019-07-22 at 10 00 56 AM

app/views/orgs/lti_configurations/_form.html.erb
thebradbain: I've decided to remove it from the UI for now, as we don't have any functionality which uses the attribute it sets.

thebradbain:
@d12 ready for another round of review!

app/models/lti_configuration.rb
```diff
@@ -3,6 +3,8 @@
 class LtiConfiguration < ApplicationRecord
   belongs_to :organization
 
+  validates :lms_type, presence: true
```
d12: Won't this throw exceptions for all LMS configurations that don't currently have a lms_type? ActiveRecord will run validations any time a record is saved. We may need a backfill here first.

thebradbain: Within the migration we specified a default LMS type (`:other`) that automatically backfilled any that was previously nil, as it's a required column on postgres now.

d12: Oh awesome, I didn't know we had a default value.

PR #2063, opened 2019-07-22
Integration flow for LMS and Google Classroom

stephaniegiang:
## What
This PR has the new UI changes for LMS and Google Classroom.

This has both back-end and front-end. So this PR will probably be split. But for now this is to keep track of changes and for discussion!


stephaniegiang:
Closing this PR since we are going with a different UI for LMS and Google Classroom

PR #2069, opened 2019-07-22
Add the ability to create assignments via template repos to the new unified creator service

shaunakpp:
## What
Repo templates was recently added to the creators in #1985, with this PR we get the same functionality in the new unified service for creating GitHub repositories. This will enable us to roll it out under the feature_flag `unified_repo_creators` without losing out on this amazing new feature!


app/services/create_github_repo_service.rb
```diff
@@ -59,6 +68,35 @@ def create_github_repository!
...
+    options = {
+      name: exercise.repo_name,
+      owner: exercise.organization_login,
+      private: assignment.private?,
+      description: "#{exercise.repo_name} created by GitHub Classroom",
+      accept: TEMPLATE_REPOS_API_PREVIEW,
+      include_all_branches: true
+    }
+
+    github_repository = client.post(
```
d12: Are we able to move this GitHub logic to `github_repository`?

shaunakpp: @EricPickup and I had a chat on this, we are waiting for the Octokit upgrade and we are then going to update this in a separate PR, is that okay?

d12: Yup, I saw that PR after I left this comment. No problem :)

shaunakpp: Also created an issue for this, tagging here for reference: #2075

PR #2071, opened 2019-07-22
LMS platform specific language

thebradbain:
## What
Part of issue #2029. Similar to #2057, but more related to changing language rather than functionality.

Changes language around the LMS information/setup pages based on specified `lms_type`

**Currently (Pre-change):** 
_Note: We are also moving to get rid of the sidebar within these views, as specified in #2029_
Screen Shot 2019-07-22 at 4 17 36 PM

**After:**
_Note: First two are examples when lms_type is specified. The last one is the general case._
Screen Shot 2019-07-22 at 3 52 55 PM

Screen Shot 2019-07-22 at 3 53 09 PM

Screen Shot 2019-07-22 at 3 54 08 PM


app/controllers/orgs/lti_configurations_controller.rb
```diff
@@ -35,6 +35,11 @@ def new
       @lti_configuration = LtiConfiguration.new(lms_type: nil)
     end
 
+    def info
+      lms_type = params[:lms_type] || nil
```
d12: What is the value of `params[:lms_type]` here in the nil case? We only need `|| nil` if it can be `false`.

app/views/orgs/lti_configurations/info.html.erb
```diff
@@ -1,35 +1,36 @@
...
-          GitHub Classroom.</p>
-          <ul>
-            <li>Consumer key</li>
-            <li>Shared secret</li>
-          </ul>
+    <div class="site-content-body markdown-body">
+      <p>Connecting GitHub Classroom to <%= lms_name %> will allow you to manage your students and submissions easier!</p>
+      <h3>What integrating Classroom with <%= lms_name %> can do for you</h3>
+      <ul>
+        <li>Sync rosters from <%= lms_name %> with Classroom</li>
```
d12: Small copy suggestion:

```suggestion
        <li>Sync rosters from <%= lms_name %> to Classroom</li>
```

Issue #2023, opened 2019-07-17
Untangle Google Classroom functionality from RostersController


thebradbain:
Currently, Google Classroom functionality is extremely coupled with `Orgs::RostersController` and its views, so much so that almost every action within `RostersController` interacts with Google Classroom in some way.

<img width="749" alt="Screen Shot 2019-07-16 at 6 22 41 PM" src="https://user-images.githubusercontent.com/2936741/61340375-e5296700-a7f6-11e9-9769-5e294d823d2c.png">

This doesn’t seem deliberate (but I may be wrong! I haven’t interacted too much with this code), as right now it doesn't look possible to interact with Google Classroom outside of `RostersController`, including authentication.

Once LMS / Google Classroom integration is in a stable state, we should probably consider refactoring -- I'd love to help out! 

Also tagging any other interns in case anyone is looking for another project 👀

d12:
Currently this is all in the rosters controller because all of the UI interactions with google classroom is on roster/show! We have a meeting today with @brandonrosage to discuss how to better design this feature which will help in this code untangling.

stephaniegiang:
I'm going to take this over. Since moving the Google Classroom to a new UI, I'm already doing a lot of this. So by the end of it, it should be entirely untangled!

stephaniegiang:
Now completely removed from Rosters Controller 🎉 
PR #2074

PR #2074, opened 2019-07-23
Move Google Classroom authentication and routing to separate controller and models

srinjoym:
Addresses #2023 as we move towards using integrations for more than just roster imports.


d12:
Ping me when this is good for review!

stephaniegiang:
Ready for review @BenEmdon!

Also if you are available @d12

app/models/google_classroom_course.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class GoogleClassroomCourse
+  attr_accessor :name, :students
```
shaunakpp: can we change this to `attr_reader` as we are not mutating their values on this object anywhere else

app/controllers/orgs/rosters_controller/google_classroom_dependency.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    def sync_google_classroom
+      unless current_organization.google_course_id
+        flash[:error] = "No Google Classroom has been linked. Please link Google Classroom."
+        return
+      end
+
+      latest_students = list_google_classroom_students
+      latest_student_ids = latest_students.collect(&:user_id)
+      current_student_ids = current_roster.roster_entries.collect(&:google_user_id)
```
shaunakpp: can we replace this by something like this:
`current_student_ids = current_roster.roster_entries.pluck(&:google_user_id)`
This will result in a `SELECT roster_entries.google_user_id` instead of `SELECT roster_entries.*`.
This can be particularly be helpful if the number of roster entries is large.

stephaniegiang: Oh I like that idea a lot. I always forget that is something you can do in rails.

app/controllers/orgs/rosters_controller/google_classroom_dependency.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    def sync_google_classroom
+      unless current_organization.google_course_id
+        flash[:error] = "No Google Classroom has been linked. Please link Google Classroom."
+        return
+      end
+
+      latest_students = list_google_classroom_students
+      latest_student_ids = latest_students.collect(&:user_id)
+      current_student_ids = current_student_ids = current_roster.roster_entries.pluck(&:google_user_id)
```
shaunakpp:
```suggestion
      current_student_ids  = current_roster.roster_entries.pluck(&:google_user_id)
```

stephaniegiang: Oh looool my bad

Issue #2032, opened 2019-07-18
Add ability to change repo into a template repo through classroom


EricPickup:
With template repos shipped out, I can't imagine how many people are going to see this issue when creating an assignment and trying to use template repos (default option):
![image](https://user-images.githubusercontent.com/1490434/61420367-e8842780-a8b6-11e9-9499-eee7dc4f2a85.png)

Thanks to @cheshire137 we will soon be able to copy branches with template repos which will allow us to remove the source importer completely which would make template repos the _only_ option.

We absolutely should add an option to make a repository a template through Classroom (already possible through the API).

We could add a simple dialogue box that would appear before this error during validation asking teachers if they'd like to make it a template repo.

We could add a custom error box for this case that would include a button `Make my repository a template` (or something like that) that would send the API call.

EricPickup:
@femmebot What do you think about how we should approach this from a UX perspective?

To provide some context: We've adding the ability to import starter code using template repositories. The teacher enters the repository with the starter code which **must** template repository.

femmebot:
@EricPickup Is this actually a user error or did we change something that requires existing users to change old behaviors?

EricPickup:
@femmebot this would be a user error due to a change we made which requires users to change old behaviors (if they use the template option).

We already decided how we're going to handle it though, it's in the description under the strikethrough text.

femmebot:
Ah, awesome! That's what I was going to suggest. I.e., either hide complexity from the user or provide a way to help change their behavior without blaming them for the functionality change.

PR #2080, opened 2019-07-23
Automatically make starter repo a template

EricPickup:
## What
Step 1/2 for #2032.

If the teacher chooses to use template repos to import starter code and enters a non-template repository, we will try to make it a template repo on their behalf.

FavK6Okbbx

If the user does not have permission to change the setting:
image

I also added a message on the form informing them that we're doing this:
image


EricPickup:
@femmebot what do you think about the error message here? This is related to #2032

EricPickup:
@femmebot 
> that's good though can we suggest alternative options to try (unless source importer is their only option) for them so they're not completely stuck from completing the task?

image

There are two options to import code: templates (new) and source importer. The source importer description states that they don't need templates to use it (making it an alternative option). Let me know if that's good enough!

> Is it possible to perform a check first to see if the repo they entered is a template repo?

We're actually already doing this! In production if they enter a non-template repo, the validation will catch that when the user tries to save the assignment: 

image

> Is the only reason why we recommend using a template repository because it's faster? 

Templates is faster and much more reliable (source import success rate is 93% vs templates at 100% for the past 24 hours with a sample of ~1000 repos created).

> Since very few people will have template repos (since it's a new feature), I'm concerned that we're recommending a path where there's a higher likelihood of failure. Might be better to indicate that it's faster, rather than recommended (at least, for now, until we get more traction on template repos).

That's a good point that we caught onto earlier this week. If you look at my first screenshot in this comment you can see we removed the **Recommended** and added a **Beta** badge to it.

> Also, Import starter code using source importer shouldn't be in red since it wasn't the selected option.

Great point. I'll get that fixed!

EricPickup:
Closing this since https://github.com/education/classroom/issues/2241 will do this for us

PR #2089, opened 2019-07-25
Unify LMS/Google Classroom Creation Flow

thebradbain:
## What
Unifies the LMS/Google Classroom Creation Flow. Just an initial piece of #2082, but an important one nonetheless to be ready for user testing. Also makes some smaller UI adjustments around LMS UX (such as friendlier error messages).

Corresponds to Creation Flow Candidate no. 2 within [this UI design Google Doc](https://docs.google.com/document/d/1YHOpckAkSPoLqWwN7HJsNIon-jYFDZ7QDBQ2c2yFhCM/edit?usp=sharing).

_Adds a single setup step to the creation flow, which a user may choose to skip_
J3UiyYY1oR

_Handholds through LMS Configuration Walkthrough if they do choose to setup LMS_
Vic79Ch5Bq

Issue #2082, opened 2019-07-23
LMS UX Flows


thebradbain:
Provide UX stories for tasks we anticipate our users to do to better gather feedback (from both our designers and our testers). 

Currently, these cover the most commonly anticipated use-cases of configuring an LMS and also importing a roster from the LMS. You'll notice the creation flow differs whether its Google Classroom or a standard LMS we're integrating, and would love your thoughts on which flow makes the most sense to focus on moving forward.

[Google Doc](https://docs.google.com/document/d/1YHOpckAkSPoLqWwN7HJsNIon-jYFDZ7QDBQ2c2yFhCM/edit?usp=sharing)

sophshep:
Thanks for sharing these yesterday on a call, and here in the doc! I left some specific feedback as comments in the doc. 

Overall, I don't feel that one makes more sense over the other since they are somewhat similar. Instead, my feedback for both is to decide whether customization should happen during the creation flow or after, and to separate creation/onboarding from the classroom itself. Right now, both flows are a hybrid, which might be confusing to users.

As mentioned in the doc, this could be done in two different ways:

### 1. Set-up during onboarding

A user is guided through every step of setting up their new classroom.

![IMG_6240](https://user-images.githubusercontent.com/527589/61811800-e0bd0980-ae07-11e9-9637-d95d4fe6da02.jpg)

Pros:
* More guided, users may feel less lost
* Important settings are not hidden, so a user is more likely to set it up fully up front
* This flow not only helps a user set up, but also guides them through some features of classroom

Cons:
* If they get stuck on an area, will they know how to come back to it?
* Does it introduce too much friction? Will doing the set-up work up front feel like a burden?

### 2. Blank slate

A user is able to quickly create a classroom, and then presented with a list of different things they can do next.

![IMG_6243](https://user-images.githubusercontent.com/527589/61812000-45786400-ae08-11e9-9b25-73e83baca75f.jpg)

Pros:
* Allows a user to explore the classroom UX themselves, which may make them get more familiar with it sooner
* Reduce time and effort between clicking "new classroom" and having a classroom

Cons:
* Users might miss important set-up steps
* This doesn't work super well with the layout/navigation as it is now, since some things are settings and some are assignments

In an ideal world, we would test both of the above approaches to see which results in the most successful classrooms and happy users!

d12:
I'm a fan of # 1 personally. I'd like to do deeper investigations into both options in the future though :)

PR #2097, opened 2019-07-25
Roster import UI tweaks

thebradbain:
## What
Google Classroom and LMS integrations already have the exact same roster import flow, but we had redundant/duplicate buttons scattered around and brittle if/else view logic.  This does **not** change any of our assumptions / rules around when a roster can be imported or not, etc. It is purely view tweaks.

This PR refactors those components so the experience and UI is consistent such that it won't matter from a view standpoint whether the instructor is linked to Google Classroom or a generic LMS. 

Before: 
_Two buttons that do the same thing_
Screen Shot 2019-07-25 at 1 34 55 PM
Screen Shot 2019-07-25 at 1 36 04 PM

After:
_One button which redirects to the standard "Choose an LMS" screen from #2089, which includes Google Classroom_
Screen Shot 2019-07-25 at 1 35 15 PM
Screen Shot 2019-07-25 at 1 35 45 PM


thebradbain:
Yeah definitely! That was something that @sophshep made a note of in #2082 on the Google Doc, and the next change I'd like to make after this is merged.

Issue #2095, opened 2019-07-25
Upgrade to Rails 5.2.x


iancanderson:
Rails 6 is almost stable, so let's get on 5.2.x now!

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2

PR #2102, opened 2019-07-26
Update rails to 5.2.x

iancanderson:
Closes https://github.com/education/classroom/issues/2095

## What
Update rails to the latest stable version so we can be ready to go to 6.0 when it's stable!

## TODO
- [x] Fix test failures
- [x] Fix deprecation warnings in test suite
- [x] Review changelogs for updated gems to make sure there are no surprising new behaviors/breaking changes for the app
  - [x] rails
  - [x] bullet
  - [x] dotenv-rails
  - [x] lograge
- [X] Add bootsnap (NOTE: Bootsnap was manually added 2 years ago, I added a step in our bootstrap script to clear it's cache as that can become massive over time)
- [x] Locally test - Make sure existing sessions continue to be valid in Rail 5.2.x
- [ ] Manually test branch on production


Gemfile
```diff
@@ -8,7 +8,7 @@ git_source(:github) do |repo_name|
 end
 
 ruby File.read(File.expand_path("../.ruby-version", __FILE__)).chomp
-gem "rails", "~> 5.1", ">= 5.1.7"
+gem "rails", "~> 5.2.3"
```
d12: 🚀

app/models/roster_entry.rb
```diff
@@ -13,16 +13,19 @@ class IdentifierCreationError < StandardError; end
...
+      ON roster_entries.user_id = assignment_repos.user_id
+      AND assignment_repos.assignment_id='#{sql_formatted_assignment_id}'
+     SQL
     order("assignment_repos.created_at")
-      .joins <<~SQL
-        LEFT OUTER JOIN assignment_repos
-        ON roster_entries.user_id = assignment_repos.user_id
-        AND assignment_repos.assignment_id='#{sql_formatted_assignment_id}'
-      SQL
+      .joins(join_query)
```
d12: Did you mean to use `Arel.sql` here?

config/initializers/new_framework_defaults.rb
```diff
@@ -20,6 +20,3 @@
 
 # Require `belongs_to` associations by default. Previous versions had false.
 Rails.application.config.active_record.belongs_to_required_by_default = false
-
-# Do not halt callback chains when a callback returns false. Previous versions had true.
-ActiveSupport.halt_callback_chains_on_return_false = true
```
d12: Will this break any expectations within the app that returning false within a callback halts the chain?

config/environments/production.rb
```diff
@@ -41,6 +40,9 @@
   # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
+  # Store uploaded files on the local file system (see config/storage.yml for options)
+  config.active_storage.service = :local
```
jeffrafter: This feels odd for production? Is this meant to be only for development?

JessRudder: @shaunakpp, I believe you were working with Ian on the upgrade. Do you have insight into why this was added? And if it should be for production? If not, I can dig deeper. Just didn't want to chase answers down if you already had them. 😄

shaunakpp: It came up with the `app:update` task afaik, but we don't really use  active storage in classroom, yet, so we can comment it out I think.

config/initializers/new_framework_defaults.rb
JessRudder: Thanks for bringing this up. I dug into all of our existing callbacks and they should not be affected by this change.

For current before callbacks, the methods either end with a statement that will always be true (a var assignment of a truthy value, a string, etc), they explicitly have `true` as the final return value to ensure the callback won't be halted or they raise an exception which intentionally halts the callbacks.

PR #2106, opened 2019-07-26
Change layout of LMS links

stephaniegiang:
## What
Currently, Google Classroom is displayed as "Google Class" which we should not do. So this PR makes it so that we can display the full name. To do that I'm also changing how many LMSs are displayed in one row.

Also updated the text copy!

### Before
image

### After
image


thebradbain:
Love it ! ❤️

PR #2107, opened 2019-07-26
This pull request removes ElasticSearch from app as we've moved to PostgreSQL full text search

shaunakpp:
## What
🔴 This PR removes elastic search from the application. We should not merge this until we have completed the following steps:
* Safely remove all the jobs from `Sidekiq`'s `chewy` queues
* Make sure the new `Postgres` search gives us the same exact features as old `ES-search`
* Monitor the performance of `postgres` full text search
* Evaluate if ES going to be useful in future

I'm keeping this PR in draft till we evaluate whether or not to do this.

CC: @andrewbredow @d12


d12:
My only input here is that the ES cloud provider we're forced to use on Heroku has caused us multiple availability incidents in the past. My suggestion is to get rid of it if we don't need it, or ensure we can continue serving users when ES is down if we do need it.

d12:
@shaunakpp if you have bandwidth, Let's move forward with this :+1: Please get review from @education/ce-classroom here.

PR #2113, opened 2019-07-27
Report swallowed errors with templates to Haystack

EricPickup:
## What
We're currently swallowing GitHub errors (only reporting to Datadog which doesn't give us any of the error details).

With the recent release of template repos and a rise in errors with templates for the past couple of days, we should report these errors to Haystack so we can determine the source of the issue(s).

cc: @shaunakpp


EricPickup:
Tests will be fixed once https://github.com/education/classroom/pull/2112 goes in

shaunakpp:
Since we call the creators in a background job, if there's an error its going to be reported multiple times because of the retry mechanism. Should we do something to handle this? 
Or we could keep it as it is if it is a temporary change 😄

EricPickup:
@shaunakpp didn't see this till I started the deployment but that's a good point! I think that since it's a temporary change (and already deploying) it's no big deal to have the repeats since they'll collapse under the same error in Haystack anyway.

PR #2114, opened 2019-07-29
Refactor LtiConfiguration to use LMS Specific settings

thebradbain:
## What
In anticipation of avoiding a giant "everything and the kitchen sink" `LtiConfiguration` model, refactors out LMS-specific settings (e.g. icons, platform name, whether an LMS supports autoconfiguration or not, etc) based on the already existing `lms_type`. 

Introduces no new functionality, but will hopefully make adding new functionality more straightforward in the future. Also relevant to handling edge-cases where an LMS may behave in a nonstandard manner for certain LTI capabilities (looking at you, Blackboard  👀)


app/models/lti_configuration.rb
```diff
@@ -32,12 +34,38 @@ def context_membership_url(use_cache: true, nonce: nil)
...
+      vendor_url: "https://classroom.github.com/"
+    )
+
+    builder.add_vendor_attributes(lms_settings.vendor_domain, lms_settings.vendor_attributes)
+    builder.to_xml
+  end
+
+  private
+
+  def lms_settings
```
thebradbain: This is where the only interesting thing in this PR happens, which is dynamically determining what kind of `LtiConfiguration::Settings` to use based on the already existing `lms_type`.

app/models/lti_configuration.rb
```diff
@@ -5,6 +5,10 @@ class LtiConfiguration < ApplicationRecord
 
   validates :lms_type, presence: true
 
+  delegate :icon, to: :lms_settings, prefix: :lms
```
d12: TIL `prefix:` , thats awesome :sparkles:

app/models/lti_configuration.rb
```diff
@@ -32,12 +34,38 @@ def context_membership_url(use_cache: true, nonce: nil)
...
     return nil unless membership_url
 
     self[:context_membership_url] = membership_url
     save!
 
     membership_url
   end
+
+  def xml_configuration(launch_url)
+    return nil unless supports_autoconfiguration
```
d12: `return unless` here instead of `return nil unless`

app/views/orgs/lti_configurations/show.html.erb
```diff
@@ -33,26 +33,37 @@
...
             <%= octicon 'clippy' %>
             Copy
           </button>
         </span>
       </div>
 
       <br>
-
+      <% if current_lti_configuration.supports_autoconfiguration %>
       <dt><label>Application Auto-Configuration URL</label></dt>
```
d12: Indentation is off here because of the `if` introduced above.

app/models/lti_configuration/blackboard_settings.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class LtiConfiguration
+  class BlackboardSettings < GenericSettings
+    def initialize
```
d12: There's another way of designing this that _I_ believe is a bit cleaner,

If you define each of `platform_name`, `icon`, etc as methods instead of `attr_readers` in `GenericSettings`, we can override those methods in these subclasses with our LMS-specific values.

It'd look something like this

```
class GenericSettings
  def icon
    # What is the default icon? Or leave empty if nil
  end
end

class BlackboardSettings < GenericSettings
  def icon
    "blackboard-logo.png"
  end
end
```

Each settings class overrides the specific methods it needs, and anything  not overridden falls back on the generic class. Defaults are defined in methods instead of a large hash or part of opts[:this] || "default" clauses. 

Feel free to ignore, I'm not sure what the preferred way of doing this is :)

spec/controllers/orgs/lti_configurations_controller_spec.rb
```diff
@@ -254,10 +254,23 @@
...
+        end
+      end
+
+      context "with autoconfiguration disabled" do
+        before(:each) { LtiConfiguration.any_instance.stub(:supports_autoconfiguration).and_return(false) }
+
+        it "does not return an xml configuration" do
+          get :autoconfigure, params: { id: organization.slug }
+          expect(response).to have_http_status(:not_found)
+        end
```
d12: The settings classes are very simple, but it'd be worth writing simple assertions making sure the super / method override stuff is doing what you think it's doing.

app/models/lti_configuration/blackboard_settings.rb
thebradbain: Good idea, I like that!

thebradbain:
@d12 ready for a second round of review! I changed the classes to use methods rather than `:attr_reader`, but am wondering how I should test that: just write unit tests testing that the inheritance chain works by asserting the properties are what we expect?

app/models/lti_configuration/moodle_settings.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class LtiConfiguration
+  class MoodleSettings < GenericSettings
+    def initialize
```
d12: We shouldn't need this initializer anymore!

thebradbain: oop! good catch

thebradbain:
@d12 tests are in 💯

PR #2120, opened 2019-07-30
Fix Google Authentication

### LINKED ISSUE #2116: Need to re-check user authenication
This issue happens when a user goes to Google and removes GitHub Classroom from their authorized applications. 

To fix this, we need to ask the user to re-authenticate. 

Google::Apis::AuthorizationError

**[Google::Apis::AuthorizationError](https://haystack.githubapp.com/classroom/rollups/410639761472c358cf1b15f929b114cb)** &mdash; First occurred on 2019-07-29 13:06:20 -0700

```
Unauthorized
```

*Brought to you by Haystack's `!` hotkey.  Try it on a needle or rollup today!*

stephaniegiang:
## What
Fixes #2116 

This PR fixes the edge case when the user deletes GitHub Classroom from their authorized application.

Before, this would cause a 500 since we think that the user is authenticated when methods like `GoogleClassroomConfiguration#index`. 

So this PR makes the user to re-authenticate when we hit the issue.


app/controllers/orgs/google_classroom_configurations_controller.rb
```diff
@@ -29,6 +29,15 @@ def search
 
     def index
       @google_classroom_courses = fetch_all_google_classrooms
+    rescue Google::Apis::AuthorizationError
+      google_classroom_client = GitHubClassroom.google_classroom_client
+      login_hint = current_user.github_user.login
+      redirect_to google_classroom_client.get_authorization_url(login_hint: login_hint, request: request)
+      nil
```
srinjoym: Do we need this nil here? I think this was there before to return nil from a helper method

app/controllers/orgs/google_classroom_configurations_controller.rb
```diff
@@ -29,6 +29,15 @@ def search
...
+    rescue Google::Apis::ServerError, Google::Apis::ClientError
+      flash[:error] = "Failed to fetch classroom from Google Classroom. Please try again."
+      redirect_to organization_path(current_organization)
+      nil
```
srinjoym: Same as above ^

app/controllers/orgs/google_classroom_configurations_controller.rb
stephaniegiang: Ahh good catch. Yea we don't need this. I'll remove it.

PR #2122, opened 2019-07-30
Detect/redirect from iFrame to separate window on embedded LTI Launch

thebradbain:
## What
All of our support docs instruct upon configuring GitHub Classroom within an LMS to open in a new window, but some LMSs (Blackboard & Brightspace 👀 )  default to loading third party applications within iFrames and make it confusing to change from a user-standpoint. 

Pre-change, GitHub Classroom would obviously refuse to even respond to the request for security reasons, but would also probably leave the user confused in this case. Post-change, just for the `Sessions#lti_launch` action, we render a special splash page, which, when in an iFrame, will direct the user to launch in a separate tab instead of failing spectacularly as it is does now. 

From a security-perspective, which I'm _not_ qualified to talk about 😅, I don't believe this would(?) open us up for attacks, as no sensitive information is exposed or rendered (and the template itself doesn't use any of the traditional layout templates, so is fully self-contained), and a request within the iFrame to any other endpoint would still be refused. Tagging appsec to get official input ✨ 

Pre-change:
Screen Shot 2019-07-29 at 11 17 17 PM

Post-change:
Screen Shot 2019-07-29 at 11 16 22 PM


cktricky:
Hi @thebradbain 👋,

Do we know which top level domains that we'll be iframe'd for in advance? Its fine if we don't know the subdomain in advance (ex: `some-random-subdomain-we-wont-know-in-advance.blackboard.com`). I'm just wondering if we might have a more targeted solution using Content-Security Policy headers. 

Specifically the [`frame-ancestors` solution](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) where we can designate entire subdomains (ex: `frame-ancestors: https://somethingrandom.blackboard.com`).

Everything else makes sense such as skipping the CSRF check for that one endpoint as well as the way you did it (❤️ for using `:only` versus `:except` 😄).

cktricky:
I've updated the frame-ancestors example, I realized I forgot to put in the wildcard `*` for the subdomain in my example 🤦‍♂.

thebradbain:
@cktricky I never even knew about the `from-ancestors` CSP settings, that's awesome! Unfortunately, most LMSs offer self-hosted solutions that use custom domains set by the institution (fro example, my school's `sakai.claremont.edu`) -- and on top of that, GitHub Classroom will support any LMS which follows the LTI standard, so we don't necessarily have any whitelist (but on `lti_launch`, we do check if it's a valid LTI launch per the standard, so not just any request is allowed through).

cktricky:
Ahh, okay in that case I don't think we really have any other choice but to allow that specific endpoint to be iFramed 👍.

stephaniegiang:
Can we instead of using the GitHub logo use the GitHub Classroom logo?

thebradbain:
I'm currently using the GitHub Octicon icon, but I don't see why not? Is there a logo we have that's about this size?

stephaniegiang:
Yea we have one here we can resize 
`app/assets/images/classroom-logo.png`

thebradbain:
Screen Shot 2019-08-02 at 4 47 17 PM

Updated -- good call @stephaniegiang, it looks great

PR #2127, opened 2019-07-30
Add ability to change visibility of all assignments/repos in a classroom (back-end changes)

### LINKED ISSUE #2031: Option to publicize all student repos
Related to #1996.

Teachers should have the ability to publicize all student assignment repos. A use-case is after the course is over, the teacher may allow the students to show off/access their work on their GitHub profiles.

A positive side-effect would be that this would make it easy to also privatize all student repos and add the ability to publicize/privatize the assignment and its repos _after_ repos have already been created (e.g. if teacher accidentally created the assignment as public and wants to switch to private).

EricPickup:
Second part that closes #2031 (front-end changes merged in #2121).

## What
This PR adds the back-end changes that allow us to change the visibility of an entire classroom (all of its assignments and student repos).

It adds the `ClassroomVisibilityJob` which iterates through all of the assignments and modifies the visibility attribute, then all of the assignment's repos and changes the visibility through the API.


app/jobs/organization/classroom_visibility_job.rb
```diff
@@ -0,0 +1,29 @@
...
+class Organization
+  class ClassroomVisibilityJob < ApplicationJob
+    queue_as :classroom_visibility
+
+    # rubocop:disable SkipsModelValidations
+    def perform(organization, visibility)
+      public_classroom = (visibility == "public" ? true : false)
+      assignments = organization.assignments + organization.group_assignments
+      assignments.each do |assignment|
+        assignment.update_attribute("public_repo", public_classroom)
```
EricPickup: FYI this update method skips validation. I'm doing this due to cases where another attribute has become invalid in the assignment (e.g. starter code repo was deleted) we would not be able to update the visibility.

d12: How does the starter code repo being deleted make the record invalid? The ID will still be a sensible ID as far as AR is concerned. We should avoid skipping validations, this can be dangerous.

app/jobs/organization/classroom_visibility_job.rb
```diff
@@ -0,0 +1,29 @@
...
+        change_visibility_in_all_assignment_repos(assignment, public_classroom)
+      end
+    end
+
+    private
+
+    def change_visibility_in_all_assignment_repos(assignment, public_repo)
```
d12: `public_repo` doesn't really imply this is a boolean, something like `is_public` may be better.

EricPickup: @d12 you're right, I used a bad example. Consider old records which haven't been validated with newer validators (for example, empty starter code repos) or a template repo which was changed to a regular repo. When updating the public/private attribute with validation it will fail (starter code repo is not a template error).

EricPickup:
@d12 really good point. 

According to [the docs](https://developer.github.com/v3/#rate-limiting) all requests using OAuth are limited to `5000 requests per hour`, so the speed at which we're changing these repos doesn't really matter. When a user tries to change the visibility we can `GET /rate_limit` and determine if they have enough requests left to complete this job and flash an error if they can't.

What do you think of this solution?

app/jobs/organization/classroom_visibility_job.rb
EricPickup: @d12  I used a bad example. Consider old records which haven't been validated with newer validators (for example, empty starter code repos) or a template repo which was changed to a regular repo. When updating the public/private attribute with validation it will fail (starter code repo is not a template error).

EricPickup: @d12 this issue will also come up here when we disable assignment invitations when we archive classrooms since we'll have to update the `invitations_enabled` column on the assignment: https://github.com/education/classroom/issues/2139

d12: I spoke with Eileen here https://github.slack.com/archives/C0HV3F37A/p1564670196110900 about this, I don't think skipping validations is the solution here. If there are validations introduced that invalidate old records, that's a problem and we need to update those validations.

I noticed a couple of our assignment validations actually make API calls, and get hit on every update which probably slows down the app a lot. Do we need to be checking if the starter code repo is empty on every update? Same with the template check. Just on creation should be fine, and then if we error while trying to generate a repo, we know that the status has changed.

I updated `updated_at` on one of my assignments and tried saving it, it takes 800ms:

```ruby
PRODUCTION [27] pry(main)> Benchmark.ms { a.save }
D, [2019-08-01T15:16:51.635549 #4] DEBUG -- :    (1.0ms)  BEGIN
D, [2019-08-01T15:16:51.649581 #4] DEBUG -- :    (2.9ms)  SELECT  "users"."token" FROM "users" INNER JOIN "organizations_users" ON "users"."id" = "organizations_users"."user_id" WHERE "organizations_users"."organization_id" = $1 ORDER BY RANDOM() LIMIT $2  [["organization_id", 30645], ["LIMIT", 1]]
D, [2019-08-01T15:16:51.654531 #4] DEBUG -- : Cache read: 577d044a0bea470aa51718d9a6242f96345447e5
D, [2019-08-01T15:16:51.798154 #4] DEBUG -- : Cache read: 577d044a0bea470aa51718d9a6242f96345447e5
D, [2019-08-01T15:16:51.807274 #4] DEBUG -- : Cache write: 577d044a0bea470aa51718d9a6242f96345447e5
D, [2019-08-01T15:16:51.992203 #4] DEBUG -- :   User Load (2.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 132810], ["LIMIT", 1]]
D, [2019-08-01T15:16:51.997458 #4] DEBUG -- :   Assignment Exists (3.9ms)  SELECT  1 AS one FROM "assignments" WHERE "assignments"."title" = $1 AND ("assignments"."id" != $2) AND "assignments"."organization_id" = $3 LIMIT $4  [["title", "foo"], ["id", 63495], ["organization_id", 30645], ["LIMIT", 1]]
D, [2019-08-01T15:16:52.001876 #4] DEBUG -- :   Assignment Exists (3.1ms)  SELECT  1 AS one FROM "assignments" WHERE "assignments"."slug" = $1 AND ("assignments"."id" != $2) AND "assignments"."organization_id" = $3 LIMIT $4  [["slug", "foo"], ["id", 63495], ["organization_id", 30645], ["LIMIT", 1]]
D, [2019-08-01T15:16:52.006004 #4] DEBUG -- :   AssignmentInvitation Load (1.2ms)  SELECT  "assignment_invitations".* FROM "assignment_invitations" WHERE "assignment_invitations"."deleted_at" IS NULL AND "assignment_invitations"."assignment_id" = $1 LIMIT $2  [["assignment_id", 63495], ["LIMIT", 1]]
D, [2019-08-01T15:16:52.013544 #4] DEBUG -- :   GroupAssignment Load (4.1ms)  SELECT "group_assignments".* FROM "group_assignments" WHERE "group_assignments"."deleted_at" IS NULL AND "group_assignments"."slug" = $1 AND "group_assignments"."organization_id" = 30645  [["slug", "foo"]]
D, [2019-08-01T15:16:52.018158 #4] DEBUG -- : Cache read: a373842a89aa87e0e26c1522806a37429bbb0d64
D, [2019-08-01T15:16:52.151738 #4] DEBUG -- : Cache read: a373842a89aa87e0e26c1522806a37429bbb0d64
D, [2019-08-01T15:16:52.157859 #4] DEBUG -- : Cache write: a373842a89aa87e0e26c1522806a37429bbb0d64
D, [2019-08-01T15:16:52.174458 #4] DEBUG -- : Cache read: 121b52507195a0b2dc569ca2bb06b79869abbc62
D, [2019-08-01T15:16:52.269993 #4] DEBUG -- : Cache read: 121b52507195a0b2dc569ca2bb06b79869abbc62
D, [2019-08-01T15:16:52.272245 #4] DEBUG -- : Cache write: 121b52507195a0b2dc569ca2bb06b79869abbc62
D, [2019-08-01T15:16:52.423626 #4] DEBUG -- :   SQL (5.9ms)  UPDATE "assignments" SET "updated_at" = $1 WHERE "assignments"."id" = $2  [["updated_at", "2019-08-01 15:16:47.070757"], ["id", 63495]]
D, [2019-08-01T15:16:52.429015 #4] DEBUG -- :    (4.2ms)  COMMIT
=> 794.9617860140279
```

What do you think about changing these validations to on_create?

d12:
a) Classes with more than 5k assignment repos will never be able to convert repos to public/private, and there are classes with more than 5k repos :/

b) If we suck up all of the teachers rate limit, the app will stop working for them.

We need a way of doing this in a way that solves both of the above problems. What if we only performed the operation on 100-500 repos at a time (check how many API calls are made in this operation, GET requests are counted in the API rate limit), then enqueue a job to start working on the next 100-500 in an hour. When they choose to publicize/privatize repos, we can give them an estimate of how long it will take (we know how many repos they have, and we know how many we process per hour), and we don't allow them to publicize/privatize while the operation is in process.

Let me know what you think, I'm also open to other solutions, but we need to be able to support larger classrooms here :)

andrewbredow:
Closing this PR since #2031 is in the icebox.

PR #2133, opened 2019-07-31
Fix bug when there is an error on classroom settings

### LINKED ISSUE #2086: Styling breaks when there's an error on classroom settings form
**Describe the bug**
We have some style issues when an error comes up on the manage classroom page.

**To Reproduce**
1. Go to "Manage classroom"
2. Make the classroom title blank
3. Save

**Screenshots**
**Before:**
image

**After:**
image

Notice the manage classroom button is off, and the banner is quite different in size and content.

GQuirino:
fix #2086 
Bug: In the banner, the title of the organization is being erased if receive some validation error, because of that missing information the height of that banner was lower than the usual.

fix: before render the edit page, reload the current organization to get the last information saved


GQuirino:
That field title, in classroom settings, is getting the organization title as default, should i set the default value as empty?

stephaniegiang:
@GQuirino For that page, we should grab the original classroom name and org name.

stephaniegiang:
Oh also, ping us when this is ready for review 😄

d12:
Thanks for the contribution! One thing I'd like to add, are you able to write a regression test for this?

GQuirino:
@d12 sure !!

GQuirino:
@12 test done!

stephaniegiang:
You have some syntax issues so CI is failing.

But you can run `rubocop -a spec/controllers/organizations_controller_spec.rb` and that command will automatically fix it for you. Then commit those changes. I don't have push permissions.

GQuirino:
@stephaniegiang syntax fixed :smile: :+1:

spec/controllers/organizations_controller_spec.rb
```diff
@@ -285,11 +285,22 @@
...
+        expect(response).to redirect_to(organization_path(Organization.find(organization.id)))
+      end
+    end
+
+    context "when fail" do
+      it "not change current organization title" do
+        options = { title: " " }
+        patch :update, params: { id: organization.slug, organization: options }
+
+        expect(assigns(:current_organization).title).not_to eql(options[:title])
```
d12: Can you add one final assertion in here showing we redirect to :edit? Then I think this will be good to go :sparkles:

GQuirino: all done

d12:
It looks like your tests are failing on some cassette errors. Let us know if you need help resolving those.

GQuirino:
i updated the cassettes, ready to review

PR #2151, opened 2019-08-01
Update Link LMS to match new style

stephaniegiang:
## What
Updates the Link LMS page to match GitHub.com's style and the current on-boarding style. This way the user isn't presented with a sudden UI change. 

image

For reference here is the older style:
image


d12:
Can we go ahead and remove Blackboard here?

stephaniegiang:
Done!

PR #2152, opened 2019-08-01
Use the correct Result class when redeeming AssignmentInvitation

shaunakpp:
## What

With the new `unified_repo_creators` flag enabled, this PR enables us to use the correct `Result` class based on whether the flag is enabled or not.


shaunakpp:
Closing this as the new pull-request will remove this code

PR #2153, opened 2019-08-01
Run API-calling validators only when related attribute is changed

EricPickup:
## Problem
We have validators such as `starter_code_repository_not_empty` and `starter_code_repository_is_template` that need to call the API to run properly. These validators are run every single time an assignment is created or updated (regardless of whether the starter code repo is changed or added).

## Solution
Run these validators only when `starter_code_repo_id` is added or changed.

I ran a few before/after benchmarks for a simple update to an `Assignment` and it went from `800ms` to `30ms`. 😄 

Side note: Moved `starter_code_repository_not_empty` into the `StarterCodeImportable` concern to get rid of duplicate methods in `Assignment`/`GroupAssignment`.


shaunakpp:
Wow! Nice catch! 👏

EricPickup:
@shaunakpp @d12 caught it actually 👏

shaunakpp:
We should evaluate other such API calls across the app too...

app/models/assignment.rb
```diff
@@ -38,9 +38,8 @@ class Assignment < ApplicationRecord
   validates :assignment_invitation, presence: true
 
   validate :uniqueness_of_slug_across_organization
-  validate :starter_code_repository_not_empty
-
-  validate :starter_code_repository_is_template
+  validate :starter_code_repository_not_empty, if: :will_save_change_to_starter_code_repo_id?
```
d12: Wow, til these `will_save_change_to_starter_code_repo_id?` methods. Nice find :sparkles:

shaunakpp: Rails magic ✨

PR #2143, opened 2019-08-01
Disable invitations in all assignments when archiving classroom

EricPickup:
## What
Closes #2139.

When we archive a classroom we should disable assignment invitations for all assignments.

I've included this in the button description for archiving/unarchiving.

image
image

PR #2155, opened 2019-08-02
Add confirmation modal for archiving classrooms

### LINKED ISSUE #2145: Add prompt for classroom archiving
We should add a confirmation prompt modal when archiving classroom since it makes some irreversible (can be fixed manually) changes by disabling assignment invitations (#2143).

EricPickup:
## What
Closes #2145.

#2143 gives us enough reason to add a confirmation modal to archiving classroom (we already do this with all other buttons in the danger zone).

image


d12:
Disregard my comment in #2143 , it looks like we won't be re-enabling invitations? Why is that?

EricPickup:
@d12 When a teacher archives a classroom we know that they're done with that entire classroom - it makes sense to disable invitations. When they un-archive the classroom we don't know that they want the entire classroom and all of its assignments to be re-enabled.

I'm just imagining a classroom where they have invitations disabled for a couple assignments that are old. If a teacher unintentionally or intentionally archives a classroom and later tries to un-archive it, they probably don't want us to enable those assignments that were previously disabled before the archive.

Does that make sense?

d12:
Yeah that totally makes sense. So maybe to solve both problems, there's a more simple solution here that requires no data changes to disable assignment invitations.

When checking if an invitation is enabled, that can be a product of `(assignment.invitations_enabled? && !organization.archived?)`.  Now when a teacher un-archives, they get back to the exact same configuration they had before without us needing to do special work to "remember" what the state was before we went and changed everything.

This implementation also gives us the ability to show a different error screen if the class is archived, something like "Your teacher has archived this classroom, so assignments cannot be accepted anymore"

EricPickup:
@d12 genius 🙌 I'll get to it!

EricPickup:
@d12 Your suggestions are added in #2160 and I've updated this PR and the photo to reflect that 😄

PR #2158, opened 2019-08-02
First iteration for removing Repo creators - Remove creator use from specs

shaunakpp:
## What

In this pull request, I've replaced the `AssignmentRepo::Creator` and `GroupAssignmentRepo::Creator` use from all the specs in favor of the new `CreateGitHubRepoService`.

Since the use of these creators was supplementary in the specs, we can safely replace it with the unified creator.


app/controllers/group_assignment_invitations_controller.rb
```diff
@@ -181,7 +181,6 @@ def create_group_assignment_repo(selected_group: group, new_group_title: nil)
       selected_group,
       new_group_title
     )
-
```
d12: This delta probably shouldn't be in this PR.

shaunakpp: I'll fix this 👍

shaunakpp:
I was making changes to the specs, and was getting `resource not found` errors for the team with id `2_973_107`, so I had to change the specs 😢

PR #2161, opened 2019-08-02
Removes two feature_flags: multiple_orgs_on_classroom and search_assignments

shaunakpp:
## What
We have the above feature flags in production for a while now, and its safe to move these features out of the feature flag.


shaunakpp:
TIL: if you add back-ticks around something in commits, they disappear from the message! 🤦‍♂

app/views/organizations/new.html.erb
```diff
@@ -16,7 +16,7 @@
       <div class="d-flex flex-wrap gutter-spacious">
         <% @users_github_organizations.each do |org| %>
           <div class="col-6 col-md-4 col-lg-3 mb-6">
-            <% if org[:role] == 'admin' && (!org[:classroom].present? || multiple_classrooms_per_org_enabled?) %>
```
stephaniegiang: Does removing `!org[:classroom].present?` do anything? Just curious on what that line was for.

shaunakpp: I did not notice any difference in behavior after removing it, but maybe @srinjoym can help us out here

PR #2167, opened 2019-08-05
Add LMS user id field on Roster Entries table

stephaniegiang:
## What
For the Google and LMS integration, we are focusing on allowing duplicate entries. Currently with LMS, we cannot tell if the roster entry has been imported from the LMS so we will keep adding the same student to the roster. 

To fix that we are adding a field that can keep track to see if the Roster Entry was imported from the LMS so we don't keep adding duplicates to the Roster. That way we can also map the student to their roster entry as well.


thebradbain:
I especially like this migration because we can move the roster's google specific `google_classroom_student_id` to be a more general `student_id` that both LTI and Google Classroom use to keep track of which roster entry is which when it comes to avoiding duplicates

d12:
Yeah, this is good, I'd like to see this replace the google specific field. This will be especially useful when we get into stuff like grade syncing, assignment syncing, etc. where we need a mapping from Classroom student to LMS student.

PR #2168, opened 2019-08-05
Don't show "Import by" option to user if that attribute type is all nil

thebradbain:
## What
Small change that _proactively_ lets the user make an informed decision regarding how to import their roster from an LMS: before, even if the LMS didn't give us student emails, we were allowing the user to import by student email and then throwing an error when nothing was created. Now, we simply don't let the user choose that option if we know it's something they can't import users by.

**Current**
_E.g. if names are not available we still allow users to import by name_
Screen Shot 2019-08-05 at 9 42 31 AM

**New**
_E.g. if names are not available we do not allow users to import by name in the first place_
Screen Shot 2019-08-05 at 9 41 24 AM


app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -17,10 +17,17 @@ def import_from_lms
...
         "Names": students.map(&:name),
         "Emails": students.map(&:email)
-      }
+      }.select { |_, v| v.any? }
 
       GitHubClassroom.statsd.increment("lti_configuration.import")
 
+      if @identifiers.empty?
+        msg = "GitHub Classroom received an empty roster from your Learning Management System. Please ensure you
+        have enabled GitHub Classroom to receive student information from within your Learning Management System."
```
d12: Are we able to reword this error to be more actionable? WHat does "enable GitHub Classroom to receive student information" mean? Do they need to enable the membership service?

thebradbain: Yeah, instead of an error message here I was instead thinking of actually changing the view so that instead of a select list on an empty response, we explained in more detail what the problem was and later possibly link them to support docs.

The error here isn't that the membership service isn't enabled or the request failed (in fact this screen will be reached only when it receives a successful response from the membership service), it's that the membership service _is_ enabled but the teacher set the privacy settings (or they are that way by default) such that individual student information isn't being sent in the roster, so it's essentially an empty response.

d12: Gotcha. If you'd like to ship this as a V1 then iterate, that sounds :+1:, but :100: to explaining the error in more detail :)

PR #2176, opened 2019-08-06
Support LTI 1.0 responses

### LINKED ISSUE #2179: Add LTI 1.0 Support
Part of the LTI 1.1 certification requires being backwards-compatible with LTI 1.0 messages, even though 1.1 has a different message response format (replaces XML with JSON, and whether the OAuth signature is in the request headers or the body). Fortunately it's a pretty straightforward fix, and we will now have out-of-the-box support for LMSs which haven't yet implemented LTI 1.1 but do support LTI 1.0 (the biggest being Sakai)

See PR #2176

thebradbain:
## What
Closes #2179.

Refactors the `LTI::Mixins::RequestSigning` into a testable mixin, and also adds support for LTI 1.0 tool-originating requests (which are completely different response format than LTI 1.1 tool-originating requests, go figure). 

Also uses the `IMS::LTI` gem's robust built-in parsing capabilities, so we don't have to rely on wrangling LMS responses into models on our own, as we were previously. 

With this change, we now have built-in support for LMSs which send roster in the form of LTI 1.0 responses, such as Sakai.


thebradbain:
Aside from rubocop errors, this should be ready for review!

Issue #1942, opened 2019-07-02
Adding a numerical suffix to the end of duplicate roster entries


stephaniegiang:
For Google Classroom, sometimes we may have students with the same first and last name which causes a bug in the roster creation. In the controller, this will cause only one user to be created as we do not allow duplicates.

A fix for this is when a duplicate entry is found, we add a numerical suffix such as `-1` to the end of the entry. We should also allow teachers to edit the roster entries after they are imported from a LMS or Google Classroom.

stephaniegiang:
Done in #2177 and #2178

PR #2177, opened 2019-08-06
Allow duplicate roster entries

stephaniegiang:
## What
Part 1 of #1942 

j6H6pCZ4O9

In preparation for Google and LMS integration, we are going to be allowing duplicate entries for rosters. If teachers chose to import students using their full names and there are two students with the same name, it will result to only one identifier being added. This PR fixes this. 

When a duplication is found, we will begin appending a numerical suffix.

For example, if `John Smith` is already in the roster and we are adding another `John Smith`, we will then append `-1` and the final identifier being `John Smith-1`.


app/models/roster/creator.rb
```diff
@@ -74,18 +74,33 @@ def perform
...
+        identifier = check_for_duplicates(@roster.roster_entries, identifier)
+
+        @roster.roster_entries << RosterEntry.new(
+          roster: @roster, identifier: identifier, google_user_id: google_user_id
+        )
+        @roster.save!
       end
     end
 
+    def check_for_duplicates(roster_entries, identifier)
```
d12: Are you able to bring this logic together in one place instead of duplicating in two places?

I think a good interface would be a method that takes a list of identifiers, and outputs a list of identifiers with duplicates given a suffix. Then we could call this on line 82 like

```ruby
identifiers = add_suffix_to_duplicates(identifiers)
```

stephaniegiang: I'll look into this 👍

stephaniegiang:
@d12 ready for another review! Moved the duplicate method into a concern.

app/models/concerns/duplicate_roster_entries.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module DuplicateRosterEntries
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    # rubocop:disable Metrics/MethodLength
+    def add_suffix_to_duplicates(identifiers: [], roster_entries: [])
```
d12: What is the purpose of `roster_entries:` here? It looks like this method mutates the argument which would be unexpected for a method without a `!` in the signature. Do we need to mutate anything here?

stephaniegiang: The reason I added the field `roster_entries` was to pass in any existing roster entries that are on the roster. This way we can take them into consideration for duplicates. This is mainly only used for adding more students into the roster or syncing from integrations. 

I just changed the code a little bit, so that we are not mutating `roster_entries`. Also that field is now renamed to `existing_roster_entries` so that is is more clear on what that variable does. Instead of mutating, I added `list_of_identifiers` and `existing_roster_entries` together so we can also take into consideration the new identifiers that are being added.

stephaniegiang:
@d12 ready for another review!

PR #2178, opened 2019-08-06
Allow editing of roster entries

stephaniegiang:
## What
Part 2 of #1942
Branching off of #2177 (Needs to merge after that PR)

gb4HFyA4EW

This PR allows teachers to edit existing roster entries. It checks to make sure when updating the identifier, that there are no duplicates. 

Would love some feedback on UI, the copy of errors, and copy of text.


d12:
A couple of questions:

1) Is the edit button on linked + unlinked roster entries?
2) If there's an error while editing, does the form scroll back to the top? Or if we aren't on pg 1, does it jump back to page 1? We've introduced issues like that in the past and it makes it quite frustrating when trying to edit a large list.

stephaniegiang:
@d12 
1. The edit button will show up for linked and unlinked roster entries. 
image

2. Oh that is a good point. Right now it jumps back to page 1. I'll change this so that it stays on the current page.

d12:
Great! Let me know when this is ready for review again.

stephaniegiang:
@d12 fixed the reloading. It now loads the page that the user was on

db/schema.rb
```diff
@@ -232,7 +232,9 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "google_user_id"
+    t.string "lms_user_id"
     t.index ["google_user_id"], name: "index_roster_entries_on_google_user_id"
+    t.index ["lms_user_id"], name: "index_roster_entries_on_lms_user_id"
```
d12: Looks like there's DB changes in the PR! I'm assuming these aren't supposed to be here, please ping me when they're cleaned up :)

stephaniegiang: Opps thanks for catching that! Ready for another review!

stephaniegiang:
@d12 ready for another review!

srinjoym:
@stephaniegiang Not sure if that has been discussed but it might be easier to make the fields editable inline instead of opening a modal for each student. It would be similar to how the GitHub profile bio section works:

image

d12:
This would be cool, how much extra work would it be to ship this? We can also ship this version then iterate.

stephaniegiang:
@d12 Looking into this right now. Not sure where this sits in priority. Do we want this merged before the LMS integration launch?

d12:
Good point, let's ship as is, and hopefully ship @srinjoym 's version next week? :+1:

PR #2181, opened 2019-08-06
Start using "lms_user_id"

stephaniegiang:
## What
This PRs changes the use of the field of `google_user_id` to `lms_user_id`.

This PR also makes the `lms_user_id` field to be used for LMS integrations.


stephaniegiang:
LMSs don't have specific tests for checking the roster attributes since it's hard to stub. But for Google, I can add tests to these. We are only checking for the roster identifier to be the same.

stephaniegiang:
@BenEmdon we don't test for the `google_user_id`s since when stubbing out the response, they don't give us a `user_id` to test with. So no there are no tests that can test this.

app/views/orgs/rosters/import_from_lms.html.erb
```diff
@@ -15,6 +15,7 @@
               <dt class="Box-subtitle">Select from the choices below</dt>
             </dl>
             <ul>
+            <%= hidden_field_tag "lms_user_ids", @identifiers.[:"User IDs"] %>
```
d12: Could you explain what's going on in `@identifiers.[:"User IDs"]`?

stephaniegiang: Yeah, it looks weird. That is because of how we are setting `@identifiers`. 

https://github.com/education/classroom/blob/master/app/controllers/orgs/rosters_controller/lti_dependency.rb#L17-L21

stephaniegiang:
PR is not ready yet for review. I have one more thing I want to add. Will ping when ready 😄

app/views/orgs/rosters/import_from_lms.html.erb
stephaniegiang: Changed the approach of this, so it is more readable

thebradbain:
Leaving this message here just in case it gets lost in Slack, but:

You’ve made a few changes to the LMS roster import error modal that don’t at first glance seem to be tied to the actual roster creation but rather the handling of roster import errors. Could you revert those? I’ve got a PR out in #2202 dealing with revamping errors messages and it’ll be easier to avoid the merge conflicts in the first place than deal with them after the fact :)

PR #2187, opened 2019-08-06
Add parent class to support bulk API jobs and scheduling

### LINKED ISSUE #2184: Add support for bulk API-calling jobs
The after-life project will introduce a lot of bulk changing features (publicize **all** repos in a classroom, transfer ownership, archive, etc.). We also already have the `RepositoryVisibilityJob` (https://github.com/education/classroom/blob/master/app/jobs/assignment/repository_visibility_job.rb) that changes the visibility of all repos in an assignment when we switch the assignment between public/private.

We need to be able to support the largest orgs in Classroom. Imagine a course that has 1k students per assignment. If we want to change the repo visibility for all repos in the classroom, that's 1k * ~2 calls to change visibility * x assignments = potentially thousands of API calls. The current rate limit for authenticated API calls is 5000 per hour.

EricPickup:
Closes #2184. More context here: https://github.com/education/classroom/pull/2127#issuecomment-517335542

## The Problem

Over the next few weeks we'll be introducing a lot of jobs that make changes to student repos in bulk. We already have the `RepositoryVisibilityJob` (https://github.com/education/classroom/blob/master/app/jobs/assignment/repository_visibility_job.rb) that can be adjusted to use this. I plan to introduce `ClassroomVisibilityJob` (#2127), `TransferOwnershipJob`, `ArchiveRepositoriesJob`, etc. that will execute thousands of API calls on behalf of a teacher for large classrooms. 

It makes sense to create a parent class that does all of the error handling, rescheduling and prevents us from executing bulk API jobs in parallel.

## The Solution

This PR adds the `BulkApiJob` class which all bulk jobs should inherit. It's essentially a queue that forces a 1-hour interval between each bulk API job and prevents the User from running another in parallel.

When a bulk job is enqueued, it will check the `User`'s cooldown time. If it does not exist or has passed, it will set the cooldown for 1 hour from now and perform the job. If the cooldown is still in effect, it will schedule the job for after the cooldown.

If a job fails mid-execution because of the rate limit (`GitHub::Forbidden`) it will be re-scheduled similarly (added to the end of the queue).

twTRkxSHqA
_This case shows how Teacher (ID: 12345) splits the ClassroomVisibilityJob into 3 jobs to cover all 453 repos. The second fails mid-execution and is rescheduled._

### DIFF
--- a/app/jobs/bulk_api_job.rb
+++ b/app/jobs/bulk_api_job.rb
@@ -33,10 +33,10 @@ class BulkApiJob < ApplicationJob
 
     if cooldown.nil? || Time.zone.now > cooldown
       job.scheduled_at = Time.zone.now
-      GitHubClassroom.redis.set("user_api_job:#{user.id}", (Time.zone.now + 1.hour).to_datetime)
+      GitHubClassroom.redis.set("user_bulk_job_cooldown:#{user.id}", (Time.zone.now + 1.hour).to_datetime)
     else
       job.scheduled_at = cooldown.to_i
-      GitHubClassroom.redis.set("user_api_job:#{user.id}", (cooldown + 1.hour).to_datetime)
+      GitHubClassroom.redis.set("user_bulk_job_cooldown:#{user.id}", (cooldown + 1.hour).to_datetime)
     end
   end
 end
--- a/app/jobs/bulk_api_job.rb
+++ b/app/jobs/bulk_api_job.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class BulkApiJob < ApplicationJob
+  class Error < StandardError
+    class JobAlreadyRunning < Error
+      def message
+        "User is already running a bulk API job. Users can only run one bulk job at a time."
+      end
+    end
+
+    class MissingUser < Error
+      def message
+        "First argument in perform method must be the User performing the job."
+      end
+    end
+  end
+
+  queue_as :bulk_api_job
+
+  rescue_from(GitHub::Forbidden, Error::JobAlreadyRunning) do
+    user = arguments.first
+    retries = arguments.last[:retries] if arguments.last.is_a? Hash
+    arguments.last[:retries] -= 1 if retries
+    api_wait_time = user.bulk_api_job
+
+    self.class.set(wait_until: api_wait_time).perform_later(*arguments) if retries && retries.positive?
+  end
+
+  before_perform do |job|
+    user = job.arguments.first
+
+    raise Error::MissingUser unless user.is_a? User
+    raise Error::JobAlreadyRunning if user.running_bulk_api_job?
+
+    GitHubClassroom.redis.set("user_api_job:#{user.id}", (Time.zone.now + 1.hour).to_datetime)
+  end
+end
--- a/app/jobs/bulk_api_job.rb
+++ b/app/jobs/bulk_api_job.rb
@@ -2,15 +2,15 @@
 
 class BulkApiJob < ApplicationJob
   class Error < StandardError
-    class JobAlreadyRunning < Error
+    class ArgumentError < Error
       def message
-        "User is already running a bulk API job. Users can only run one bulk job at a time."
+        "Invalid arguments. Arguments should be provided as a hash."
       end
     end
 
-    class MissingUser < Error
+    class UserNotFound < Error
       def message
-        "First argument in perform method must be the User performing the job."
+        "User performing the bulk API job must be provided."
       end
     end
   end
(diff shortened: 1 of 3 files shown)


app/jobs/bulk_api_job.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  queue_as :bulk_api_job
+
+  rescue_from(GitHub::Forbidden, Error::JobAlreadyRunning) do
+    user = arguments.first
+    retries = arguments.last[:retries] if arguments.last.is_a? Hash
+    arguments.last[:retries] -= 1 if retries
+    api_wait_time = user.bulk_api_job
+
+    self.class.set(wait_until: api_wait_time).perform_later(*arguments) if retries && retries.positive?
```
shaunakpp: we can merge `retries && retries.positive?` to `retries&.positive?`

app/jobs/bulk_api_job.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  before_perform do |job|
+    user = job.arguments.first
+
+    raise Error::MissingUser unless user.is_a? User
```
shaunakpp: This assumes that `BulkApiJob` will always be a at a `User` level, so we can't use this for running a `organization` level bulk API job, while this is okay for the first iteration, but can we figure out a way to make it use `Organization` or `User`. Or somehow abstract this out?

shaunakpp: Like adding a `BulkApiJobSupportable` module to a model, and adding methods like `supports_bulk_api_jobs?` for checking support.

app/models/user.rb
```diff
@@ -64,6 +64,17 @@ def api_token(exp = 5.minutes.from_now)
     MessageVerifier.encode({ user_id: id }, exp)
   end
 
+  def running_bulk_api_job?
+    redis = GitHubClassroom.redis
+    existing_api_job = redis.get("user_api_job:#{id}")&.to_datetime
+    existing_api_job && existing_api_job.future?
```
shaunakpp: can be merged to one statement as above

app/jobs/bulk_api_job.rb
```diff
@@ -0,0 +1,37 @@
...
+      def message
+        "First argument in perform method must be the User performing the job."
+      end
+    end
+  end
+
+  queue_as :bulk_api_job
+
+  rescue_from(GitHub::Forbidden, Error::JobAlreadyRunning) do
+    user = arguments.first
```
shaunakpp: this restricts the first argument to be a `User`, maybe we could consider using keyword arguments, so we are not dependent on `.last` and `.first` methods?

EricPickup:
Thanks for the great reviews @shaunakpp, I'll make some adjustments! ❇️ 

Going to add this as a WIP since I just noticed a flaw and have to make some significant changes to the logic

shaunakpp:
Would love to pair on this, let me know 👍

EricPickup:
I've updated this so that all the checks are done before _enqueueing_ the job, rather than before _performing_.

When a job is being enqueued via `perform_later` it will automatically check the cooldown time and schedule the job for after the cooldown time. It will also add an hour to that cooldown time now since this job will execute then.

app/jobs/bulk_api_job.rb
EricPickup: Correct me if I'm wrong, but in the end the token calling the API and the rate limit will always belong to a User.

shaunakpp: Yes, thats true, but if it is a `Organization` level job, then like we do now, we can pick a random admin user and use his/her access_token for the job or we can distribute the API calls using tokens of _all_ the admins. This may help us achieve two things:

1. Not exhaust any one particular User's rate limit
2. Faster execution of the `BulkApiJob` as we're using more than one user's token

TLDR; In the future iterations, we could consider passing an organization, then randomize( or distribute) the token selection in order to not exhaust rate limits.

What do you think?

EricPickup: Sounds like a great idea 👍 This is something we should definitely add on a future (if not next) iteration

d12:
This is a good one for someone on ce-education to review, it's very Ruby-y and not dependent on any domain knowledge.

cc @jeffrafter @jessrudder

jeffrafter:
Just wanted to note here that @EricPickup and I talked this one over and he is experimenting in a couple other directions currently... we may come back to this one.

jeffrafter:
For the moment we are going to pursue some additional strategies around rate-limiting and will close this. We may reopen it along the way.

PR #2189, opened 2019-08-07
Front-end changes to archived classrooms

EricPickup:
## What
This PR adds a `Archived` label on an archived classroom's page and makes their banner color gray to easily distinguish between active and archived classrooms.

I considered adding a label on the organization card layout but this week I'll be changing the dashboard layout so that archived classrooms are in a separate section and won't appear next to active classrooms.

image

image

N6DTYRsvSp


d12:
Can we change that flash message when we update to be a little more user friendly?

Besides that, this is looking good!

EricPickup:
@d12 how's this?
image

d12:
:+1:

EricPickup:
@d12 sorry, I forgot to push the change 😆 should be good to go now!

Issue #2190, opened 2019-08-07
Error Linking Student to their Github Account


dicer2000:
**Describe the bug**
When I try to link a student to their Github account, I get an error.

**To Reproduce**
1. Go to Settings => Roster Management
2. Click on Link to GitHub Account
3. "Select an unlinked GitHub account to link this student to"

**Expected behavior**
Expected to enter a GitHub account for this student.  Am I missing something?

**Screenshots**
![Screen Shot 2019-08-07 at 11 11 33](https://user-images.githubusercontent.com/3097749/62639157-8b532300-b904-11e9-9fde-92f76c2ea781.png)

EricPickup:
Hi @dicer2000 👋 
This is most likely because no GitHub users have accepted the assignment or all the users have already been linked to a roster entry. When your students accept the assignment they will be able to select their identifier (looks like you use names) in the roster list to link their GitHub account to the roster entry.

![image](https://user-images.githubusercontent.com/1490434/62641925-8046c580-b8f9-11e9-8e03-929b8a841d05.png)

Once they've selected their identifier that GitHub account will be associated with that roster entry across your whole classroom. The page you're on is used to modify those links (e.g. a student selected the wrong identifier). You can unlink a user from an identifier and link them to another.

I can see how this experience can be improved. I'll add a message indicating why that list is empty and create an issue to put up some help docs for roster management.

PR #2193, opened 2019-08-07
Add message indicating empty unlinked user list

EricPickup:
## Problem

Related to #2190 

We have no message indicating when/why the unlinked github user list is empty.

## Solution

Add a message!

**Before**
image

**After**
image


d12:
Can we go one step further and remove the "Select an unlinked..." prompt in this case? It doesn't make sense if there isn't anything to select.

EricPickup:
@d12 done!
image

stephaniegiang:
Nit pick: The sentence is kind of hard to read and is long. Maybe can do this:

```
There are no GitHub accounts to link. GitHub accounts will appear when a student:
- Has accepted an assignment
- Is not linked to a roster identifier
```

Doesn't have to be exactly that. Just a suggestion.

EricPickup:
@stephaniegiang I think that separating it in the list format implies an "either or" with those two points (i.e. they will appear if they have accepted an assignment in the classroom _and/or_ also if they're not linked). For now I'll deploy this but if we see any more confusion we can adjust it later 👍

PR #2194, opened 2019-08-07
Hide archived classrooms into a sorting option

EricPickup:
## What
This and #2189 will finish the classroom archiving feature (#602).

This PR moves archived classrooms out of the dashboard into its own `Archived` section.

R7RkrXebON


d12:
Filtering or showing archived repos isn't a "sort", I feel like this is abusing the sort functionality a bit, and it's not where people are going to look when trying to find their archived classes. Our designers may be able to lend some advice here.

PR #2198, opened 2019-08-07
Skip user cache in a few places where we make API calls

d12:
We were using cached GitHub logins when making a few API requests which causes errors when users change their handle. All fixed!


app/controllers/organizations_controller.rb
```diff
@@ -123,7 +123,7 @@ def search
   def authorize_organization_addition
     new_github_organization = github_organization_from_params
 
-    return if new_github_organization.admin?(current_user.github_user.login)
+    return if new_github_organization.admin?(current_user.github_user.login(use_cache: false))
```
jeffrafter: What would you think about extracting the login to a `current_user_login` method? I'm mostly wondering about memoizing this so we aren't busting the cache multiple times...

d12: @jeffrafter Sorry for the late :eyes:, could you elaborate here? The login will be memoized to `current_user.github_login` in this case.

jeffrafter: Sorry, I don't see where that is happening and might be confused. Looking at the code it looked like there were multiple code-paths in this controller that led to: `current_user.github_user.login(use_cache: false)`. This is the before filter, but lower it is hit in `create_user_organization_access` and that method is being called N times (where N is the number of classrooms here: https://github.com/education/classroom/blob/049a9c0ef89a5ae42f9bfb94a1c9b79bd1704edb/app/controllers/organizations_controller.rb#L173 (I think 🤐). My main concern here is that by adding `use_cache: false` we will be busting the cache every time the `login` is accessed in this controller which could be multiple. If that is true this controller would be using up to date information but would dramatically increase the potential number of API calls. 

That said; if this is being memoized somewhere I am missing then all of this is 💯

app/controllers/concerns/organization_authorization.rb
```diff
@@ -9,7 +9,12 @@ module OrganizationAuthorization
 
   def authorize_organization_access
     return if @organization.users.include?(current_user)
-    github_organization.admin?(current_user.github_user.login) ? @organization.users << current_user : not_found
+
+    if github_organization.admin?(current_user.github_user.login(use_cache: false))
```
jeffrafter: In looking more deeply at this... the method is called as part of the `before_action`: https://github.com/education/classroom/blob/ecc1747835db3cb6d0cb62801550e9fa16424624/app/controllers/concerns/organization_authorization.rb#L7. That means anywhere we include `OrganizationAuthorization` we are potentially introducing a new API call (if the user is not included). This is actually 9 controllers (most of the surface area of assignments and organizations and more). I'm not sure how to understand the impact of that. Are users _usually_ already part of the org (meaning the guard clause above will help us?)

app/controllers/organizations_controller.rb
jeffrafter: This is fixed in https://github.com/education/classroom/pull/2198/commits/e76b055e347d66bbc479eeb43e72137099c2bdbf

PR #2200, opened 2019-08-07
Fourth Iteration: Part 2: Remove GroupAssignmentRepo::Creator

shaunakpp:
## What
Deletes the old `GroupAssignmentRepo::Creator`.


shaunakpp:
This is the last of the deletion pull requests 💣

PR #2203, opened 2019-08-07
Add sakai support

thebradbain:
## What
Due to now supporting LTI 1.0 in addition to LTI 1.1 (PR #2179) and the refactoring to be more modular (PR #2174), we can now _very_ easily add support for new LMSs which adhere to the LTI 1.1/1.0 standard. 

For example, with very little changes we now support Sakai:
Screen Shot 2019-08-07 at 3 02 50 PM

sakai

On a separate note: this is personally exciting for me because my school uses Sakai as its LMS, and there's conversations around updating our CS curriculum tooling department-wide, so... who knows, maybe it's a match made in heaven 😍


app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -52,11 +52,12 @@ def lms_membership
...
         membership_service_url,
         current_organization.lti_configuration.consumer_key,
-        current_organization.lti_configuration.shared_secret
+        current_organization.lti_configuration.shared_secret,
+        lti_version: current_organization.lti_configuration.lti_version
       )
 
       begin
-        membership_service.students
+        membership_service.students(body_params: current_organization.lti_configuration.context_membership_body_params)
```
thebradbain: The changes to this file should have actually been made in #2174, but it's fine they weren't because they have default fallbacks. Better late than never I suppose 😅

PR #2205, opened 2019-08-08
Google Classroom / LTI better error messages

thebradbain:
## What
In preparation for the release based upon the user testing insights, now is as good a time as any to go through and clean up relevant instructions / provide more actionable errors when something goes awry. 

In addition to providing user-facing errors whenever an LTI Launch goes wrong (e.g. a user enters the wrong consumer key), refactors the LTI-specific code for validating LMS launches on `SessionsController` into `sessions_controller/lti_dependency.rb`, which allows us to break up the `lti_launch` method into easily digestible components. 

Screen Shot 2019-08-07 at 10 46 48 PM

_E.g. When there's an error during the LMS-to-GitHub Classroom handoff, notify them in the LMS_


thebradbain:
As an aside -- this is the last PR I plan to deploy re: LMS integration before launch (barring any small copy tweaks etc) so now I'll be free to help tidy up any other loose ends in preparation for next week.

spec/controllers/orgs/google_classroom_configurations_controller_spec.rb
```diff
@@ -53,8 +53,7 @@
           end
 
           it "flashes error message" do
-            message = "A LMS configuration already exists. Please remove configuration before creating a new one."
-            expect(flash[:alert]).to eq(message)
+            expect(flash[:alert]).to be_present
```
d12: In these tests, why aren't we able to assert exact messages anymore? This loses the assertion that we're displaying the correct error.

thebradbain: Oh I can add those back! I didn’t think about it that way, but that makes sense.

PR #2207, opened 2019-08-08
Swallow NotFound errors when creating template repo

EricPickup:
## Problem
After much investigation we've found the the notorious `NotFound` template repo error is due to organizations restricting third-party access to their repositories.

Since this is expected behavior in this case we should swallow the error and not report it to Haystack of Datadog as a failure.

## Solution
Report a different error message explaining the issue. Report it under its own stat in datadog to prevent it from affecting our success/failure rate.

## Note
Once I find the most effective way for teachers to fix their settings I will add a section to the help docs on how to fix the issue.


app/services/create_github_repo_service.rb
```diff
@@ -239,16 +228,37 @@ def repo_from_template_options
     }
   end
 
+  # rubocop:disable MethodLength
+  def handle_template_error(error, options)
+    raise Result::Error.new errors(:template_repository_not_found), error.message if error.class == GitHub::NotFound
```
jeffrafter: Not a blocker, but I wonder if it would be more clear to add this as a specific matching rescue at https://github.com/education/classroom/pull/2207/files#diff-63d3a0a1bbef5e8b4071f230c0a305aaR85. In there we could re-throw the Result::Error as we are doing and it would be a little closer to the concern. 🤷‍♂

EricPickup: Makes sense! 👍

PR #2209, opened 2019-08-08
Tie up LTI loose ends

thebradbain:
## What
Two very small tweaks I made after giving everything a second look:

1. Ensure our `LtiLaunchError` handler will never re-raise an error on its own. We were previously implicitly assuming that the Launch Message passed to LtiLaunchError would always be a well-formed instance of `IMS::LTI::Message`. While this is true on our end, I just don't like having assumptions in error handlers (especially because this is an error handler which covers invalid LTI Launches from an LMS, so assuming the message is well-formed is maybe stretching it).

2. UI tweak --> a lone "cancel" button in a modal should have been styled as a link.


app/controllers/sessions_controller/lti_dependency.rb
```diff
@@ -94,7 +94,7 @@ def handle_lti_launch_error(err)
     error_msg = err.message
     message = err.lti_message
 
-    if message && message.launch_presentation_return_url
+    if message.try(:launch_presentation_return_url)
```
jeffrafter: Non blocking: does this do the same?

```suggestion
    if message&.launch_presentation_return_url
```

thebradbain: I wish it did, but within the `IMS::LTI` gem our LTI code is built on top of, there’s a few different subclasses of `Message`, though for some reason they’ve decided to only let valid subclasses respond to `launch_presentation_return_url`, so in the case an LTI Launch Message is not well formed and it’s treated as a generic `Message` instance  it actually won’t respond to `launch_presentation_return_url`.

PR #2211, opened 2019-08-08
Add dependent destroy to roster entries

stephaniegiang:
## What
Currently, we do not delete roster entries when a roster is deleted. So this PR fixes that.

I don't think us holding onto roster entries that are associated to deleted rosters is valuable to us.


thebradbain:
🔍 good detective work!

stephaniegiang:
Going to wait for @d12's input before merging 😄

PR #1835, opened 2019-05-17
[WIP] Roster Management and Classroom Assignment Organization UI Updates

srinjoym:
New classroom and assignment UI layouts!! ✨

PR #2223, opened 2019-08-09
Assignment invitation URL improvements

### LINKED ISSUE #2115: Suggested improvement for Enable assignment URL functionality
### Problem

Currently, the functionality for enabling the assignment URL lacks context for what it does. Since it's located within the assignment creation flow, it's divorced from the actual URL itself. Conversely, once the assignment has been created and the user is shown the assignment URL, we don't provide any feedback to indicate whether the URL is enabled or not. 

### Proposal

Relocate the functionality _after_ the assignment creation and combine it with the URL. Simplify the URL invitation display to reduce the visual weight and help draw more attention to the URL enabling checkbox. By grouping the URL enabling functionality with the URL itself, not only do we make it clearer _what_ is being enabled but also what the enablement state is.

**Mockup**

Figma: https://www.figma.com/file/97NA4vopFIccJHuzVVSO8V/Education?node-id=25%3A10

1 4 enable-URL

**Interaction**

The interaction pattern for the URL copying is similar to what we currently use in PRs to indicate the PR branch (for example: https://github.com/github/github/pull/120069)

url-copy

EricPickup:
Closes #2115.

This PR makes the assignment invitation URL stand out more on the assignment view page, and adds the ability to enable/disable the invitation URL through ajax calls directly on the page.

6IImxsJjGk


app/views/shared/_invitation.html.erb
```diff
@@ -1,9 +1,16 @@
-<form>
-  <% if mobile? %>
-    <%= render partial: "shared/invitations/mobile", locals: local_assigns %>
-  <% else %>
-    <%# if for some reason the user has a really tiny desktop screen fallback to the mobile view" %>
-    <%= render partial: "shared/invitations/mobile", locals: local_assigns %>
-    <%= render partial: "shared/invitations/desktop", locals: local_assigns %>
-  <% end %>
-</form>
+<dt class=<%= mobile? ? "mobile-invitation-content" : "desktop-invitation-content" %>>
```
femmebot: Was this page supposed to be responsive? It doesn't appear to be the case for the current site's layout.

app/views/shared/_invitation.html.erb
```diff
@@ -1,9 +1,16 @@
...
+      url: url_for(action: :toggle_invitations, id: (@assignment.nil? ? @group_assignment.slug : @assignment.slug)),
+      method: "POST"
+    } %>
+  <label>Enable assignment invitation URL</label>
+  <span style="color: green"><%= octicon 'check', height: 16, id: "checkmark", style: "display: none; padding-bottom: 3px;" %></span>
+  <%= image_tag 'octocat-spinner-128.gif', width: 14, height: 14, class: 'spinner', id: "loading-indicator", style: "display:none" %>
+  <div class="input-group input-block">
+    <%= render partial: "shared/invitations/clipboard_form", locals: local_assigns %>
+  </div>
+  <br>
```
femmebot: Remove the <br> to make it more visually connected with the URL component.

app/views/shared/_invitation.html.erb
```diff
@@ -1,9 +1,16 @@
...
+    onchange: "$(this).data('params', 'invitations_enabled=' + this.checked);showLoadingIndicator(true);showCheckmark(false);",
+    data: {
+      remote: true,
+      url: url_for(action: :toggle_invitations, id: (@assignment.nil? ? @group_assignment.slug : @assignment.slug)),
+      method: "POST"
+    } %>
+  <label>Enable assignment invitation URL</label>
+  <span style="color: green"><%= octicon 'check', height: 16, id: "checkmark", style: "display: none; padding-bottom: 3px;" %></span>
+  <%= image_tag 'octocat-spinner-128.gif', width: 14, height: 14, class: 'spinner', id: "loading-indicator", style: "display:none" %>
+  <div class="input-group input-block">
```
femmebot: Add `.py-2` to add 8px above and below the `.input-group.input-block` div so the text isn't sitting too close to the input.

app/views/shared/_invitation.html.erb
```diff
@@ -1,9 +1,16 @@
...
-  <% if mobile? %>
-    <%= render partial: "shared/invitations/mobile", locals: local_assigns %>
-  <% else %>
-    <%# if for some reason the user has a really tiny desktop screen fallback to the mobile view" %>
-    <%= render partial: "shared/invitations/mobile", locals: local_assigns %>
-    <%= render partial: "shared/invitations/desktop", locals: local_assigns %>
-  <% end %>
-</form>
+<dt class=<%= mobile? ? "mobile-invitation-content" : "desktop-invitation-content" %>>
+  <%= check_box_tag 'invitations_enabled',true, (@assignment.nil? ? @group_assignment.invitations_enabled : @assignment.invitations_enabled),
```
femmebot: Since we moved the checkbox to enable the URL here, do we still need the checkbox in the assignment creation flow?

d12:
Curious what the reasoning is behind adding so much whitespace around the invitation link. It's important, but a teacher typically copies it once and after that, they only care about the list of submissions. This work pushes the useful information _way_ down the page. Is there a better way we can design this? Possibly making better use of horizontal space that we aren't using right now?

d12:
Gotcha :) I 100% agree that the checkbox lives on this page, that makes sense.

To show why I think the vertical space could be an issue, I pulled down the branch and un-fullscreened my Chrome (I don't know how many people use fullscreen apps vs non fullscreen, but I've seen both)

Screen Shot 2019-08-12 at 11 32 07 AM

Notice how headers, more headers, and the invitation link dominate the screen. This screenshot is also representative of users with resolutions below 1920*1080, or users who increase the font size in their browser.

With a bit of re-organization, I wonder if the invitation link could go within the assignment header where the download + settings buttons are.

I don't want to block any work, just wanted to give my two cents :)

stephaniegiang:
@femmebot @d12 I have a PR in the works that will make the changes that Brandon proposed in https://github.com/education/classroom/pull/1835 so this area will be moved to the header. Maybe that will help with the poor use of space.

app/views/shared/_invitation.html.erb
EricPickup: It looks like we have some flaws with this. The mobile/desktop class is decided on the page load (as seen on this line) so switching on inspect element with mobile will not work instantly and requires refresh. We could fix this in another issue using media queries.

This is what it looks like for me on mobile:
image

app/views/shared/_invitation.html.erb
EricPickup: @femmebot do you mean to remove it only from creation or from the assignment settings page as well?
@d12 what are your opinions on this? I think it should remain in the settings page since it is still a setting. Would be weird keeping one specific setting outside of the settings page.

EricPickup:
image

How does this spacing look? I removed all of the extra padding that was originally added. Also fixed the issue where the spinning octocat gif bumped the block up/down.

app/views/shared/_invitation.html.erb
d12: I'll defer to @femmebot or the engineering team here to make the correct decision :)

d12:
This spacing looks much better, thanks!

EricPickup:
@femmebot let me know what you think 😄

app/views/shared/_invitation.html.erb
femmebot: Let's keep it for the time being.

app/assets/stylesheets/layout.scss
```diff
@@ -432,6 +429,11 @@ hr {
   }
 }
 
+.desktop-invitation-content {
```
JessRudder: Open question (that I do not know the answer to) - Do we need the custom CSS or could this be achieved by conditionally adding/removing Primer classes?

EricPickup: We do have a primer class for auto margin but I can't find anything for 50% width so we'd still have to add this class 😞I'll leave it as-if for now since we need the class either way and it's easier to distinguish the difference between mobile vs desktop in separate classes

Issue #2037, opened 2019-07-18
ActionView::Template::Error


stephaniegiang:
ActionView::Template::Error

**[ActionView::Template::Error](https://haystack.githubapp.com/classroom/rollups/04af95995f467d51f44781c46c6459b7)** &mdash; First occurred on 2019-04-20 09:50:06 -0700

```
undefined method `github_repository' for nil:NilClass
```

*Brought to you by Haystack's `!` hotkey.  Try it on a needle or rollup today!*

PR #2231, opened 2019-08-12
Add 406 Error Handling for Requests for Non-Specified Formats

JessRudder:
Closes https://github.com/education/classroom/issues/2037

Most of our routes use implicit rendering and/or do not specify the format types that are handled by that controller end point.

When people try to specify an unsupported format, an `ActionView::Template::Error` is generated.

This PR provides users with a better experience by giving them an informative error page. On the back end, we'll be reducing error noise while still tracking the occurrence of this issue using statsd.

This will not change any behavior that is already specified in `format` blocks.

image


d12:
Nice 🤘 

Could you add a test for this? Something in [here](https://github.com/education/classroom/blob/ad7e73dca646733253135f47e826c81c1c13bada/spec/controllers/application_controller_spec.rb) would do.

spec/controllers/application_controller_spec.rb
```diff
@@ -25,7 +31,7 @@ def index
 
     context "user access token is not authorized" do
       before do
-        User.any_instance.stub(:authorized_access_token?).and_return(false)
+        allow_any_instance_of(User).to receive(:authorized_access_token?).and_return(false)
```
JessRudder: This doesn't change any behavior but gets rid of the deprecation warning for the old `any_instance` syntax.

PR #2232, opened 2019-08-12
Validate Grouping while validating GroupAssignment

### LINKED ISSUE #2230: 500 Internal server error when creating a Grouping with a duplicate name
This came up in #2227 

**Describe the bug**
We get a 500 internal server error when creating a new Grouping with the same name as an existing Grouping on the org.

**To Reproduce**
1. Go to new Group Assignment form
2. Under ` Create a new set of teams` enter a name of a grouping that already exists (check your dropdown)
3. Submit

**Expected behavior**
We should handle this error properly.

EricPickup:
Fixes #2230.

## Problem

`Grouping`s are persisted when their associating `GroupAssignment` is saved, however, if the grouping fails its validation the group assignment will still save with a nil grouping_id.

## Solution

Use the `validates_associated` helper on `GroupAssignment` to validate the grouping as part of the assignment's validation. I've also added a couple methods to fetch the errors straight from the grouping object since `validates_associated` only adds a generic `Your grouping is invalid` error message on the group assignment object.


app/models/group_assignment.rb
```diff
@@ -42,6 +42,8 @@ class GroupAssignment < ApplicationRecord
   validate :starter_code_repository_not_empty, if: :will_save_change_to_starter_code_repo_id?
   validate :starter_code_repository_is_template, if: :will_save_change_to_starter_code_repo_id?
 
+  validates_associated :grouping
```
d12: Can we add a test for this?

PR #2235, opened 2019-08-12
[Hotfix!] Avoid creating duplicate roster entries after creation

thebradbain:
## What
It appears that #2181 introduced the following (subtle) bug: 

1. Create a new roster imported from an LMS with at least 2 students. This should succeed as expected.
2. Sync from the LMS. You should be told there are no students created as roster is up to date. This is also expected.
3. Delete one of the roster entries.
4. Sync your roster again. You should now have the deleted roster entry re-added to your roster. This is expected and correct.
5. Sync your roster again. You should now have a duplicate entry of that roster-entry. This is a bug, as it should have informed the user no students were created as the roster is up to date 🐛 

The fix is straightforward: it appears that when updating the `import_from_lms.html.erb` `RostersController` view with a hidden field, that corresponding field was not added to `_select_identifier_modal.html.erb` which is used in syncing an existing roster. 

In the future, we should unify these two views to use the same form. For now, though, with launch imminent, we should probably merge this asap.


d12:
LGTM! This is good to go with a regression test

PR #2237, opened 2019-08-13
Improves .order_for_view query on RosterEntry

shaunakpp:
## What
Potentially fixes: https://github.com/education/classroom/issues/1969#issue-465872705

Improves the `.order_for_view` query on `RosterEntry` by replacing the `IN` clause with a `LEFT JOIN`. Although, theoretically, joins are faster than large `IN` queries, In my local setup I am seeing marginal improvements. But there _is_ an improvement. 

~**TODO:**~
~Run experiments on a few classrooms to check whether it actually improves the performance by a significant margin.~ I believe the performance would significantly improve for a classroom with large number of linked and accepted students. ~But because I don't have enough evidence, I'm keeping this a draft pull request.~

*Update:*

I ran the query for a large enough assignment, and here's the improvement:
Old code:
~~~ruby
Benchmark.ms { roster.roster_entries.order_for_view(a) }
=> 4.379558842629194 # 4.3s without pagination
~~~

New code:
~~~ruby
 Benchmark.ms { roster.roster_entries.order_for_view_new(a) }
=> 0.5706739611923695 #0.5s without pagination
~~~~


thebradbain:
SQL king 👑

app/models/roster_entry.rb
```diff
@@ -69,18 +69,25 @@ def self.to_csv(user_to_group_map = {})
...
+    SQL
 
-    order <<~SQL
+    order_sql = <<~SQL
       CASE
-        WHEN roster_entries.user_id IS NULL THEN 2                   /* Not linked */
-        WHEN roster_entries.user_id IN #{sql_formatted_users} THEN 0 /* Accepted */
-        ELSE 1                                                       /* Linked but not accepted */
+        WHEN roster_entries.user_id IS NULL THEN 2      /* Not linked */
+        WHEN roster_entries.user_id IS NOT NULL THEN 0  /* Linked but not accepted */
```
jeffrafter: I wonder if there is a spec covering this behavior. I think we may need a small change:

```suggestion
        WHEN roster_entries.user_id IS NOT NULL AND assignment_repos.user_id IS NOT NULL THEN 0  /* Linked but not accepted */
```

Essentially I think this method is attempting to find the three set:

- There is no roster user id
- There is a roster user id and they have a repo
- Everyone else (there is a roster user id and they do not have a repo)

In this case I think the `LEFT JOIN` is a fantastic solution... but we need to actually verify here whether or not the `LEFT JOIN` found a result or not.

shaunakpp: Thanks for catching that! before this, we only sorted roster_entries into two buckets instead of intended three!

PR #2240, opened 2019-08-13
Add another LMS statsD

stephaniegiang:
## What
Adds datadog metrics for:
- successful roster imports
- number of students that are being imported

Also added tests for `Roster#create` for LMS since that was missed. 

Also did a bit of refactoring and pulling the statsD into a different method because I was getting Rubocop errors about complexity. So instead of ignoring, I moved them out into a different method.


stephaniegiang:
@education/classroom-reviewers Ready for review!

thebradbain:
✨ beautiful

PR #2244, opened 2019-08-13
Fix auth failure redirect

### LINKED ISSUE #2239: Moodle error message is unfriendly
**Describe the bug**
On Moodle LMS integrations, whenever a user has entered the wrong `consumer_key` or `shared_secret`, authentication will fail (as expected). However, when we redirect the user back to their LMS with a friendly error message explaining they've entered the wrong credentials, they're instead greeted with a different error: "sesskey invalid"

This bug only occurs for Moodle, and not others I've tested, such as Sakai and Canvas.

thebradbain:
## What
Closes #2239. Adds more tests around LTI Auth failures and handling redirections. Now Moodle users are presented with an actionable error message on their LMS if they've misconfigured GitHub Classroom, as is consistent with the behavior of other LMSs. 

Essentially this bug was being caused by the method we were passing error context around on authorization failure: the default [OmniAuth failure handler](https://github.com/omniauth/omniauth/blob/master/lib/omniauth/failure_endpoint.rb) swallows the request which failed's parameters, and we previously tried to workaround it by providing the lti request parameter's `launch_presentation_return_url` as the message type passed to authorization failure callback (probably an abuse of functionality), which caused error handler itself to fail when the url had a query string because it couldn't be converted to a symbol.

Screen Shot 2019-08-13 at 3 05 15 PM


lib/omniauth/authorization_failure_endpoint.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module OmniAuth
+  # rubocop:disable AbcSize
+  class AuthorizationFailureEndpoint < OmniAuth::FailureEndpoint
```
thebradbain: Here we override the default OmniAuth failure endpoint in order to pass through the failing request parameters to our exception handler, as the default behavior is to simply swallow the request and redirect with no other context, which lead to a workaround of passing a fully qualified `error_url` in the `message` parameter which breaks when the url contains a querystring

PR #2245, opened 2019-08-14
Clean up error dialog box removal

### LINKED ISSUE #1988: Error box covers other form fields and can't be closed
The error dialogue box on the assignment forms covers the next field or descriptions and cannot be closed. This is annoying especially since the descriptions under fields may help explain the confusion that caused the error in the first place.

image
image
image

JessRudder:
Closes #1988 

Expands the existing error message removal so it works for all of the fields and not just the template import field.

error-handling


app/assets/javascripts/assignment_form.js
```diff
@@ -7,12 +7,13 @@ function importOptions(starterCodeFieldValue) {
...
-function removeErrorBox() {
-  errorBoxes = document.getElementsByClassName("error");
-  for (var i = 0; i < errorBoxes.length; i++) {
-    errorBox = errorBoxes[i]
-    if (errorBox.innerText == "Starter code repository is not a template repository. Make it a template repository to use template cloning.") {
-      errorBox.style.display = 'none';
-    }
-  }
+function removeErrorBox(el) {
+  parent = el.closest(".errored");
```
jeffrafter: I think we need a keyword in front of this: `const`, `let` or `var`. The code you're replacing didn't have them, but we should start adding them in to eliminate unneeded globals:

```suggestion
  var parent = el.closest(".errored");
```

app/assets/javascripts/assignment_form.js
```diff
@@ -7,12 +7,13 @@ function importOptions(starterCodeFieldValue) {
...
+
+  if (!parent) return;
+
+  errorBoxes = parent.getElementsByClassName("error");
```
jeffrafter: Similar to above we should add `const`, `let` or `var`. Also, what if we started leaning toward `querySelector` so that we can start moving (slowly) to more modern and consistent finders? If we do that we can select a single `errorBox`, we're only expecting one:

```suggestion
  var errorBox = parent.querySelector(".error");
```

app/assets/javascripts/assignment_form.js
```diff
@@ -7,12 +7,13 @@ function importOptions(starterCodeFieldValue) {
...
+
+  parent.classList.remove("errored");
+  errorBoxes.length && errorBoxes[0].remove();
```
jeffrafter: If the above change is applied this could be simplified to:

```suggestion
  errorBox && errorBox.remove();
```

PR #2246, opened 2019-08-14
Handle paginated membership service responses

thebradbain:
## What
While double checking LTI specs, I noticed that in [section 3.3.3 of the LTI 1.1 Names and Role Provisioning Service](https://www.imsglobal.org/specs/ltimemv1p0/specification-3), there’s a small aside about paginated responses:
```
If the response from a Tool Consumer does not comprise all of the members, a 
nextPage element will be included in the JSON response with a URL to request the next 
set of members. The absence of a nextPage element indicates that no more members 
are available.
```
It’s optional for an LMS to implement and up for the LMS to decide what the cutoff is, but GitHub Classroom should be ready to handle paginated rosters in the case an LMS does return membership in that way.


andrewbredow:
Is there a stubbed out readable representation of the response format somewhere?

`#lazy-manager`

thebradbain:
Yeah, definitely. Within the specification, the example JSON response is exactly the same as usual but with a `NextPage` key: 

```
{
  "@context" : [
    "http://purl.imsglobal.org/ctx/lis/v2/MembershipContainer",
    {
      "liss" : "http://purl.imsglobal.org/vocab/lis/v2/status#",
      "lism" : "http://purl.imsglobal.org/vocab/lis/v2/membership#"
    }
  ],
  "@type" : "Page",
  "@id" : "http://lms.example.com/sections/2923/memberships/?rlid=49566-rkk96",
  "nextPage" : "http://lms.example.com/sections/2923/memberships/?rlid=49566-rkk96&p=2",
  "differences" : "http://lms.example.com/sections/2923/memberships/?x=1422554502",
  "pageOf" : {
    "@type" : "LISMembershipContainer",
    "membershipSubject" : {
      "@type" : "Context",
      "contextId" : "2923-abc",
      "membership" : [
        {
          "status" : "liss:Active",
          "member" : {
            "@type" : "LISPerson",
            "sourcedId" : "school.edu:user",
            "userId" : "0ae836b9-7fc9-4060-006f-27b2066ac545",
            "email" : "user@school.edu",
            "familyName" : "Public",
            "name" : "Jane Q. Public",
            "image" : "http://...",
            "givenName" : "Jane"
          },
          "message" : [
            {
              "message_type" : "basic-lti-launch-request",
              "lis_result_sourcedid" : "83873872987329873264783687634",
              "ext" : {
                "user_username" : "jpublic"
              },
              "custom" : {
                "country" : "Canada",
                "user_mobile" : "123-456-7890"
              }
            }
          ],
          "role" : [
            "lism:Instructor"
          ]
        }
      ]
    }
  }
}
```

I can stub out the whole response if we like in the tests, if that's desirable. My thought-process was mostly that the pagination tests should only test that we make another membership request for the `nextPage` key if it's available, and leave the actual request/response/parsing validation to the already existing membership tests.

lib/github_classroom/lti/membership_service.rb
```diff
@@ -20,25 +20,42 @@ def instructors(body_params: nil)
...
-        req = membership_request(roles, body_params)
-        response = send_request(req)
+      def membership(membership_url: @context_membership_url, roles: [], body_params: nil)
+        total_membership = []
+
+        body = fetch_raw_membership(membership_url, roles, body_params)
+        total_membership += parse_membership(body)
+
+        # pagination
+        if @lti_version == 1.1
```
stephaniegiang: Does the pagination only exist for 1.1?

thebradbain: Yup! Version 1.0 doesn't support that capability.

EricPickup: Not sure if we need to check the version since we're still handling the situation where `nextPage` doesn't exist.

stephaniegiang:
@EricPickup The next page only exists for version 1.1. So we don't want to make that same call in the other version has that will cause a break in the code.

EricPickup:
@stephaniegiang We're already checking if the response contains a `nextPage` element though:
https://github.com/education/classroom/blob/64135f505ec46666ec7aa32f9cd77e044a925897/lib/github_classroom/lti/membership_service.rb#L32
So even if it's an older version, `next_page` would be nil and we would skip it

stephaniegiang:
@EricPickup Ahhh true. I think I rather keep it in. It makes the code more readable. It gives other developers more context on that this is only in version 1.1 making debugging easier. Is this ok with you?

Issue #2241, opened 2019-08-13
Fork all starter code repositories onto classroom organization


EricPickup:
Related to #2222

When teachers create an assignment and wants to use a starter code repository, we should fork the starter code repo onto the classroom's organization and use that as the actual starter code repository.

Advantages:
- We will have full permissions to change repo settings (e.g. we can make it a template)
- All of the classroom owners/admins will be have the same access to starter code repositories (fixes #2222)
- We can automatically make the repos into a template, eliminating a step for teachers to use template repos
- Dodges the issue where we can't use template repos belonging to different organizations (#2234)

Important notes:
- We can only fork a repository on an organization. If teachers re-use a repo for multiple assignments they will have to point to the same forked repo
- The only case we shouldn't fork the repo is if the starter code repo already exists on the classroom organization
- Since we can only fork a repo once, we need to consider how we update the fork. For example, if we use `education/classroom` as the starter code repo for a Fall 2018 course. Fall 2019 comes around and we want to do a similar assignment using `education/classroom`. Teacher uses that as starter code repo, they'll want the most recent version of classroom (forked version on the account will still be from 2018)

UX/UI concerns:
- We need to make sure that teachers know we're forking this repo onto their organization and why
- Teachers need to know that if they want to update the starter code repo, they should update the **fork** and not the original repo

cc @femmebot @sophshep for UX decisions
cc @d12 @shaunakpp for back-end stuff

PR #2250, opened 2019-08-14
Added Assignment::Creator class

EricPickup:
## What
This PR adds a `Creator` class for `Assignment` so that we can isolate everything related to assignment creation outside of the controller. We'll be adding a lot more logic to assignment creation with #2241.

We already have an `Assignment::Editor` class and this is similar to `Organization::Creator`, `Group::Creator`, and `Roster::Creator`.

## Notes
- This is under a feature flag and will be gradually rolled out
- Old logic/tests inside the controller will be removed once we're 100% with this Creator
- I plan to merge group assignment creation into this creator, unifying `GroupAssignment` and `Assignment` creation under one creator class

### DIFF
--- a/app/models/assignment/creator.rb
+++ b/app/models/assignment/creator.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+class Assignment
+  class Creator
+    attr_reader :assignment, :options
+
+    class Result
+      class Error < StandardError; end
+
+      def self.success(assignment)
+        new(:success, assignment: assignment)
+      end
+
+      def self.failed(assignment, error)
+        new(:failed, assignment: assignment, error: error)
+      end
+
+      attr_reader :assignment, :error
+
+      def initialize(status, assignment: nil, error: nil)
+        @status     = status
+        @assignment = assignment
+        @error      = error
+      end
+
+      def success?
+        @status == :success
+      end
+
+      def failed?
+        @status == :failed
+      end
+    end
+
+    # Public: Create an Assignment.
+    #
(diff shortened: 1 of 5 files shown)


spec/models/assignment/creator_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+
+      it "returns result containing unsaved assignment" do
+        result = subject.perform(options: options)
+        expect(result.assignment.persisted?).to be false
+      end
+    end
+  end
+
+  context "deadlines" do
```
EricPickup: All of these deadline tests are the exact same tests [taken from `AssignmentsController`](https://github.com/education/classroom/blob/master/spec/controllers/assignments_controller_spec.rb#L133). The old tests in the controller will be removed once we get rid of the feature flag.

thebradbain:
I don't have too much experience with the `Creator` pattern, but is there a plan to also create a `GroupAssignment::Creator` class? Or will #2241 not concern `GroupAssignmentRepo` and `GroupAssignment`?

EricPickup:
@thebradbain I'm hoping to merge GroupAssignment creation into this same creator after this PR goes in. `GroupAssignment` already shares `Assignment::Creator` for editing assignments

app/models/assignment/creator.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+class Assignment
+  class Creator
+    attr_reader :assignment, :options
+
+    class Result
+      class Error < StandardError; end
```
jeffrafter: I don't see this being used...

Issue #2222, opened 2019-08-09
Teachers using private repo belonging to their user account


EricPickup:
A lot of teachers have been using private repos created on their **user** account (rather than on the org) as their starter code repo.

This isn't an issue when they create the assignment because we're querying that repo on their behalf. The repo is accessible.

However, when a student accepts that assignment, we pick a random org owner's token to create a repo on the classroom org. If that random token is not the teacher who owns the starter repo, the repo is inaccessible.

I'm guessing we never noticed this because most orgs/classrooms will only have one owner (aka the teacher that owns the private repo).

This is the _actual reason_ templates were failing for the past few weeks.

d12:
100/100 investigative work, nice work 😄 

@EricPickup @shaunakpp I'd love for you to try to solve this in your last week - if possible. If you don't have time to implement anything, outlining a proposed fix for handoff to @education/classroom-reviewers to work on would be :100:

EricPickup:
@shaunakpp let's set up some time today to work on this!

A couple solutions I've thought about..

1. When a new assignment is created or a new teacher/TA is added as a classroom owner, add them as collaborators to any starter code repositories. We'd definitely add some message to indicate this. We could check who created each assignment, grab their token and add all other classroom owners. This would help prevent reaching rate limits in our largest classrooms.

2. When creating a repo for the student, iterate through (or choose at random) the list of classroom owners and send a GET for the starter code repo on their behalf. If we get a 404, skip to the next owner. Once we get a successful response use their token. This is the safest way to do it and requires no unexpected changes to the teacher's repo. However, this does leave a higher chance of hitting the rate limits on large classrooms.

@d12 @shaunakpp let me know what you guys think.

shaunakpp:
> When a new assignment is created or a new teacher/TA is added as a classroom owner, add them as collaborators to any starter code repositories.

This may not work if the starter code repository is a public repository not belonging to the teacher or the organization.

> When creating a repo for the student, iterate through (or choose at random) the list of classroom owners and send a GET for the starter code repo on their behalf. If we get a 404, skip to the next owner. Once we get a successful response use their token.

This makes an already long-running job longer, is it possible to short-circuit this step while assignment creation? Like, keep a list of user_ids that are valid owners/collaborators for the starter code repository and use one of them while creating the assignment repo for a student?

EricPickup:
> This may not work if the starter code repository is a public repository not belonging to the teacher or the organization.

[...]

PR #2265, opened 2019-08-19
Use assignment creator's token to create repo in certain cases

EricPickup:
## What
Back-end solution to #2222 

If the starter code repository is private and does not belong to the classroom organization, use the assignment creator's token to create the student repo.

In all other cases, continue our current method of grabbing a random teachers token to create the students repositories.


spec/services/create_github_repo_service_spec.rb
```diff
@@ -48,8 +48,7 @@
         options = {
           private: false,
           is_template: true,
-          auto_init: true,
-          accept: "application/vnd.github.baptiste-preview"
+          auto_init: true
```
EricPickup: These changes are unrelated to the PR, just removing un-needed header!

app/services/create_github_repo_service.rb
```diff
@@ -248,6 +248,16 @@ def report_template_error_to_failbot(error, options)
...
 
+  def github_organization_with_access
+    github_client = assignment.creator.github_client
+    starter_code_repository = github_client.repository(assignment.starter_code_repo_id)
+
+    return organization.github_organization if !starter_code_repository.private ||
+        starter_code_repository.owner.login == exercise.organization_login
+
+    GitHubOrganization.new(github_client, starter_code_repository)
+  end
```
shaunakpp: Awesome! 

Can we use a `attr_accessor`  like `@organization_client` for this? Use the logic to initialize the correct `GitHubOrganization` object, and use that instead of a method call? It can also help us out in testing where we can mock/stub the object.

Another idea would be:

We currently delegate `organization` to `Exercise`, so maybe we can move the logic there? Something on the lines of `Exercise#find_github_organization_with_access(starter_code_repository)` which will modify the `@github_organization` `attr_accessor` on `Exercise` and then we can use it in this service everywhere.

EricPickup: Good idea! I combined both your ideas into an instance variable `@github_organization` on `Exercise`

EricPickup:
@shaunakpp Sorry for the force-push, but let me know how this looks!

EricPickup:
Closing this and moving to #2286 where there are no unrelated changes/cassettes

PR #2266, opened 2019-08-20
Improve UX/UI for Roster List Shown when an Assignment is Accepted

### LINKED ISSUE #1882: Improve UX/UI for roster list on assignment invitation
When joining the classroom roster, the roster list box takes up only about 1/3 of the vertical space and lists around 7 students at a time which seems inefficient considering most rosters will have 50+ students.

We should expand this list to take up more of the page which will make it easier to navigate and find your identifier. 

Also, when hovering over the box I don't see a scroll bar to navigate the list. The only option I can find is to use trackpad gestures or a scroll wheel to scroll through the list. This could be difficult for less tech-savvy users.

_This image shows the roster list box for a 40+ student roster, only showing 7 students at a time in the view._
Screen Shot 2019-06-18 at 11 41 05 AM

JessRudder:
Closes #1882 

The roster area was previously set to be 200px regardless of the screen size. This updates it to be a portion of the view height.  There will still be some unused space on very large screens or screens with high resolutions. I'm open to suggestions for ways of handling that better with CSS.

Here's an example of the space it will take up at 50% zoom:
image

While I was in there, I also replaced a number of custom classes with Primer classes so we can benefit from the accessibility and responsiveness of Primer.

I also did some small look and feel updates to the roster list to make it more obvious which name you were about to select. We now highlight the row you're about to select when you hover over it!

And, I added a flash notice to indicate you were successfully linked on the roster. It indicates that you should reach out to the instructor if you connected with the wrong listing as there is not currently a way for a student to unlink on their own:
image

I've tested on Chrome, Safari and Firefox but all screenshots below are from Chrome.

### Updated Screenshots
Updated Roster with One Student:
image

Updated Roster with Multiple Students:
image

### Current Screenshots
Current Roster with One Student:
image

Current Roster with Multiple Students:
image


JessRudder:
@femmebot this PR was addressing an issue where the roster list took up very little space on large/high-resolution screens since it was frozen to 200px. I tried to find a middle-ground solution of allowing it to be longer on high-res screens without letting it be so long that you had to scroll extra to see the skip button. I also added the `skip` as a link in the text at the top though it's not quite as visible as the button.

I definitely think there's a lot that could be done to improve the roster management experience from the student and teacher side. From our meeting today, it sounds like there is a larger conversation that's happening with product/design around roster management. Is there anything specific that should be included here as part of this PR?

shaunakpp:
I've created a high-level issue for roster improvements, should I add this update to it?

JessRudder:
@shaunakpp if that issue is focused solely on the teacher side of things, no need to add this. If we want it to track all things roster, definitely add this in! If we go that route, I'd love to get some thoughts around student-side improvements like searching (rather than scrolling) to find a name.  We should probably add something in for security as well (currently, everyone sees the entire list of unclaimed roster entries - it feels weird to provide an entire class list of names/emails/school ids/etc to the other students. I'm wondering if there's a way to show less without adding too much friction to the roster process.

app/controllers/assignment_invitations_controller.rb
```diff
@@ -73,6 +73,9 @@ def success; end
   def join_roster
     super
 
+    entry = organization.roster.roster_entries.find_by(user_id: current_user.id)
+    flash[:success] = "Your account is linked to #{entry.identifier} on the roster. If this is wrong, please reach out to your instructor."
```
EricPickup: Great idea!

Issue #1580, opened 2018-09-18
Show teachers their plan constraints in the assignment creation screen 


BenEmdon:
### Feature request :sparkles:
This was initially implemented in https://github.com/education/classroom/pull/924
And reverted in https://github.com/education/classroom/pull/944

BenEmdon:
@d12 we should try achieve this as part of the work being done with assignment setup. This should be a small piece of work.

d12:
🙌 Adding to the project board!

spinecone:
I've been poking around the original PR, the revert, and slack discussion around them, and it looks like what happened was:
- https://github.com/education/classroom/pull/924 was deployed which started causing some kind of error
- https://github.com/education/classroom/pull/943 was created because 924 was using an incorrect method name (`private_repos_available?` rather than `has_private_repos_available?`)
- 943 didn't deploy correctly for some reason (the heaven log has already been pruned so I can't see what the error was 😭 ) so we just reverted the whole PR with https://github.com/education/classroom/pull/944

Does that sound accurate? If so I might attempt what 943 was originally trying to do and just fix the method name, and see if that works 🤞 .

PR #2269, opened 2019-08-20
unrevert plan info in assignment form

spinecone:
For https://github.com/education/classroom/issues/1580

This reverts https://github.com/education/classroom/pull/944 , which originally reverted https://github.com/education/classroom/pull/924 . I had to make some minor adjustments based on changes between then and now but this is still mostly @d12 's work.

This change is intended to default `public_repo` to "false" on new repo assignment if the classroom owner has private repos enabled so that student work isn't accidentally publicized. It also disables the "false" option if they do not have private repos enabled.

New assignment form when the org does not have private repos:
Screen Shot 2019-08-21 at 12 51 33 PM

New assignment form when the org does have private repos:
Screen Shot 2019-08-21 at 12 51 46 PM


d12:
Nice detective work! Tbh, I have _no_ idea what happened here, this was so long ago. 🤞

PR #2273, opened 2019-08-21
Assignment page UI changes

stephaniegiang:
## What
Taking the changes from #1835 🎉
This removes more custom css and uses primer. Also super different view, opting for tabs instead of buttons. I did some tweaks from the original PR so that it is clear what each setting tab is. 

image

And for reference, the current page: 
image


stephaniegiang:
@femmebot this PR isn't exactly done yet since I have to work on group assignments. But I want to get your feedback early on. What do you think of the overall design?

I also have a gif showing how it looks mobile. Not sure what I should do with the enable invitations part. 

9v83rUQ80p

stephaniegiang:
@femmebot the primary action is to wait for students to accept the assignment. Then their repositories will show up.

stephaniegiang:
Going to close this PR, we are currently working on a different re-design!

Issue #2268, opened 2019-08-20
Improve roster creation flow


shaunakpp:
This is a high-level issue of the three part proposed changes that for improving the Roster creation process. With LMS integrations, rosters have become an essential part of classroom and we could work on improving this experience for teachers. Here are the following improvements that we could work on:

- [ ] Move roster updates(#2233 ) and creation to background jobs(#2242)
- [ ] Move/Copy the link to roster management from settings to a tab on classroom show page.
- [ ] Rename rosters to students on the UI.
- [ ] Add a setup page for creating rosters, with better messages and updates
- [ ] Roster updates via LMSes are a great version 1, but we can iterate over it by keeping rosters in LMS and classroom in complete sync. E.g. delete roster_entry of a student no longer part of the roster in LMS.
- [ ] Teachers can accept invitations to assignments just to see a student's point of view, this already happens, but is not mentioned anywhere. Should we add communication that they can do that, or figure out any other solution that can enable teachers to get a student's PoV?

cc: @femmebot @stephaniegiang

PR #2274, opened 2019-08-21
[WIP] Move rosters to a new tab and rename to students

stephaniegiang:
## What
Part of https://github.com/education/classroom/issues/2268
This PR moves rosters for ease of access. We also renamed rosters to students so that it is more clear. Rosters used to live in the settings tab. 

| Action | Screenshot |
|--------|------------|
| View of all students in classroom | image |
| Creating a new roster | image |
| Deleting roster | image |

### DIFF
--- a/app/controllers/orgs/rosters_controller.rb
+++ b/app/controllers/orgs/rosters_controller.rb
@@ -62,7 +62,7 @@ module Orgs
         create_statsd(lms_user_ids: params[:lms_user_ids])
         imported_students_lms_statsd(lms_user_ids: params[:lms_user_ids])
         flash[:success] = \
-          "Your classroom roster has been saved! Manage it <a href='#{roster_url(current_organization)}'>here</a>."
+          "Students have been added to your classroom! Manage students <a href='#{roster_url(current_organization)}'>here</a>."
 
         redirect_to organization_path(current_organization)
       else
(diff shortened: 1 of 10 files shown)


EricPickup:
I **love** the idea of moving the roster out of settings and into here.

I'm not sure about renaming everything "roster" to "students" because I think a decent chunk of classrooms don't use rosters and this could get confusing. E.g. a classroom without a roster sees the "Students" tab and is confused because they've had 30 github users accept their assignment but their students list is empty because they didn't use rosters.

This could be avoided by adding a small description to the top of the "Students" tab?

EricPickup:
@femmebot @stephaniegiang nevermind, I can see on the roster setup screen you describe it pretty well. I was looking at the other tabs (after roster creation) and noticed there wasn't any description of rosters.

LGTM 😄

stephaniegiang:
@femmebot are we good to go with this PR? I'll fix the errors and then I can deploy.

stephaniegiang:
Oh that looks cool, I'll look into it 😃

stephaniegiang:
@femmebot I don't think we can do a popover, just because of how the navigation is built. But we can do a flash message.

stephaniegiang:
@femmebot For sure! I'm excited for the features page 😄

andrewbredow:
Closing, we're taking this in a different direction.

Issue #2212, opened 2019-08-08
Deleting roster entries that are not associated with an active roster


stephaniegiang:
We currently do not delete roster entries when we delete a roster.

With this PR https://github.com/education/classroom/pull/2211 we should think about running a job to delete the 61,000+ roster entries that are not associated to a roster.

JessRudder:
@stephaniegiang, looks like the related PR has been merged in. Is there a cleanup issue we need to do as a follow on (to remove the roster-less roster entries)?

If not, please feel free to close this ticket!

PR #2276, opened 2019-08-21
Remove roster entries from deleted rosters

stephaniegiang:
## What
Closes https://github.com/education/classroom/issues/2212

This task removes all roster entries from deleted rosters.


lib/tasks/remove_deleted_roster_entries.rake
```diff
@@ -0,0 +1,27 @@
...
+
+  BATCH_SIZE = 100
+
+  roster_entries_deleted = 0
+  batch_num = 1
+
+  non_existing_rosters = RosterEntry.where.not(roster_id: Roster.pluck(:id)).pluck(:roster_id).uniq
+
+  non_existing_rosters.each do |roster_id|
+    roster_entries_to_delete = RosterEntry.where(roster_id: roster_id)
```
shaunakpp: this will find all the roster entries and generate an array of ActiveRecord objects, we can fix that by not assigning the output to `roster_entries_to_delete` and chain the query with find in batches. let me know if this works.

lib/tasks/remove_deleted_roster_entries.rake
```diff
@@ -0,0 +1,27 @@
...
+
+    roster_entries_to_delete.find_in_batches(batch_size: BATCH_SIZE) do |roster_entries|
```
shaunakpp:
```suggestion
    RosterEntry.where(roster_id: roster_id).find_in_batches(batch_size: BATCH_SIZE) do |roster_entries|
```

lib/tasks/remove_deleted_roster_entries.rake
stephaniegiang: Ahhh that makes more sense

stephaniegiang:
@shaunakpp Made the changes!

PR #2283, opened 2019-08-22
Run template repo validator when import method changed

### LINKED ISSUE #2271: When changing from porter -> template, check if repo is a template
If you update an existing assignment with starter code from porter -> template, it doesn't seem like we check if the repo is a template. I gave it a go and the assignment update succeeded.

EricPickup:
## What
Fixes #2271 

Run the validator that checks if a repo is a template when `starter_code_repo_id` is changed (already doing this) and/or when `use_template_repos` is changed.

Keep in mind that even if `use_template_repos` is changed to `false`, the validator checks `assignment#use_template_repos` and will not be run if the teacher is switching to porter.


app/models/assignment.rb
```diff
@@ -39,7 +39,8 @@ class Assignment < ApplicationRecord
 
   validate :uniqueness_of_slug_across_organization
   validate :starter_code_repository_not_empty, if: :will_save_change_to_starter_code_repo_id?
-  validate :starter_code_repository_is_template, if: :will_save_change_to_starter_code_repo_id?
+  validate :starter_code_repository_is_template,
+    if: -> { :will_save_change_to_starter_code_repo_id? || :will_save_change_to_template_repos_enabled }
```
jeffrafter: The code here works good and I'm comfortable moving forward with it. I wonder, though, based on your comment if we should instead only check when the option is enabled and the value is not blank?. With this code if someone points to a third party template and the owner unchecks the template box... it is going to be very difficult to change this.

jeffrafter: Ah, I misunderstood and can clearly see this checked in the concern: https://github.com/education/classroom/blob/0ec9560df085de5b341d857b6ae1c901097f4991/app/models/concerns/starter_code_importable.rb#L30 👏

PR #2284, opened 2019-08-22
Log LMS import errors

stephaniegiang:
## What
This will log LMS import errors and hopefully give us a better idea of what is happening with Canvas hosted on Instructure and other LMSs as well.


app/controllers/orgs/rosters_controller/lti_dependency.rb
```diff
@@ -71,7 +71,14 @@ def lms_membership
 
       begin
         membership_service.students(body_params: current_organization.lti_configuration.context_membership_body_params)
-      rescue Faraday::ClientError, JSON::ParserError
+      rescue Faraday::ClientError, JSON::ParserError => error
```
EricPickup: What do you think about moving this error builder/reporter to its own method like we do here? https://github.com/education/classroom/blob/master/app/services/create_github_repo_service.rb#L89

stephaniegiang: True. That is a lot better. I'll make those changes 😄

PR #2286, opened 2019-08-22
Use assignment creator's token when random token fails

EricPickup:
## What
Back-end solution to #2222 

If the starter code repository is private and does not belong to the classroom organization, use the assignment creator's token to create the student repo.

In all other cases, continue our current method of grabbing a random teachers token to create the students repositories.


spinecone:
Rather than creating a fake `GithubOrganization` around the assignment creator's token and needing to worry about other places where we expect it to behave like an organization (such as checking if it has a free/paid plan, which the fake org returns `nil` for), I'm working on thinking of a different approach for this problem.

Instead of changing which `GithubOrganization` object calls `create_repository`, it might make more sense for us to continue just using the real organization that owns the assignment, and having it try the assignment creator's token itself if it doesn't have access with its own random token.

PR #2287, opened 2019-08-22
Remove Session ID from API 

stephaniegiang:
## What
Fixes https://github.com/github/education/issues/952

For the API, we are currently using session's to verify the user. We should instead not rely on sessions.


stephaniegiang:
Hey @anglinb could you take a look at this PR?

PR #2288, opened 2019-08-23
Update Canvas help page

stephaniegiang:
## What
This updates Canvas' help page, to say that we do not support free teacher accounts. This is primarily due to the fact that Canvas does not allow the membership service setting to be turned on for free accounts.

### DIFF
--- a/docs/setup-canvas.md
+++ b/docs/setup-canvas.md
@@ -2,6 +2,9 @@
 
 These steps will walk you through the process of configuring GitHub Classroom as an external tool within Canvas.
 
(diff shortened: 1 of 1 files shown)


docs/setup-canvas.md
```diff
@@ -2,6 +2,9 @@
 
 These steps will walk you through the process of configuring GitHub Classroom as an external tool within Canvas.
 
+**Note**: We currently do not support free teacher accounts on Canvas
```
stephaniegiang: Not sure if I like the wording on this, but please give any suggestions.

PR #2289, opened 2019-08-23
fix default public_repo

spinecone:
Followup to https://github.com/education/classroom/pull/2269 (original PR) and https://github.com/education/classroom/pull/2280 (bug fix for when the original PR turned out to be reversing the value of public_repo in assignment forms).

I paired with @jeffrafter on redoing these radio buttons and writing tests for them. We differentiated between the `checked` and `value` parameters on the radio buttons to fix the original bug that was caused by https://github.com/education/classroom/pull/2269 , and I added a `views` directory to specs so that we could test that the radio buttons are rendered correctly.

### DIFF
--- a/app/views/assignments/_assignment_form_options.html.erb
+++ b/app/views/assignments/_assignment_form_options.html.erb
@@ -16,7 +16,7 @@
 <dl class="<%= view.form_class_for(:public_repo) %> my-4">
   <div class="form-checkbox">
     <label>
-      <%= f.radio_button :public_repo, true %>
+      <%= f.radio_button :visibility, "public", checked: !view.private_repos_available? %>
       <div class="d-flex">
         <div>
           <%= octicon 'repo', height: 32, class: "text-gray mr-2" %>
@@ -31,7 +31,7 @@
 
   <div class="form-checkbox">
     <label>
-      <%= f.radio_button :public_repo, false, disabled: !view.private_repos_available? %>
+      <%= f.radio_button :visibility, "private", checked: view.private_repos_available?, disabled: !view.private_repos_available? %>
       <div class="d-flex">
         <div>
           <%= octicon 'lock', height: 32, class: "text-orange mr-2" %>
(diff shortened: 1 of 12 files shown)


app/views/assignments/_assignment_form_options.html.erb
```diff
@@ -16,7 +16,7 @@
 <dl class="<%= view.form_class_for(:public_repo) %> my-4">
   <div class="form-checkbox">
     <label>
-      <%= f.radio_button :public_repo, true %>
+      <%= f.radio_button :visibility, "public", checked: !view.private_repos_available? %>
```
stephaniegiang: This partial also shows up when you click "Assignment settings". Will the `checked` override the actual current visibility of an assignment?

app/views/group_assignments/_group_assignment_form_options.html.erb
```diff
@@ -16,7 +16,7 @@
 <dl class="<%= view.form_class_for(:public_repo) %> my-4">
   <div class="form-checkbox">
     <label>
-      <%= f.radio_button :public_repo, true %>
+      <%= f.radio_button :visibility, "public", checked: !view.private_repos_available? %>
```
stephaniegiang: Same as above for this partial as well.

spec/views/orgs/roster_entries/assignment_repos/linked_accepted.erb_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "display student username", type: :view do
```
stephaniegiang: Also want to add, LOVE these new view tests! ❤️

app/views/assignments/_assignment_form_options.html.erb
spinecone: I'll change this to look at the assignment's visibility if the assignment has been persisted!

PR #2333, opened 2019-08-29
Update docs with LTI version + Names and Roles provisioning

stephaniegiang:
## What
This PR updates our docs, to specify more clearly the LTI version we support. As well as another dependency needed: `LTI Names and Roles Provisioning 1.X`

### DIFF
--- a/docs/connect-to-lms.md
+++ b/docs/connect-to-lms.md
@@ -6,7 +6,7 @@ GitHub Classroom integrates with your institution's learning management system (
 - Assignment creation (Planned)
 - Submission syncing (Planned)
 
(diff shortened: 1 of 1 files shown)


docs/connect-to-lms.md
```diff
@@ -8,6 +8,10 @@ GitHub Classroom integrates with your institution's learning management system (
 
 Classroom works with any LMS that uses the Learning Tools Interoperability (LTI) standard. Using LTI keeps your information safe and secure. LTI is an industry-standard protocol and Classroom's use of LTI is certified by the Instructional Management System Global Learning Consortium.
```
stephaniegiang: Should I remove the `Classroom works with any LMS that uses the Learning Tools Interoperability (LTI) standard`?

PR #2335, opened 2019-08-30
Rescue only JSON::ParserError

stephaniegiang:
## What
@spinecone brought up a really good point. We should only be rescuing the `JSON::ParserError` error so that we aren't swallowing other errors.

### DIFF
--- a/lib/github_classroom/lti/membership_service.rb
+++ b/lib/github_classroom/lti/membership_service.rb
@@ -74,6 +74,7 @@ module GitHubClassroom
           json_membership = JSON.parse(raw_data)
         rescue JSON::ParserError
           Rails.logger.error("raw_data: #{raw_data}")
+          # raising new error as the encoded raw_data may be causing failbot issues
           raise JSON::ParserError
         end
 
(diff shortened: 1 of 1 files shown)


lib/github_classroom/lti/membership_service.rb
```diff
@@ -72,7 +72,7 @@ def parse_membership(raw_data)
       def parse_membership_service(raw_data)
         begin
           json_membership = JSON.parse(raw_data)
-        rescue
+        rescue JSON::ParserError
           Rails.logger.error("raw_data: #{raw_data}")
           raise JSON::ParserError
```
spinecone: Could we add a comment about why we're not re-raising the error that was rescued? (talking to @jeffrafter it sounded like it's because we might cause problems with our logging applications if we log weird encoding).

stephaniegiang: We are re-raising the error that was rescued in line 77.

spinecone: The error we're raising is a new error of the same type. To re-raise the exact error that was rescued, we'd need to set it to a variable with something like `rescue JSON::ParserError => e`.

stephaniegiang: Made the changes 😄

PR #2338, opened 2019-09-02
Fix GitHubClassroom.github_url when using Enterprise

favonia:
Sorry that I do not know how to create a regression test, but this is definitely a bug to fix.

### DIFF
--- a/config/initializers/github_client.rb
+++ b/config/initializers/github_client.rb
@@ -24,7 +24,7 @@ module GitHubClassroom
 
   def self.github_url
     if enterprise?
-      Rails.application.secrets.github_enterprise_url.present?
+      Rails.application.secrets.github_enterprise_url
     else
       "https://github.com"
     end


config/initializers/github_client.rb
```diff
@@ -24,7 +24,7 @@ def self.enterprise?
 
   def self.github_url
     if enterprise?
```
stephaniegiang: Hi @favonia 👋Thank you so much for contributing!

Looking at this PR, I think what you are trying to do is this: 

```suggestion
     if enterprise? && Rails.application.secrets.github_enterprise_url.present?
```

favonia: I do not think this is necessary. `enterprise?` returns `true` only when `Rails.application.secrets.github_enterprise_url.present?` return trues. Please tell me if I was misguided.

https://github.com/education/classroom/blob/6561b4630e45e5455fdda6f479cde2ec6aaf54d5/config/initializers/github_client.rb#L19-L23

jeffrafter: Confirmed this is a bug and great catch, thanks @favonia! I'll work on a regression test!

jeffrafter: Actually in digging into this more, I don't expect it to regress and I think your fix is great.

jeffrafter (approved):
Thanks for the fix @favonia, your help has been extremely valuable in getting Classroom working on Enterprise!

Issue #2337, opened 2019-09-01
Title of Organization does not match GitHub's parameters


jules2689:
**Describe the bug**

My Organization has a long title:
<img width="995" alt="image" src="https://user-images.githubusercontent.com/3074765/64078486-62c30c80-cca9-11e9-8801-c7c6ab3a3a98.png">

When I try to grant access in Classroom, it fails without a seemingly valid reason:
![2019-09-01 11-13-09 2019-09-01 11_13_47](https://user-images.githubusercontent.com/3074765/64078512-9aca4f80-cca9-11e9-883b-f2706119fefc.gif)

It took a moment to realize what was up as my org is only 5 letters long (then I remembered the long name in GitHub).

**To Reproduce**
1. Create an org with a long name
2. Grant access and try to create classroom on that org
3. Fail and see error

**Expected behavior**
Match the constraints of github.com

jules2689:
Related PR: https://github.com/education/classroom/pull/248

@tarebyte do you have context as to why a limit of 60 was placed on orgs? https://github.com/education/classroom/pull/248#issuecomment-144863273 Seems to suggest that there's no specific reason.

spinecone:
Looking at the validations on github.com, the limit for organization login names is only 40 characters, but profile names can be 255 characters. If we're copying over the profile name we should probably be using the 255 character limit instead.

jules2689:
@spinecone For clarification, for [this org](https://github.com/dcsil/) where is says `Department of Computer Science Innovation Lab`, that is what failed. It used to say `Department of Computer Science Innovation Lab [University of Toronto]`

spinecone:
Discussed with @femmebot and @stephaniegiang , the changes we'll want to make are:
- increase organization title limit to 255
- change organization thumbnails in dashboard from 4 per row to 3 per row
- remove @ from organization titles in dashboard

jules2689:
https://classroom.github.com/classrooms/54954454-a-really-long-name-that-never-ends-it-goes-on-and-on-my-friends-some-people-starting-saying-it-not-know-what-it-was-but-they-ll-continue-saying-it-forever-just-because-goto-0-classroom-1

It works with https://github.com/testing-dcsil :smile:


I'll let you close this as it sounds you had other things you were using this to track as per https://github.com/education/classroom/issues/2337#issuecomment-528065535

PR #2341, opened 2019-09-04
allow longer organization titles and display logins better

spinecone:
For https://github.com/education/classroom/issues/2337

- Changes the organization title maximum length from 60 to 255 characters (to match the limit on github.com organization profile names)
- Removes the `@` from the organization login on the new classroom form
- Changes the organizations on the new classroom form to be displayed 3 per row rather than 4 per row
- No longer truncating the organization login, which can be up to 40 characters

The new classroom form now looks like this:
Screen Shot 2019-09-05 at 10 37 41 AM

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -26,7 +26,7 @@ class Organization < ApplicationRecord
   validates :github_id, presence: true
 
   validates :title, presence: true
-  validates :title, length: { maximum: 60 }
+  validates :title, length: { maximum: 255 }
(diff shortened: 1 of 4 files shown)


spinecone:
@stephaniegiang the changes in this PR are kind of unrelated and it's a little confusing. We only increased the org title limit to 255 characters. The classroom form is displaying the org login, which can only be up to 40 characters.

stephaniegiang:
Ahhh that makes more sense. Looks good then!

PR #2351, opened 2019-09-09
Add `onboarding_redesign` feature flag helper method

### LINKED ISSUE #2349: Create a feature flag for the new design called `redesign`
We're working on tightening up the design and flow of the onboarding process.

To avoid the confusion of a mixed process where some parts of the flow are new and some are old, we'll be keeping the new flow behind a feature flag until it's ready to launch.

JessRudder:
Closes #2349

This adds the `onboarding_redesign_enabled?` feature flag helper method. I've also created the feature flag in devtools so it will be available in production.

This will enable us to easily check if the flag is enabled in future development as we work on the onboarding redesign.

Because onboarding can start before a person is logged in, I've set this up to check for global enablement or enablement for a specific user. This will make local dev easier since you can enable globally in order to get the updated version even if your local user is logged out.

### DIFF
--- a/app/controllers/application_controller/feature_flags_dependency.rb
+++ b/app/controllers/application_controller/feature_flags_dependency.rb
@@ -38,8 +38,8 @@ class ApplicationController
   end
   helper_method :classroom_visibility_enabled?
 
-  def redesign_enabled?
-    GitHubClassroom.flipper[:redesign].enabled? || (logged_in? && current_user.feature_enabled?(:redesign))
+  def onboarding_redesign_enabled?
+    GitHubClassroom.flipper[:onboarding_redesign].enabled? || (logged_in? && current_user.feature_enabled?(:onboarding_redesign))
   end
-  helper_method :redesign_enabled?
+  helper_method :onboarding_redesign_enabled?
 end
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller/feature_flags_dependency.rb
```diff
@@ -37,4 +37,9 @@ def classroom_visibility_enabled?
     logged_in? && current_user.feature_enabled?(:classroom_visibility)
   end
   helper_method :classroom_visibility_enabled?
+
+  def redesign_enabled?
+    GitHubClassroom.flipper[:redesign].enabled? || (logged_in? && current_user.feature_enabled?(:redesign))
```
spinecone: Should we specify that this is an `onboarding_redesign` ? (unless we're covering more than that, just wanted to make it a little clearer what's involved)

JessRudder: I'm down for updating. The original note created by @jeffrafter  said to create one called `redesign` so I just went with that, but, I definitely like greater specificity. 

@jeffrafter is it alright if we make the feature flag `onboarding_redesign`?

JessRudder: We spoke offline and more specific is 👍

app/controllers/application_controller/feature_flags_dependency.rb
```diff
@@ -37,4 +37,10 @@ def classroom_visibility_enabled?
     logged_in? && current_user.feature_enabled?(:classroom_visibility)
   end
   helper_method :classroom_visibility_enabled?
+
+  # rubocop:disable Metrics/LineLength
```
JessRudder:
```suggestion
```

PR #2352, opened 2019-09-09
remove linter rule for line limit

spinecone:
I've seen myself/others get CI failures for exceeding the 120 character line limit a few times, and my preference is to review line length on context rather than a hard rule. We're also disabling these metrics 20 times across the codebase already. I'm open to feedback if folks feel strongly that we should keep these though!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,12 +19,6 @@ TrailingBlankLines:
 Documentation:
   Enabled: false
 
-Metrics/ClassLength:
-  Max: 120
-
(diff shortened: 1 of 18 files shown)


JessRudder:
Looks like deleting causes it to go to a default of enabled. I think we need:

```
Metrics/LineLength:
    Enabled: false

Metrics/ClassLength:
  Enabled: false
```

I would have added it as a suggestion, but, I discovered you can't do suggestions on deleted lines.

PR #2354, opened 2019-09-10
Add a split between current header and v2 header

### LINKED ISSUE #2353: Create a copy of the header/layout which is new/old based on the `redesign` feature flag
The redesign will include a new header and we'll need to be able to serve both side-by-side (depending on the state of the feature flag).

For this we'll need to do the following:

Add logic for showing one header vs the other (depending on the feature flag)

Add the front end code for the new header

JessRudder:
Closes #2353 

When `onboarding_redesign` flag is enabled, people should see the new v2 header.

Old Header:
image

New Header:
Coming Soon!!!!

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -20,7 +20,11 @@
     <%= render 'peek/bar' %>
     <main class="bg-gray-light">
       <%= render 'shared/message_of_the_day' %>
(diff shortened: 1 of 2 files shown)


JessRudder:
I'm going to close this out for now. We can revisit when the onboarding redesign work is ready to move forward (as this may or may not be the right approach for the final plan).

PR #2366, opened 2019-09-11
Remove lti_launch feature flag and related helper methods

JessRudder:
This is a clean up PR to remove the feature flag and related helper methods for the `lti_launch` flag.

This removes direct references to the `lti_launch` feature flag and the following helper methods that existed to check whether or not the flag was enabled:

`ensure_lti_launch_flipper_is_enabled`
`lti_launch_enabled?`
`verify_lti_launch_enabled`

I removed all tests that tested the feature with the flag disabled.

I updated the 'flag enabled' tests to indicate it is expected behavior and not feature flag related now.

There were a few places where `||` was used to test whether this flag or another were true. In those cases, since removing the flag is the equivalent of it always being true, I removed the entire check (even though the other flag still exists).

### DIFF
--- a/app/controllers/application_controller/feature_flags_dependency.rb
+++ b/app/controllers/application_controller/feature_flags_dependency.rb
@@ -5,10 +5,6 @@ class ApplicationController
     not_found unless team_management_enabled?
   end
 
-  def ensure_lti_launch_flipper_is_enabled
-    not_found unless lti_launch_enabled?
-  end
-
   def ensure_google_classroom_roster_import_is_enabled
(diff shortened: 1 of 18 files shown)


JessRudder:
Removing the auto-request for @andrewbredow because I want to make sure this is reviewed by @stephaniegiang instead since she is very familiar with the LTI integration code.

Though, in retrospect, I could have just added her as an additional review.....not really sure what the protocol is with the auto-assignees. Definitely feel free to review it too, Andrew!

andrewbredow:
I’ll remove myself from the auto assignment :)

PR #2368, opened 2019-09-12
[DO NOT MERGE] Remove archive_classroom feature flag code

JessRudder:
Note: It looks like this is currently only staff shipped. Need to follow up on whether this should be shipped outside of the staff group. If so, we should ship and then merge this PR.

If this is not something we intend to ship, we should consider removing all related code.


JessRudder:
The feature was GA-ed and the feature flag has been removed. Closing this PR out as it is no longer necessary.

PR #2371, opened 2019-09-13
Change order of sort on Classrooms

stephaniegiang:
## What
This PR changes the order of sort on Classrooms. This makes the default to "Newest first". 

image


stephaniegiang:
@femmebot Ooops! I forgot to change the main index query as well. Thanks for catching that 😄

PR #2388, opened 2019-09-20
Add ability to filter classrooms by archived/active

### LINKED ISSUE #602: Archiving a Classroom

JessRudder:
Closes #602 

It's behind a feature flag, but, we already have the ability to archive classrooms.

This PR finishes that work by adding a `view` filter that allows teachers to filter their classrooms by `All`, `Archived`, and `Active`. These filters can be combined with the existing text search and sort filters to more easily drill down into large classroom lists.

The search and sort that was already in place expects that there will only be one drop down menu and relied on using jquery to update links. It also expects that there will only be one drop down sorting option. Having more than one causes problems in both the javascript and in the shared views. I opted to create new functionality for this searching and sorting. This allowed me to use Primer-based components for the drop down menus and write JavaScript that will allow for as many dropdowns as desired (without hacking links together). In future PRs we can update the other search/sort being used elsewhere in the code so that it uses this more flexible and robust solution.

Currently, the default on page load is set to `All` (which will maintain current behavior of loading all of a teacher's classrooms); however, we can consider making the default `Active` if that is not going to be confusing.

This is based on the designs that can be found [here](https://www.figma.com/file/97NA4vopFIccJHuzVVSO8V/Education?node-id=425%3A273)

view-archived

Included Fixes:
 - Params in the URL now update consistently whether you're using the dropdown filter or the search bar
 - If you navigate to the `/classroom` endpoint with filtering params in the URL, you will now see the filtered view (previously it would ignore those on page load)
 - If someone passes in a dummy param, it will no longer be added as the filtering menu label


app/controllers/organizations_controller.rb
```diff
@@ -9,13 +9,12 @@ class OrganizationsController < Orgs::Controller
...
   before_action :set_filter_options,                  only: %i[search index]
+  before_action :set_filtered_organizations,          only: %i[search index]
 
   skip_before_action :ensure_current_organization,                         only: %i[index new create search]
   skip_before_action :ensure_current_organization_visible_to_current_user, only: %i[index new create search]
 
-  def index
-    @organizations = current_user.organizations.order(created_at: :desc).page(params[:page]).per(12)
-  end
+  def index; end
```
JessRudder: Previously on a page load we were always loading all of the organizations regardless of whether or not params were provided. This change (setting `@organizations` using `set_filter_options` allows people to navigate directly to this page with a filtered URL and see only the classrooms that match those filters)

app/controllers/organizations_controller.rb
```diff
@@ -146,25 +137,52 @@ def set_users_github_organizations
...
+      .per(12)
   end
+  # rubocop:enable Metrics/MethodLength
 
   # Check if the current user has any organizations with admin privilege,
   # if so add the user to the corresponding classroom automatically.
   def add_current_user_to_organizations
     @users_github_organizations.each do |github_org|
-      user_classrooms = Organization.where(github_id: github_org[:github_id])
+      user_classrooms = Organization.includes(:users).where(github_id: github_org[:github_id])
```
JessRudder: I added this to get one of the eager load checks to stop yelling at me every page load.

app/models/organization.rb
```diff
@@ -39,7 +39,10 @@ class Organization < ApplicationRecord
   scope :order_by_oldest, ->(_context = nil) { order(created_at: :asc) }
   scope :order_by_title,  ->(_context = nil) { order(:title) }
 
-  scope :search_by_title, ->(query) { where("title ILIKE ?", "%#{query}%") }
+  scope :search_by_title, ->(query) { where("organizations.title ILIKE ?", "%#{query}%") }
```
JessRudder: This was failing tests due to being an ambiguous field so I added in the explicit reference to the `organizations` table.

app/assets/javascripts/organization-filtering.js
```diff
@@ -0,0 +1,36 @@
...
+        clearTimeout(timer);
+      }
+      timer = setTimeout(callback, ms);
+    };
+  })();
+
+  $("#js-filtering-component").on("change keyup input", function(e) {
+    const componentContext = $(e.target).closest("#js-filtering-component");
+    const $searchForm = componentContext.find("#js-filtering-form");
+    const formData = $searchForm.find('input[name!=utf8]').serialize();
```
jeffrafter: The above two fixes work... but then we get to this line which would crash if we make those changes. The same pattern applies (but I wrap it in $() to get the serialize function):

```suggestion
    const formData = $(document.getElementById('your-classrooms-filter')).serialize();
```

app/assets/javascripts/organization-filtering.js
```diff
@@ -0,0 +1,36 @@
...
+
+    history.replaceState(null, '', '?' + formData);
+    debounce(function() { $searchForm.submit(); }, 300);
+  });
+
+  $("#js-filtering-component .SelectMenu-item").on("change", function(e) {
+    const $clickedItem = $(e.target).closest(".SelectMenu-item");
+    const $currentActiveItem = $clickedItem.siblings("[aria-checked='true']");
```
jeffrafter: Relying on `aria-checked` feels new to me. Is this a pattern we're using elsewhere?

app/controllers/organizations_controller.rb
```diff
@@ -146,25 +137,52 @@ def set_users_github_organizations
     end
   end
 
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
   def set_filter_options
     @sort_modes = Organization.sort_modes
+    @view_modes = Organization.view_modes
+
+    @current_sort_mode = if @sort_modes.dig(params[:sort_by])
```
jeffrafter: Non-blocking: Just curious why you used `dig` here (and below) and not just `@sort_modes[params[:sort_by]]` or `@sort_modes.keys.include?(params[:sort_by])`

app/assets/javascripts/organization-filtering.js
JessRudder: In a previous iteration where we rolled our own dropdown menus, we were relying on the addition of a `selected` class to indicate which item was selected. With that, we were looking for that class to determine which item had previously been selected.

With the Primer dropdown component, the item is selected using `aria-checked=true`. I was keeping the spirit of the code (finding the previously selected item by looking for the markup that indicated the item was selected) but updating it to reflect the way the Primer component worked.

I've updated to add the `selected` class so we can search based on the class instead. This would be consistent with what we were doing before; however, it means we now have to update both the class and the `aria-checked` attribute. It's not a large amount of extra code, but it does feel a bit `extra`. Is it a best practice in JS no to select using aria attributes?

app/assets/javascripts/organization-filtering.js
JessRudder: This won't quite work. The `your-classrooms-filter` id is only the query part of the form and we want to serialize all of the fields. It looks like we're not using that id anywhere so I've gone ahead and removed it.

If we don't care about including `utf8=%E2%9C%93` in the URL params, we can do `$(searchForm).serialize()`.

If we want to stay consistent with the other forms on Classroom whose params end up in the URL, we should do `$(searchForm).find('input[name!=utf8]').serialize()` to filter out the utf8 param.

For consistency, I've got `$(searchForm).find('input[name!=utf8]').serialize()` right now, but, I'm happy to change it (no strong feelings on this one).

app/controllers/organizations_controller.rb
JessRudder: Ah....this was me completely forgetting why John often recommended dig to me. Those were cases where we were going multiple levels deep in a hash and one of the keys might not be there (which would otherwise error). That doesn't apply here (so I've gone ahead and replaced it with `.keys.include?`.

Thanks for catching this!

app/assets/javascripts/organization-filtering.js
jeffrafter: Ah, that makes a lot of sense. Let's leave it as you have it! 🙌

app/assets/javascripts/organization-filtering.js
jeffrafter: No not necessarily. Your reasoning makes perfect sense (thank you!) - based on that I would say let's use only the aria-checked like you originally had.

PR #2391, opened 2019-09-25
Add ability to create assignment or archive class from org#index page

JessRudder:
Part of #602
Note - this includes an update to the `octicon` gem

Adds a dropdown menu to the upper right corner of each classroom card on the classroom index page.

That menu currently contains two options:

`Create lesson` - Takes the user to the lesson creation flow for that classroom

archive-create-new

`Archive classroom` - Allows a user to archive the classroom from this view (Note: This becomes `Unarchive classroom` if the classroom is archived). It preserves all filters that were already in place for the view (e.g. if you are only viewing active, when you archive a classroom, it will be removed from view).

archive-from-index


Gemfile
```diff
@@ -46,7 +46,7 @@ gem "ims-lti", "~> 2.2.3"
 gem "local_time", "~> 2.0"
 
 gem "oauth",           "~> 0.5.4"
-gem "octicons_helper", "~> 2.1"
+gem "octicons_helper", "~> 9.1.1"
```
JessRudder: Updated because our old set didn't have the kebab-horizontal that was needed for the menu design.

app/controllers/organizations_controller.rb
```diff
@@ -56,12 +56,21 @@ def show_groupings
...
 
-    if result.success?
-      flash[:success] = "Successfully updated \"#{current_organization.title}\"!"
-      redirect_to current_organization
-    else
-      current_organization.reload
-      render :edit
+    respond_to do |format|
+      format.html do
+        if result.success?
```
JessRudder: I didn't change any of the current functionality of the html action. I just moved it up into the `respond_to` section for html.

config/initializers/bullet.rb
```diff
@@ -14,6 +14,9 @@
       end
 
       Bullet.raise = true if Rails.env.test?
+
+      # Need to eager load :users which is causing it to complain about the join table
+      Bullet.add_whitelist type: :unused_eager_loading, class_name: "Organization", association: :organizations_users
```
JessRudder: I previously fixed a warning that we needed eager loading of :users. That did get rid of that warning, but now says we're unnecessarily loading `:organizations_users` (which is the join table). This adds it to a whitelist so we won't be incessantly pinged by an alert that cannot be fixed. 😄

stephaniegiang:
This looks great! I just have one question. On an archived classroom, if I click the menu I still get `Create assignment` as one. Should teachers still be able to add assignments to archived classes?

stephaniegiang:
That makes sense. This can be a later discussion then 😄

JessRudder:
@stephaniegiang  great catch! I think currently archiving a classroom only stops people from accepting invites. I guess the thinking was that if a teacher has archived a classroom, they won't be making new lessons for it anyway.

@femmebot should we explicitly prevent creation of more assignments for archived classrooms? It should be a simple relatively simple change to make.

PR #2395, opened 2019-09-27
Do not allow new lessons to be created for archived classrooms

JessRudder:
This update prevents the creation of new assignments/group assignments for classrooms that have been archived.

On the classroom's assignment/group assignment page, the `New assignment` button is automatically disabled if the classroom is archived.

If someone alters the button markup to re-enable the button or navigates to the new assignment route directly, they will be redirected back and see a flash message telling them `You cannot create new assignments for archived classrooms`

archived-classroom-creation

On the classroom#index page, the `Create assignment` option will not show in the dropdown for archived classrooms.

image

No changes were made to the assignment creation flow for unarchived classrooms.


app/controllers/organizations_controller.rb
```diff
@@ -101,7 +101,14 @@ def remove_user
     redirect_to settings_invitations_organization_path
   end
 
-  def new_assignment; end
+  # rubocop:disable Style/GuardClause
+  def new_assignment
+    if current_organization.archived?
```
spinecone: Would it make sense to use the guard clause here rather than disabling the rubocop linter for it?

app/views/organizations/_organization_card_layout.html.erb
```diff
@@ -13,8 +13,10 @@
               </summary>
 
               <details-menu class="dropdown-menu dropdown-menu-sw">
-                <%= link_to new_assignment_organization_path(organization), class: 'dropdown-item btn-link btn-block text-left', role: 'button' do %>
-                  Create assignment
+                <% unless organization.archived? %>
+                  <%= link_to new_assignment_organization_path(organization), class: 'dropdown-item btn-link btn-block text-left', role: 'button' do %>
```
spinecone: Why did we decide to remove the "Create assignment" button here rather than disabling it, like in the `organizations/show` view?

app/views/organizations/show.html.erb
```diff
@@ -3,7 +3,9 @@
 <div class="container-lg p-responsive">
   <div class="d-md-flex flex-items-center flex-justify-between mt-5 mb-3">
     <h2 class="f2 text-normal">Assignments</h2>
-    <% if @assignments.present? %>
+    <% if @assignments.present? && current_organization.archived? %>
+      <%= link_to t('views.organizations.new_assignment'), new_assignment_organization_path(current_organization), class: 'btn btn-primary right disabled', role: 'button'%>
+    <% else %>
```
spinecone: Should this be `<% elsif @assignments.present? %>` to preserve the old behavior?

spinecone:
Should we also add a model validation to prevent assignments from being created on archived classrooms?

app/views/organizations/_organization_card_layout.html.erb
JessRudder: I left it in the classroom#show page because it's a page people are used to dealing with (and seeing the button there) and it seemed standard to leave the button in place but disable it if it can't be used.

I didn't include it in the dropdown because that's a new menu and feels more like, "What actions can I take on this classroom right now". It also seems more complicated to indicate that an item in a dropdown menu can't be selected vs making a button inactive.

I've pinged @femmebot to weigh in as it's definitely not consistent and I don't want to cause more confusion in my attempts to reduce confusion.

app/controllers/assignments_controller.rb
```diff
@@ -9,9 +9,16 @@ class AssignmentsController < ApplicationController
   before_action :set_filter_options, only: %i[show]
   before_action :set_unlinked_users, only: %i[show]
 
+  # rubocop:disable Style/AndOr
   def new
+    if @organization.archived?
```
spinecone: Should we keep this redirect now that there's a model validation that'll fail if they try to use the form?

app/views/organizations/_organization_card_layout.html.erb
JessRudder: I confirmed with @femmebot that this behavior is consistent with our designs elsewhere.

For the button, we want to disable it.

For the dropdown, we only want to include actions the user can take.

JessRudder:
@spinecone this is ready for a re-review when you have a chance.

For the [guard clause](https://github.com/education/classroom/pull/2395#pullrequestreview-295291012) I couldn't do a one-liner since I have to set the flash message and redirect if the organization is already archived.

To get around that, I pulled the check into its own method and then I call that method in the guard clause. It feels a bit over-engineered, but, it does make the linter happy. If there's a cleaner way of doing this as a guard clause instead of an if statement, let me know!

app/controllers/organizations_controller.rb
```diff
@@ -225,4 +225,3 @@ def verify_user_belongs_to_organization
     not_found unless current_organization.users.map(&:id).include?(@removed_user.id)
   end
 end
-# rubocop:enable Metrics/ClassLength
```
JessRudder: This didn't seem to have a partner that was disabling anything. No linters failed when I removed it either.

app/models/group_assignment.rb
```diff
@@ -82,4 +83,8 @@ def max_teams_less_than_group_count
       errors.add(:max_teams, "is less than the number of existing teams (#{group_count})")
     end
   end
+
+  def organization_is_not_archived
```
spinecone: Someday these models will be generalized 🙏 😭

Issue #2363, opened 2019-09-11
Show exact deadline datetime on hover


d12:
When you set a deadline, this is what you see on on the assignment page:

<img width="252" alt="Screen Shot 2019-09-11 at 2 58 59 PM" src="https://user-images.githubusercontent.com/30920216/64726475-bc46ea80-d4a4-11e9-829a-ced78b2a31cf.png">

It's not very descriptive, it's hard to tell exactly what day it's talking about, and I don't know what time the deadline is at.

2 possible solutions:

a) Hovering can show the exact datetime
b) Change the text to be human readable but more useful. "This Friday at 8:00 am"?

cc https://education.github.community/t/github-assignment-date-what-time-zone/44939/2

spinecone:
Right now we're using the `distance_of_time_in_words` function which is intended to convert a timestamp into human readable text; would it make sense to just add the exact time (formatted), so it would say something like `Deadline in 1 day (12/12/19 at 8:00 am)` ?

d12:
Yes! We'll need to check what the users timezone is, we don't store it in the database so we'll have to check in JS.

Issue #2073, opened 2019-07-23
Improve deadline picker


d12:
A couple of issues with the current deadline picker:

- [ ] Users can only pick times exactly on an hour. To get around that, teachers have to manually edit the ISO8601 datetime.
- [ ] If the datetime is edited and not perfectly formatted, clicking away from the textbox undoes all changes.
- [ ] Part of the string is the users time zone. We probably don't need to have this being user visible? They'll never need to set a deadline in a different timezone (probably?)
- [ ] Scrolling the time section of the datetime picker goes _extremely_ quickly, I usually just click the little arrows until I find the right time.

wolfieorama:
anyone on this issue ? if not willing to take it up as my first contribution

d12:
Yes please!! That'd be great @wolfieorama

wolfieorama:
cool on it @d12

yves-chevallier:
I would also say that the format should be always ISO8601 not 09/23/2019 00:00 +200 but 2019-09-23 00:00+02:00

PR #2396, opened 2019-09-30
convert deadline to local time in assignment display

spinecone:
For https://github.com/education/classroom/issues/2363

Include the exact deadline on the assignment show page (in current user's local timezone)

It currently looks like this:
Screen Shot 2019-09-30 at 11 54 14 AM

We are discussing making further improvements to the time formatting, such as using 24 hour time and including the full month name and the user's timezone in the display.

### DIFF
--- a/app/views/assignments/show.html.erb
+++ b/app/views/assignments/show.html.erb
@@ -19,7 +19,7 @@
                 <% if @assignment.deadline.passed? %>
                   - Deadline Passed
                 <% else %>
-                  <%= "- Deadline in #{distance_of_time_in_words(Time.zone.now, @assignment.deadline.deadline_at)}" %>
+                  <%= render partial: "shared/deadline_with_local_time", locals: { deadline: @assignment.deadline.deadline_at } %>
                 <% end %>
               <% end %>
             </p>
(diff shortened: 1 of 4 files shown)


spinecone:
Further down in that thread it looks like they had further questions about seeing the exact deadline time on the assignment page:
Screen Shot 2019-09-30 at 1 01 56 PM

spinecone:
It sounds like there are 2 distinct problems that were identified:
- The form field for setting a deadline using a timestamp format that is difficult to interpret the timezone of (not currently addressed by this PR)
- The assignment show page does not show an exact time for the deadline, in any time zone (WIP in this PR)

Does that sound right (cc @d12 ) ?

d12:
There are other issues with the deadline picker, outlined here https://github.com/education/classroom/issues/2073

For the datetime display issues, this looks good to me :+1:

spinecone:
Going to change the format to `Jan 1, 2020, 17:00 (EDT)` which was discussed in slack ([month name] [date] [year], [24 hour time] (timezone)).

spinecone:
It looks like this now:

Screen Shot 2019-10-01 at 10 12 57 AM

d12:
Hm, I'd remove the () around the timezone. Sorry, that was my suggestion 🙈

spinecone:
Without parentheses: 
Screen Shot 2019-10-01 at 10 36 40 AM

spinecone:
Let me know if this looks ok to deploy, @d12 and @JessRudder

PR #2405, opened 2019-10-03
use request stubs instead of vcr in assignment model specs

spinecone:
This is an example of how we might want to switch out vcr for request stubs moving forward. I used the spec helpers from education web (https://github.com/github/education-web/tree/master/spec) as a foundation to stub requests that GitHub Classroom makes for github.com data.


spinecone:
Currently working on trying to use a stubbed octokit client instead of `oauth_client`

spinecone:
Yeah, I'm conflicted about how much to abstract requests from the individual tests. It's definitely more effort, but it's also more explicit and ensures that we're only making requests that we expect. Not sure if it's a positive or negative impact on readability.

spec/spec_helper.rb
```diff
@@ -90,3 +90,27 @@ def json_payload(file_path)
...
+  @octokit_client ||= begin
+    @octokit_client = instance_double("Octokit::Client")
+    allow(Octokit::Client).to receive(:new).and_return(@octokit_client)
+    allow(@octokit_client).to receive(:access_token).and_return(1234)
+    @octokit_client
+  end
+end
+
+def stub_org_request(org_id)
+  org_response = OpenStruct.new(login: "fake org")
```
jeffrafter: Non-blocking: I'm really interested long term about the speed of OpenStructs and wonder if we might want something more purpose built. But I do like that it has the ability to mix in test-specific params. I don't have a better suggestion or anything, just curious about it.

spinecone: I also stole this approach from education web; I would be interested to know if @renz45 
or @tessgriffin have run into any performance issues with the tests over there.

andrewbredow: This is pretty interesting https://gist.github.com/jgaskins/3892455

It looks like either of the solutions that have been mentioned here are better than using RSpec Mocks whenever possible.

andrewbredow: LOL... just noticed that was from 7 years ago... *maybe* it's still true 😝

Issue #2412, opened 2019-10-07
Teams and assignment repos unable to load


englieh:
When I go to the classroom and assignment created, the teams and repo cannot be loaded.
![image](https://user-images.githubusercontent.com/6601940/66356228-e05ef400-e99c-11e9-96db-6419a12763ce.png)

jeffrafter:
@englieh thanks for the report, I am looking into this.

jeffrafter:
@englieh I am seeing this as fixed in production, sorry about the trouble!

englieh:
No worries. Fixed. Thanks!

PR #2413, opened 2019-10-07
Revert "subtract starter repo commits from assignment repo commit count"

jeffrafter:
Reverts education/classroom#2409

Unfortunately this is causing an issue for group assignments. We'll need to adjust that before moving this forward.

Fixes https://github.com/education/classroom/issues/2412


jeffrafter:
We'll want to bring this code back with a fix for group assignments. 🙇

PR #2418, opened 2019-10-08
define number_of_commits on group_assignment_repo as well

spinecone:
Second attempt at https://github.com/education/classroom/pull/2409

Defines `number_of_commits` on a module shared by `assignment_repo` and `group_assignment_repo`


spinecone:
The assignment_repo specs got lost in some git confusion, I'm digging those up now

PR #2427, opened 2019-10-14
Fix search/filter so it works on Firefox

JessRudder:
Previously the search/filter form on the `/classrooms` page was not working in Firefox. It was submitting twice (the first time correctly as an async request and the second time as a full page load).

This updates the `dispatchEvent` for the form submission so it includes `cancelable: true` which prevents the double submission.

There was some additional cleanup work included here:
 - [x] Get debounce working correctly
 - [x] Add primer's JS for the detail-menu-element so we can remove the custom JS we'd written
 - [x] Remove `data-autosubmit: true` from the markup as it's not doing anything here

Eventually, we should replace our custom debounce functions with a standardized implementation. The JS folks recommended https://github.com/github/mini-throttle/. This will be done in a later PR as it will require some larger changes.

### DIFF
--- a/app/views/organizations/_organization_filters.html.erb
+++ b/app/views/organizations/_organization_filters.html.erb
@@ -38,7 +38,7 @@
           <div class="SelectMenu-list">
             <% view_options.each do |name| %>
               <label class="SelectMenu-item" role="menuitemradio" aria-checked="<%= active_view_option == name %>" tabindex="0">
-                <%= radio_button_tag "view", name, active_view_option == name, "hidden": "true", "data-autosubmit": "true" %>
+                <%= radio_button_tag "view", name, active_view_option == name, "hidden": "true" %>
                 <%= octicon("check", class: "SelectMenu-icon") %>
                 <span class="text-normal" data-menu-button-text><%= name %></span>
               </label>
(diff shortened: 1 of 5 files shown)


app/views/organizations/_organization_filters.html.erb
```diff
@@ -38,7 +38,7 @@
           <div class="SelectMenu-list">
             <% view_options.each do |name| %>
               <label class="SelectMenu-item" role="menuitemradio" aria-checked="<%= active_view_option == name %>" tabindex="0">
-                <%= radio_button_tag "view", name, active_view_option == name, "hidden": "true", "data-autosubmit": "true" %>
+                <%= radio_button_tag "view", name, active_view_option == name, "hidden": "true" %>
```
jeffrafter: Does this still trigger when clicking in Safari? (I think `change` catches this here: https://github.com/education/classroom/pull/2427/files#diff-121f9615472808b2e0a69f8ea7a51c25R12)

JessRudder: Yep! Still works chrome and safari. And now it works in Firefox too!

I have no idea about IE and Edge....but I don't have a free way (that I know of) to check them.

PR #2433, opened 2019-10-17
Revert "Skip user cache in a few places where we make API calls"

d12:
Reverts education/classroom#2198


JessRudder:
Closing out this revert PR as this was not the commit we needed to revert. The issue was identified and solved.

PR #2439, opened 2019-10-21
Add roster identifier to assignment repos endpoint

d12:
<draft>


app/serializers/assignment_repo_serializer.rb
```diff
@@ -13,5 +14,10 @@ def username
   def displayName
     object.user.github_user.name || ""
   end
+
+  def rosterIdentifier
+    return nil unless instance_options[:roster]
+    instance_options[:roster].roster_entries.find_by(user_id: object.user.id)&.identifier
```
jeffrafter: Is this an N+1?

app/controllers/api/assignment_repos_controller.rb
```diff
@@ -9,7 +9,7 @@ class AssignmentReposController < API::ApplicationController
 
     def index
       repos = AssignmentRepo.where(assignment: @assignment).order(:id)
-      paginate json: repos
+      paginate json: repos, roster: @assignment.organization.roster
```
jeffrafter: Is there a test which covers this pagination change? Will it be surprising to folks when deployed?

jeffrafter: Is the `@assignment.organization` always the same as the `@organization` here?

d12: The `roster` kwarg here doesn't add anything to the JSON payload, it adds context that I can access in the serializer via `instance_options`. The alternative is to do something like `repo.assignment.organization.roster.roster_entries.find...` for every repo in the serializer 😬

d12: Yep, didn't see `@organization`, changed it to use that instead.

app/serializers/assignment_repo_serializer.rb
d12: The results of the `roster.roster_entries` query would _probably_ be cached, but I changed this to pass in roster entries instead of roster so we don't need to make this query every time.

d12:
@jeffrafter All comments addressed, this is ready for review again!

PR #2451, opened 2019-10-25
copy changes to help teachers come up with a useful classroom name

femmebot:
## What
Small copy changes to help new GitHub Classroom users figure out a good name for their classroom

### Why
While observing teachers create new classrooms, I noticed that none of them actually used the classroom name that was auto-generated for them. But new users, in particular, weren't sure what _would_ make a good classroom name. They didn't know whether the name would be visible to students. They weren't sure if they needed to create separate classrooms for different sections of the same course. 

Hopefully, these small copy changes can help address those issues.

screencapture-localhost-5000-classrooms-57011530-icm200-classroom-1-setup-2019-10-25-15_48_30

### DIFF
--- a/config/locales/views/organizations/en.yml
+++ b/config/locales/views/organizations/en.yml
@@ -72,8 +72,9 @@ en:
       group_assignment: 'Group assignment'
       group_assignment_desc: 'Participants form teams that work together on a shared repository.'
       create_group_assignment: 'Create a group assignment'
-      name_your_classroom: 'Name your classroom'
-      give_descriptive_name: 'Please give your classroom a more descriptive name'
+      name_your_classroom: 'Set up your classroom'
+      give_descriptive_name: 'Classroom name'
+      classroom_name_note: 'Try using your course name and section so students can identify your classroom.'
       set_of_teams: 'Set of teams'
       no_teams: "This classroom doesn't have any sets of teams yet."
       assignments: 'Assignments'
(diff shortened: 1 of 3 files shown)


config/locales/views/organizations/en.yml
```diff
@@ -72,8 +72,9 @@ en:
       group_assignment: 'Group assignment'
       group_assignment_desc: 'Participants form teams that work together on a shared repository.'
       create_group_assignment: 'Create a group assignment'
-      name_your_classroom: 'Name your classroom'
-      give_descriptive_name: 'Please give your classroom a more descriptive name'
+      name_your_classroom: 'Set up your classroom'
+      give_descriptive_name: 'Classroom name'
+      classroom_name_note: 'Try using your course name and section so students can identify your classroom.'
```
spinecone: "Try" kind of sounds like we don't allow anything else, would something like "consider" sound like less of an imperative?

femmebot: Yeah I was looking for something short so the note wouldn't extend to two lines.

femmebot: Also, since the placeholder itself isn't their course name, it seems to suggest that it isn't an imperative?

spinecone: "Using your course name and section can help students identify your classroom." is shorter if that helps? Sorry, I just assume that I don't have a choice when someone tells me to try something 😄

femmebot: Too bad we don't have a Yoda emoji. "There is no `try`"

Issue #2450, opened 2019-10-25
Request to change "Enable assignment invitation URL" checkbox to "Disable assignment invitation URL"


MeyAyre:
### Feature request :sparkles:
The 'enable assignment URL' checkbox is checked by default, which makes for a confusing user flow since unchecking it actually disables this ability. Changing this to a checkbox that you click to disable the assignment invitation URL would improve the user flow and be easier to capture in user-facing documentation.

![image](https://user-images.githubusercontent.com/14324873/67593168-678ac680-f716-11e9-8e38-5c4de2d0f1be.png)

femmebot:
>Changing this to a checkbox that you click to disable the assignment invitation URL

We probably want to keep the existing behavior due to muscle memory since current users are already accustomed to equating `checked` with `enabled`. What gives them pause is it isn't obvious to new users _why_ would you want to uncheck it? I think we can just add some verbiage to help surface that. 

A couple of options:

### Copy changes
![Enable invitation link](https://user-images.githubusercontent.com/471514/67621809-66729b80-f7e1-11e9-8997-2d205aada956.png)

### Tooltip
![Enable invitation tooltip](https://user-images.githubusercontent.com/471514/67621810-670b3200-f7e1-11e9-9cca-c9fc70d132da.png)

PR #2461, opened 2019-11-02
Enhanced help text for assignment invitation link

femmebot:
## What
Adds verbiage to explain _why_ someone might want to disable the invitation URL. Moved copy into a tooltip since this is information that they only need to read once or twice and don't need to be reminded again with repeat use.

Also made small edit to the student admin option. "Grant" is preferred over "giving" permissions (see discussion on [terminology used for permission management](https://docs.google.com/document/d/1uETpjakXLTugMyNS5Quke-mg41rk3cG8mRnLlpg0ru4/edit?usp=sharing) and [summary](https://docs.google.com/document/d/1uETpjakXLTugMyNS5Quke-mg41rk3cG8mRnLlpg0ru4/edit#heading=h.cfvzir2p60tg)).

Closes https://github.com/education/classroom/issues/2450

invitation-tooltip


app/views/shared/_invitation.html.erb
```diff
@@ -6,13 +6,10 @@
...
-  <label>Enable assignment invitation URL</label>
+  <label class="pl-1">Enable assignment invitation URL <span class="tooltipped tooltipped-no-delay tooltipped-n tooltipped-multiline d-inline-block pl-1" style="bottom: 3px" aria-label="While enabled, students will be able to accept and access this assignment. When you no longer want anyone to access this assignment, uncheck to deactivate the invitation link."><%= octicon 'info' %></span></label>
   <span style="color: green"><%= octicon 'check', height: 16, id: "checkmark", style: "display: none; padding-bottom: 3px;" %></span>
   <%= image_tag 'octocat-spinner-128.gif', width: 13, height: 13, class: 'spinner', id: "loading-indicator", style: "display:none" %>
-  <div class="input-group input-block pt-2 pb-1">
+  <div class="input-group input-block pt-2 pb-4">
     <%= render partial: "shared/invitations/clipboard_form", locals: local_assigns %>
   </div>
-  <p class="note mb-3">
-    <%= t('views.shared.share_invite_url') %>
```
jeffrafter: Is there a reason this was removed?

femmebot: Ah, good catch! I only wanted to remove the redundant copy. I just reinstated it but only keep the message about sharing the URL since that's the important action.

PR #2469, opened 2019-11-14
Security updates

jeffrafter:
There are security updates for loofah and json-jwt. This applies those and should be a no-op.


jeffrafter:
This is done! Thanks @JessRudder!! 🎉

PR #2485, opened 2019-12-02
Encrypt tokens at rest

jeffrafter:
We'd like to encrypt the stored tokens at rest. To pursue this we'll be leveraging rbnacl with SimpleBox support. This relies on libsodium and is _good_. To build this we'll be building custom readers and writers for the token field to allow for versioned token generations and revocation.

https://github.com/RubyCrypto/rbnacl/wiki/SimpleBox

We'll also need to install libsodium on Travis and Heroku (via buildpack)

To deploy this we will need to pursue one of the following strategies:

### New field option

1. Create a new field `crypted_token`

2. Write to both fields: `token`, `crypted_token`

3. Run a transition to write legacy encrypted tokens

4. Read from `crypted_token`

5. Delete `token`

### Existing field option

0. Ensure no tokens start with `GH_`, make sure there are no nil tokens (confirmed!)

1. Make a custom writer that writes the `GH_` prefix and encrypted version of the password (check length constraints)

2. Make a custom reader that checks for the `GH_` prefix and if there decrypts the remaining password or returns the token as is

3. Make sure:
    - the `token_was` and `token_changed?` functions still works (in `ensure_no_token_scope_loss`)
    - the uniqueness and presence validations still work (if you set it to nil does the prefix still get added?)

Given these two options, we are likely going to choose the existing field option.

cc @JessRudder @education/classroom 

https://api.rubyonrails.org/classes/ActiveRecord/Base.html#class-ActiveRecord::Base-label-Overwriting+default+accessors


app/models/user.rb
```diff
@@ -92,4 +105,26 @@ def ensure_no_token_scope_loss
...
+    token_secret = Rails.application.secrets.token_secret
+    raise "TOKEN_SECRET is not set, please check you .env file" if token_secret.blank?
+    @encryptor ||= RbNaCl::SimpleBox.from_secret_key(token_secret.unpack("m0").first)
+  end
+
+  def encrypt_value(value)
+    return if value.blank?
+
+    ciphertext = encryptor.encrypt(value)
+    hex = ciphertext.bytes.pack("c*").unpack("H*").first
```
jeffrafter: Hex encoding gives a consistent string (no new lines) base64 had some oddities. All the same this could be:

```
  def encrypt_value(value)
    return if value.blank?
    ciphertext = encryptor.encrypt(value)
    "GH_" + Base64.encode64(ciphertext)
  end
```

app/models/user.rb
```diff
@@ -92,4 +105,26 @@ def ensure_no_token_scope_loss
...
+    "GH_" + hex
+  end
+
+  def decrypt_value(value)
+    return value unless value && value =~ /\AGH_/
+
+    hex = value.sub(/\AGH_/, '')
```
jeffrafter: If using base64 above this would become:

```
  def decrypt_value(value)
    return value unless value && value =~ /\AGH_/
    ciphertext = value.sub(/\AGH_/, '')
    encryptor.decrypt(Base64.decode64(ciphertext))
  end
```

app/models/user.rb
JessRudder: Thanks for getting this rolling? Any reason to do Base64 here if we get a more consistent string with hex? Is Base64 more secure?

jeffrafter: No reason except encrypted key size

